PR #451, opened 2014-11-18
Rails4 upgrade

tygriffin:
Okay, this is a big one! **Rails 4 Upgrade**

## Please take special note of a few things in particular:

1. I've upgraded FriendlyId to 5 which included lots of changes. In an effort to get tests passing again, I disabled the reserved words feature, but we should probably add this back in. Also, when looking up records by a FriendlyId it is now recommended to do this: `Model.friendly.find(id)`. But again, to expedite the Rails 4 upgrade I disabled this to mimic the old behavior: `Model.find(id)`. We may want to enable the default, recommended behavior.

2. All sweeper models broke in the upgrade because Observers were extracted into a gem. However, adding the gem back in did not fix the sweepers immediately. This is because to make observers work, the sweepers must inherit from `ActionController::Caching::Sweeper`, but this class does not implement `expire_fragment`. Therefore, I made a Sweeper class that inherits from `ActionController::Caching::Sweeper` and mixes in `ActionController::Caching::Fragments` (in which `expire_fragments` resides). This seems to work fine, but it definitely needs closer attention by someone who understands Rails caching better than I. (By the way, why are the sweepers in the models directory? Shouldn't they be with controllers?)

3. I turned off verify partial doubles in RSpec, but we may want to re-enable that feature.

4. Generally, I have tried to get back to green (passing tests) as quickly as possibly, without worrying too much about new Rails 4 features. I think we should definitely consider implementing some of them in separate stories though. For example, the following:

    i. Turbolinks
    ii. Concerns
    iii. secrets.yml


db/seeds.rb
```diff
@@ -12,8 +12,8 @@ def load_data
     load_crops
     load_plant_parts
 
-    # for development environments only
-    if Rails.env.development?
+    # for development and staging environments only
+    if Rails.env.development? || Rails.env.staging?
```
tygriffin: @Skud / @pozorvlak : This change made me a bit nervous. Can you confirm this is okay?

pozorvlak: This would create test1, test2, wrangler1 and admin1 users on staging with known-in-advance passwords? That doesn't sound right. I think we *should* run `load_paid_account_types` on staging, though.

tygriffin: That makes sense. We should also run `load_products`, right? I've made [PR 627](https://github.com/Growstuff/growstuff/pull/627) for this.

Issue #582, opened 2014-12-12
Implement v1 crop API


Skud:
Using the API v1 framework, and after discussion with the developer community, create a new version of the crop API which remedies the problems of the v0 API, eg. the need to loop through all plantings to get useful planting information.

PR #616, opened 2015-01-07
Api v1 - initial start

pmackay:
Based on jbuilder and introducing some JSON-LD elements. Work in progress. For #582 (I think).


app/controllers/api/v1/crops_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Api::V1::CropsController < ApplicationController
+
+  # GET /api/v1/crops
+  def index
```
Skud: I suggest that for now we just got as minimal as we can.  Since we're only including ID/name, let's remove the sort params and all that other stuff that's included here.

config/routes.rb
```diff
@@ -84,6 +84,14 @@
   match '/admin/newsletter' => 'admin#newsletter', :as => :admin_newsletter
   match '/admin/:action' => 'admin#:action'
 
-
+  namespace :api, :defaults => {:format => :json}  do
+    namespace :v1 do
+      resources :crops
+      resources :gardens
```
Skud: let's remove the resources we're not doing yet

app/views/api/v1/crops/index.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.data @crops do |crop|
+  json.(crop, :id, :name)
+  json.set! '@id', url_for(:only_path => false) + '/' + crop.id.to_s
```
Skud: this is json-ld, right?

in re: our discussion about pretty URLs, it occurs to me that /type/id (eg. /crops/42 or /plantings/42) will be unchanging forever while the friendly URLs (as generated by the friendly_id gem) will tend to change anyway with name changes eg. /posts/unspecific-title to /posts/help-me-with-my-aphid-problem or whatever.  This is similar to wordpress keeping the ?p=nnn the same while changing the URL when you change a page title.

Anyway there's maybe a discussion to be had about having a "machine-friendly" URL that is unchangeable and using that in the API.

app/views/api/v1/crops/show.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.data do
+  json.(@crop, :id, :name, :creator_id, :en_wikipedia_url)
+  json.set! '@id', url_for(:only_path => false)
```
Skud: this is inconsistent with the @id/url given in index.json.builder... which do we want?

spec/api/v1/api_crops_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      json['license']['url'].should == "http://creativecommons.org/licenses/by-sa/3.0/"
+      json['license']['credit'].should == "Growstuff"
+      json['license']['link'].should == "http://growstuff.org/"
+      json['license']['easy_link'].should == '<a href="http://creativecommons.org/licenses/by-sa/3.0/">CC-BY-SA</a> <a href="http://growstuff.org/">Growstuff</a>'
+    end
+  end
+
+  describe "GET crops single item" do
+
+    it 'fetches one crop' do
```
Skud: should also test that it contains license information

spec/api/v1/api_crops_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    @crop = FactoryGirl.create(:tomato)
+    @crop.save
+  end
+
+  describe "GET crops index" do
+    it 'fetches the crops index' do
+      get '/api/v1/crops'
+      response.should be_success
+      response.content_type.should eq("application/json")
+    end
```
Skud: i'd like to see some tests here for the actual content of the JSON, i.e. that it contains a list with one item, which is tomato.

spec/api/v1/api_crops_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+describe Api::V1::CropsController do
+
+  before(:each) do
+    @crop = FactoryGirl.create(:tomato)
+    @crop.save
```
Skud: this .save is superfluous here... FG.create does that anyway.  But apart from that, let's use the newer RSpec syntax, "let", instead of "before :each".  Also, the tests following should use the expect(blah).to syntax, not blah.should.

spec/api/v1/api_crops_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  describe "GET crops single item" do
+
+    it 'fetches one crop' do
+      @crop2 = Crop.first
+      get '/api/v1/crops/' + @crop2.id.to_s
+      response.should be_success
+      response.content_type.should eq("application/json")
+
```
Skud: again, would like to check that this JSON actually has the tomato data

Skud:
Yes, #582 is probably the best issue to link to -- there was a PT story (as you know) for the framework but I didn't move it across as it was open and I hadn't quite figured out what to do with open stories.

I've left a bajillion line notes as you've no doubt seen already, but overall it looks good.  If you could do some of those small things I mentioned, I'll be happy to merge.  (We can have an ongoing discussion about IDs, URLs, etc over on Growstuff Talk I think. I won't let it block this PR.)

spec/api/v1/api_crops_controller_spec.rb
tygriffin: Might be a good idea to check the length of the JSON content as well, to make sure that only the specified content is contained within. That way, if we add or remove anything later, the test will fail reminding us to update it.

tygriffin: Also, what do you guys think of the be_success matcher? I generally choose the more specific numeric codes and that seems like it might be better when testing an API.

Skud: Good point. I feel like 200 is exactly what we want here, not eg. 206 
"Partial content".

For reference: 
http://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success

On 8/01/2015 7:36 am, Taylor Griffin wrote:
>
> In spec/api/v1/api_crops_controller_spec.rb 
> <https://github.com/Growstuff/growstuff/pull/616#discussion-diff-22613777>:
>
> > @@ -0,0 +1,45 @@
> > +require 'spec_helper'
> > +
> > +describe Api::V1::CropsController do
> > +
> > +  before(:each) do
> > +    @crop = FactoryGirl.create(:tomato)
> > +    @crop.save
> > +  end
> > +
> > +  describe "GET crops index" do
> > +    it 'fetches the crops index' do
> > +      get '/api/v1/crops'
> > +      response.should be_success
> > +      response.content_type.should eq("application/json")
> > +    end
>
> Also, what do you guys think of the be_success matcher? I generally 
> choose the more specific numeric codes and that seems like it might be 
> better when testing an API.
>
> —
> <https://github.com/Growstuff/growstuff/pull/616/files#r22613777>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

config/routes.rb
pmackay: I have other files for the other main content types, once crops is looking better would it make sense to enable the endpoints for those at a very basic level too?

spec/api/v1/api_crops_controller_spec.rb
pmackay: How do you feel about adding https://github.com/collectiveidea/json_spec to the development gems for this?

app/views/api/v1/crops/show.json.jbuilder
pmackay: Will add `url` to `show`.

spec/api/v1/api_crops_controller_spec.rb
Skud: Good! Go for it!

config/routes.rb
Skud: No, let's just do crops for now. Please remove the extraneous routes; we'll do them one at a time.

pmackay:
Have made various updates. The RSpec tests could probably be improved further. I've spent quite a bit of time on those but (being still quite new to rspec) I'm finding it harder to follow the docs I can find on it compared with other parts of Rails. Any good recommendations much appreciated :)

spec/api/v1/controllers/api_crops_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def check_license(response)
+    json = JSON.parse response.body
+    json['license'].should_not == nil
+    json['license']['name'].should == "Creative Commons Attribution ShareAlike 3.0 Unported"
+    json['license']['short_name'].should == "CC-BY-SA"
+    json['license']['url'].should == "http://creativecommons.org/licenses/by-sa/3.0/"
+    json['license']['credit'].should == "Growstuff"
+    json['license']['link'].should == "http://growstuff.org/"
+    json['license']['easy_link'].should == '<a href="http://creativecommons.org/licenses/by-sa/3.0/">CC-BY-SA</a> <a href="http://growstuff.org/">Growstuff</a>'
```
tygriffin: Can this be written like so? 

`expect(json['license']).to_not eq(nil)`

(Otherwise the tests are looking great!)

pmackay:
@tygriffin I've updated those, thanks!

Skud:
OK, this looks good to merge to the Growstuff API branch.  Thanks for all your fixes, @pmackay!

Wrt rspec documentation, ugh, I knooooow.  This website is a good rundown of best practices: http://betterspecs.org/  Apart from that I'm afraid I mostly just cargo-culted it until I kinda knew what I was doing.  I think there might be a good book out there, but I don't have it.

PR #623, opened 2015-01-11
Add SEO-friendly language to the crop detail page

### LINKED ISSUE #528: SEO: more searchable language on crop detail page
The crop detail page should have more search-friendly language, eg. "How to grow squash" rather than "Planting advice".  Especially, the first paragraph or so should say something about "Information about how to grow squash anywhere in the world" or something like that.

pozorvlak:
Fixes #528.

### DIFF
--- a/app/views/crops/_find_seeds.html.haml
+++ b/app/views/crops/_find_seeds.html.haml
@@ -1,4 +1,4 @@
-%h4 Find seeds
+%h4 Find #{ crop.name } seeds
 - if crop.seeds.empty?
   %p
     There are no seeds available to trade.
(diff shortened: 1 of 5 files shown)


tygriffin:
As we're updating content, would it make sense to extract strings to *config/locales/en.yml* piecemeal?

PS: I'm still not 100% certain how those locale files should be structured.

Skud:
Hmmm. If we're not sure how to structure them let's not bother with extracting these strings right now, but let's have a discussion about it soon.

Skud:
@tygriffin are you ok with this other than the to-be-figured-out i18n issue? can it be merged?

tygriffin:
Good to go!

PR #627, opened 2015-01-11
don't load test and admin users in staging

tygriffin:
This is something @pozorvlak mentioned would make sense. Seems sensible to run `load_paid_account_types` and `load_products` in staging, but what about the users? ( @Skud )

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -11,13 +11,13 @@ def load_data
     create_cropbot
     load_crops
     load_plant_parts
+    load_paid_account_types
+    load_products
 
     # for development and staging environments only
-    if Rails.env.development? || Rails.env.staging?
+    if Rails.env.development?
       load_test_users
       load_admin_users
-      load_paid_account_types
-      load_products
     end
   end
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -14,7 +14,9 @@ def load_data
     load_paid_account_types
     load_products
 
-    # for development and staging environments only
+    # We don't load these in an environment except development to
+    # prevent creating users in the wile - especially admins - with
+    # known passwords.
     if Rails.env.development?
       load_test_users
       load_admin_users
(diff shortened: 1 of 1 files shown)


Skud:
No, we certainly shouldn't be creating admin users etc, with known passwords, on staging!

db/seeds.rb
```diff
@@ -11,13 +11,13 @@ def load_data
     create_cropbot
     load_crops
     load_plant_parts
+    load_paid_account_types
+    load_products
 
     # for development and staging environments only
```
Skud: you need to change the comment back as well ;)

and maybe leave a note saying NOT to create these accounts on staging because they have publicly known passwords.

pozorvlak:
Looks good to me, modulo @Skud's requested changes to the comments.

pozorvlak:
Deployed to staging. I didn't run `scripts/deploy_tasks.sh` because it looked like staging was already up-to-date with all of those, and I didn't run the `figaro` step because AFAICT that would have overwritten staging's config with my (empty!) local one - @Skud, was this correct?

Skud:
Yup that seems right.  If you're going to be deploying regularly then 
you probably do want to fill your application.yml -- if you do "heroku 
config --app=growstuff-staging" I think it'll list the current values 
for things.

On 13/01/2015 1:09 am, pozorvlak wrote:
>
> Deployed to staging. I didn't run |scripts/deploy_tasks.sh| because it 
> looked like staging was already up-to-date with all of those, and I 
> didn't run the |figaro| step because AFAICT that would have 
> overwritten staging's config with my (empty!) local one - @Skud 
> <https://github.com/Skud>, was this correct?
>
> —
> <https://github.com/Growstuff/growstuff/pull/627#issuecomment-69574252>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

PR #635, opened 2015-01-15
Fix cache key

tygriffin:
In response to [this bug](http://talk.growstuff.org/t/please-test-rails-4/173/28).

It seems the issue is that the `cache_key_for` method (defined in ApplicationHelper) always returns the same key if the same class is passed into it, so I've added an additional argument which can be included in the key to make it uniquely identifying. As it stands the key is generated by looking at the model name, the total count of records for that model and the time of the most recently updated record. If any of those change, the key is refreshed.

I turned caching on in development and this fix appears to work.

I might suggest making a separate story to rip out all this caching stuff and then rebuild it based on New Relic stats.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -57,7 +57,7 @@ class CommentsController < ApplicationController
 
     respond_to do |format|
       if @comment.save
-        format.html { redirect_to @comment.post, notice: 'Comment was successfully created.' }
+        format.html { redirect_to @comment.post, notice: "Comment was successfully created." }
         format.json { render json: @comment, status: :created, location: @comment }
       else
         format.html { render action: "new" }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -24,6 +24,7 @@ module ApplicationHelper
       :target => "_blank"
   end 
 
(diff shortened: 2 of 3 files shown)


pozorvlak:
Good idea re: caching the things New Relic tells us to. Your PR looks good; I'll merge it iff it passes Travis.

pozorvlak:
D'oh! Looks like we somehow hit a bug in PhantomJS itself: @tygriffin, can you investigate further?

pozorvlak:
Cool, though we should submit a bug report to PhantomJS as they request: can you handle that?

tygriffin:
Hmm... I couldn't reproduce that error locally. After [poking around the internet](https://github.com/ariya/phantomjs/issues/10666) it seems that error occasionally rears its head and it's usually spurious. 

I just triggered Travis again and all is well. I say we let it slide this time. Then if it happens again, commence worrying.

pozorvlak:
Oh, known issue. Bah :-( I'm really not happy about having intermittently-failing tests, though I guess it should be easy to tell when it's this issue as opposed to something in our code causing the failure.

Skud:
Actually the original caching *was* based on new relic's performance stats!

pozorvlak:
D'oh! <.< >.>

PR #642, opened 2015-01-16
Release7 tweaks

Skud:
Mostly just tweaking the display of crops on the homepage, now that the column spacing is different

- made columns 8/4 rather than 6/6
- show 8 interesting crops, not 12
- show 6 recent plantings, not 4
- use crop thumbnail with visible name/info/etc, rather than popup version
- fix up some thumbnail CSS
- made members display 2 across not 3 across
- made some variations to what's shown on mobile screens (now: crops, which have clickable links, and were previously supressed)

Screenshot shows how it looks:

screenshot 2015-01-16 21 34 56

### DIFF
--- a/app/assets/stylesheets/overrides.css.less
+++ b/app/assets/stylesheets/overrides.css.less
@@ -124,7 +124,7 @@ p.stats {
   color: @brown;
 }
 
-.crop-thumbnail {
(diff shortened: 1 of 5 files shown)


pozorvlak:
LGTM, will merge when Travis passes.

PR #647, opened 2015-01-16
Tidied up authorization around follows

### LINKED ISSUE #638: Add ability checking to follows controller
The "follows" controller doesn't currently check for abilities, so [it's possible to follow yourself](http://talk.growstuff.org/t/please-test-allow-members-to-follow-other-members/156/5?u=skud).

This is a very minor thing because you can only really do it by editing the HTML like thorfinn did, but it should be fixed.

Skud:
Fixes #638.

Or at least, sorta-kinda fixes it as much as I'm going to.  Copypasta from my comments over there:

I've added the checks to the controller but I don't think we can realistically prevent people from doing what thorfinn did in the link above (i.e. using web developer tools to hack the page's HTML and intentionally follow himself). OTOH, if they want to go to all that trouble to follow themselves, there's no real harm done, so I don't think it's a real problem.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -94,9 +94,11 @@ class Ability
       cannot :destroy, OrderItem, :order => { :member_id => member.id, :completed_at => nil }
 
       # following/unfollowing permissions
-      can :create, Follow do |f|
-        !member.already_following?(f.followed) && f.followed_id != member.id
-      end
+      can :create, Follow
+      cannot :create, Follow, :followed_id => member.id # can't follow yourself
+
+      can :destroy, Follow
+      cannot :destroy, Follow, :followed_id => member.id # can't unfollow yourself
 
       if member.has_role? :admin
 
--- a/app/controllers/follows_controller.rb
+++ b/app/controllers/follows_controller.rb
@@ -1,8 +1,10 @@
 class FollowsController < ApplicationController
+  before_filter :authenticate_member!
+  load_and_authorize_resource
 
   # POST /follows
   def create
-    
+
     @follow = current_member.follows.build(:followed_id => follow_params[:followed_id])
 
     if @follow.save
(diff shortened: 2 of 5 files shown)


app/models/ability.rb
```diff
@@ -94,9 +94,11 @@ def initialize(member)
       cannot :destroy, OrderItem, :order => { :member_id => member.id, :completed_at => nil }
 
       # following/unfollowing permissions
-      can :create, Follow do |f|
-        !member.already_following?(f.followed) && f.followed_id != member.id
-      end
+      can :create, Follow
+      cannot :create, Follow, :followed_id => member.id # can't follow yourself
```
cesy: What will now happen if you type in the address directly and try to follow someone you're already following?

Skud: You can't do it by typing in the address in your URL bar, as it requires 
a POST.  You have to use web dev tools to make the request.  If you do, 
for some reason, then the behaviour is undefined in the classic nasal 
demons (http://catb.org/jargon/html/N/nasal-demons.html) sense, but the 
worst that would happen would be for a duplicate of the HABTM 
relationship to be created and the name to be listed twice on your 
members/whoever/following page.

There is literally no way to prevent this when defining an ability, as 
in the line:

     can :create, Follow do |f|

... that "f" is never actually passed in, and the CanCan docs say that 
it will always return true.  I guess we could put some additional logic 
in the controller "create" method to return an error message if you play 
silly buggers with it.  But that then means our authorization checking 
is scattered across three different places in the code. Sigh.

A.

On 19/01/2015 11:41 pm, Cesy wrote:
>
> In app/models/ability.rb 
> <https://github.com/Growstuff/growstuff/pull/647#discussion-diff-23157734>:
>
> > @@ -94,9 +94,11 @@ def initialize(member)
> >        cannot :destroy, OrderItem, :order => { :member_id => member.id, :completed_at => nil }
> >
> >        # following/unfollowing permissions
> > -      can :create, Follow do |f|
> > -        !member.already_following?(f.followed) && f.followed_id != member.id
> > -      end
> > +      can :create, Follow
> > +      cannot :create, Follow, :followed_id => member.id # can't follow yourself
>
> What will now happen if you type in the address directly and try to 
> follow someone you're already following?
>
> —
> <https://github.com/Growstuff/growstuff/pull/647/files#r23157734>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

Issue #566, opened 2014-12-12
Refactor: use .size for collections


Skud:
For a long time we didn't know that @member.plantings.size was more efficient than @member.plantings.count or @member.plantings.length or whatever.  There are lots of examples of bad usage throughout the code.  Let's clean them up.

PR #648, opened 2015-01-16
Work In Progress - PT80956846 crop search

oshiho3:
Please review and give me any feedback. By the way, this code should basically work.

<strong>Notes</strong> 
1) Crop search now searches against not just crop names but also scientific names and alternative names.

2) Modified the search result page so the exact match and partial match all comes under the same section. The exact matches will still be listed at the top.

3) Kept auto-suggest to use the current search method. This is because the drop down list for auto-suggest  only caters for crop names currently but not scientific names or alternative names. To avoid user confusion, I thought it's better to keep the current implementation until the UI is updated as well. A side benefit for keeping the existing search for auto-suggest is a better performance.

<strong>In order to try this on your machine, please follow the below steps in addition to the normal procedure.</strong>

1. Install and start Elasticsearch.
    http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/setup.html

2. Create index and import Crop model data into the index by running below commands
    <code>$ rails c</code>
    <code>> Crop.\__elasticsearch\__.create_index! force: true</code>
    <code>\> Crop.import</code>


pozorvlak:
Awesome! I can't test from here, unfortunately, but it looks good. Some
minor points (almost all about comments, as it turns out):

- Are those gems all needed explicitly, or do some imply others?
- We've suffered in the past from "what the heck is that in there for?"
Gemfile bloat; it's good that you've added a comment explaining what those
gems do, but it would be even better if it said *why* we're using
ElasticSearch. Maybe something like "ElasticSearch client, used for fuzzy
crop search".
- I tripped over the assign-and-test in the controller; I think that would
be better as two statements.
- Is it possible to explain the tokenizing strategy in a sentence or two of
human language? My guess is "treat alphanumeric sequences as tokens and
ignore case", but I'd need to read the ElasticSearch docs to be sure. A
comment containing such an explanation would be great.
- You can remove the "currently we're using SQL ILIKE but may do something
more fancy in the future" comment - we are now doing something more fancy
:-)
- Your explanation of why autosuggest and crop matching use different rules
makes sense, but IMHO should live in the code as a comment on the
autosuggest method. Also I think we should try to roll out clever search on
autosuggest soonish, because I can see that being confusing for new users.

tygriffin:
This looks super! 

One thing you'll definitely want to do is pull from upstream because you've got some methods in the code that are depricated in Rails 4. I'll point them out inline.

Also, we're no longer using view tests, so please write a feature test instead.

app/controllers/crops_controller.rb
```diff
@@ -51,18 +51,16 @@ def hierarchy
...
-    @search = params[:search]
-    @exact_match = Crop.find_by_name(params[:search])
+    @all_matches = Crop.search(params[:search])
 
-    @partial_matches = Crop.search(params[:search])
-    # exclude exact match from partial match list
-    @partial_matches.reject!{ |r| @exact_match && r.eql?(@exact_match) }
-
-    @fuzzy = Crop.search(params[:term])
+    if exact_match = Crop.find_by_name(params[:search])
```
tygriffin: For exact matches, we can probably use [Elasticsearch filters](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-filters.html), which should perform better than filtering out results in Rails.

app/models/crop.rb
```diff
@@ -33,6 +33,47 @@ class Crop < ActiveRecord::Base
...
+      },
+      analyzer: {
+        gs_edgeNGram_analyzer: {
+          tokenizer: "gs_edgeNGram_tokenizer",
+          filter: ["lowercase"]
+        }
+      },
+    } do
+    mappings dynamic: 'false' do
+      indexes :id, type: 'long'
```
tygriffin: I would include `created_at`, `updated_at`, `slug`, and `creator_id` in the index as well, since we will almost certainly want to search against those in fancier searches soon.

app/views/crops/search.html.haml
```diff
@@ -1,6 +1,6 @@
 - content_for :title, "Crops matching #{@search}"
 
-- unless (@exact_match || @partial_matches.length > 0)
+- unless (@all_matches.length > 0)
```
tygriffin: Maybe we can refactor this into a simpler conditional. `if @all_matches` ... `else`

spec/models/crop_spec.rb
```diff
@@ -328,6 +328,8 @@
   context "search" do
     before :each do
       @mushroom = FactoryGirl.create(:crop, :name => 'mushroom')
+      @mushroom.__elasticsearch__.index_document
+      sleep 1
```
tygriffin: This sleep is here to manage the asynchronicity of the indexing, right? Might be better to handle this without `sleep`, because this can sometimes cause spuriously failing tests. I'm not sure if this is possible without doing a little research. If it turns out to be a big hassle then we can probably live with this since indexing one record should be super fast.

pozorvlak: Good spot! Yes, definitely try to make this test run synchronously, or put
the assertion in a callback, or something.

Skud:
Big picture thoughts before I even look at the code...

1. This is yet another thing for developers to download and install.  I'm wondering whether we can make it optional in the dev environment, and fall back to the SQL search if it isn't installed?  Then only people who are working with the search feature would need to install it -- same as how you don't need eg. Flickr API keys unless you're working with the flickr stuff.

2. How does Travis test elasticsearch? (Oh, ok, you just do this: http://docs.travis-ci.com/user/database-setup/#ElasticSearch)

3. What is needed to get elasticsearch running on Heroku?  I know @oshiho3 @tygriffin and I looked at the plan options on heroku.com (it looks like the free tier will be fine for us) but I'd like a bit clearer documentation about what the steps are.

4. We *definitely* want to use elasticsearch for the autosuggest dropdown.  And we definitely need to make those UI changes ASAP.  However, just on the topic of performance for now... what is the actual performance hit with elasticsearch? How long does it take to respond to each keystroke in the autosuggest dropdown?  Can we mitigate this at all by not triggering a search until a minimum number of characters (say 3-5) have been typed?

app/models/crop.rb
Skud: @tygriffin what do you think we'd use those for in crop searches? what's the use case?

app/models/crop.rb
```diff
@@ -206,7 +247,22 @@ def add_alternate_names_from_csv(alternate_names)
   # searches for crops whose names match the string given
   # just uses SQL LIKE for now, but can be made fancier later
   def self.search(query)
-    where("name ILIKE ?", "%#{query}%")
+    search_str = query.nil? ? "" : "#{query.downcase}"
```
Skud: query.downcase shouldn't need quoting here, should it?

app/models/crop.rb
```diff
@@ -206,7 +247,22 @@ def add_alternate_names_from_csv(alternate_names)
...
+    response = __elasticsearch__.search(
+      {
```
Skud: please put this brace on the line above

app/models/crop.rb
```diff
@@ -206,7 +247,22 @@ def add_alternate_names_from_csv(alternate_names)
...
+        query: {
+          multi_match: {
+            query: search_str,
+            fields: ["name", "scientific_names.scientific_name", "alternate_names.name"]
+          }
+        },
+        size: 50
+      }
+    )
+    response.records.to_a
```
Skud: it might be just me, but i prefer to explicitly use the "return" keyword for lines like this, just so it's more readable

app/views/crops/search.html.haml
```diff
@@ -1,6 +1,6 @@
 - content_for :title, "Crops matching #{@search}"
 
-- unless (@exact_match || @partial_matches.length > 0)
```
Skud: please use size not length. see #566

app/views/crops/search.html.haml
```diff
@@ -9,18 +9,9 @@
...
-    %h2 Exact match
-
-    .row
-      .col-md-2.six-across
-        = render :partial => "thumbnail", :locals => { :crop => @exact_match }
-
-- if ! @partial_matches.empty?
-  %div#partial_matches
-    %h2 Partial matches
+- if ! @all_matches.empty?
```
Skud: please use "unless" rather than "if !"

spec/views/crops/search.html.haml_spec.rb
```diff
@@ -11,19 +11,18 @@
...
     it "shows exact matches" do
-      assert_select "div#exact_match" do
+      assert_select "div#all_matches" do
         assert_select "a[href=#{crop_path(@tomato)}]"
       end
     end
 
-    it "shows partial matches" do
-      assert_select "div#partial_matches" do
+    it "shows all matches" do
```
Skud: the issue here is still that it shows partial matches. i think the text as it previously stood is clearer. after all, we're not testing that it shows *all* the matches, are we? just that it shows at least one partial match.

Skud:
OK, this is looking great overall :)  I'm definitely keen to include alternate/scientific names in the autosuggest UI though.  I will open a new story for that and perhaps we can play with sketching/wireframing what that might look like.

Skud:
I opened #649 to discuss elasticsearch and the autosuggest dropdown.

app/models/crop.rb
tygriffin: @Skud That's a good point. Actually, now that I think about it, there really isn't a use case, is there? I think I had plantings and crops mixed up in my brain. So yeah, no need to index more than is already here. Plus, I guess the number of crops will only ever be in the thousands, so it's easy to re-index the whole thing if we wanted to search on more criteria.

One thing just occured to me, @oshiho3 : When a crop is edited or

tygriffin:
@Skud :
> [...] just on the topic of performance for now... what is the actual performance hit with elasticsearch? How long does it take to respond to each keystroke in the autosuggest dropdown? Can we mitigate this at all by not triggering a search until a minimum number of characters (say 3-5) have been typed?

One possibility is to have the Javascript hit Elasticsearch directly without going through Rails. I think we should figure out soon what JS should and shouldn't be doing in Growstuff and how to organize / document it.

oshiho3:
Taylor! do you have an hour or so today? I want your help to progress this
fast. I sent you SMS as well, sorry if I annoy you! lol

tygriffin:
Hey Shiho! Of course, you're not annoying me! :) I was just out entertaining an unexpected visitor from Tokyo, so I just noticed your SMS. 

I probably don't have any time though today, because we need to pack. Talk about it tomorrow over lunch?

oshiho3:
OK. No problem. Enjoy the sunshine!

app/models/crop.rb
oshiho3: quoting was needed in the hash, so I'm moving the quite in the hash to make it clearer.

app/models/crop.rb
oshiho3: done

app/models/crop.rb
oshiho3: agreed. done

app/views/crops/search.html.haml
oshiho3: thanks for the info. no prob

app/views/crops/search.html.haml
oshiho3: good point. I'll do that

spec/models/crop_spec.rb
oshiho3: Yeah, I didn't like the <code>sleep</code> either. Thanks both of you for those comments, they reinforced my hate so much that I fixed it! Replacing the sleep with <code>Crop.__elasticsearch__.refresh_index!</code> has worked.

I'm also adding some code under spec/support so it will recreate/clean index before each test.

oshiho3:
pozorvlak, thanks for the comment. Great points!

- Are those gems all needed explicitly, or do some imply others?
> They are all needed explicitely; elasticsarch-rails doesn't automatically
include elasticsearch-model interestingly. bonsai-elasticsearch-rails is
specifically needed for Heroku so I'm moving it under prod section.

- We've suffered in the past from "what the heck is that gem in there for?"
> Make sense, I'm adding better comment

- I tripped over the assign-and-test in the controller; I think that would be
better as two statements.
> Appreciate you picked this up. I'm all for readability and will update it
as you suggested.

- Is it possible to explain the tokenizing strategy in a sentence or
two of human
language?
> Of course! I'll meet Taylor today and am meaning to discuss setting for
the indexes with him before documenting the strategy. Elasticsearh can be
so flexible that I like to have another brain to ensure it brings up
reasonable search results for users before other people start using/testing
it.

- You can remove the "currently we're using SQL ILIKE but may do something
more fancy in the future" comment.
> Nice pick :) Getting rid of them.

- Your explanation of why autosuggest and crop matching use different rules
makes sense, but IMHO should live in the code as a comment on the
autosuggest method.
> Make sense. I'm adding the comment.

oshiho3:
1. I'm wondering whether we can make it optional in the dev environment,
and fall back to the SQL search if it isn't installed?
> Makes sense. Working on it.

2. How does Travis test elasticsearch?
> Updating .travis.yml. Test passed.

3. What is needed to get elasticsearch running on Heroku?
> From what I read in below article, it should work by just running <code>heroku
addons:add bonsai</code>command.
https://devcenter.heroku.com/articles/bonsai#installing-the-add-on
I meant to try this myself first but I'm having difficulty pushing the code
to my heroku. It may take a while for me to figure this out, so I'm
wondering if someone can try it for me.

4. We *definitely* want to use elasticsearch for the autosuggest dropdown.
> OK, I've tweaked the elasticsearch option and am updating the code to use
it for auto-suggest as well. The performance on my local environment seems
reasonable.

oshiho3:
Made it so that the default for dev and test environment uses SQL search. The option can be turned on by setting GROWSTUFF_ELASTICSEARCH to "true" in Application.yml

Skud:
Looking good!

One thing I'm wondering, @oshiho3, is what's going on with your commits.  I'm seeing lots of these "merge" commits which shouldn't be necessary... what's causing them?

Also, most of your actual commits are doubled (eg. https://github.com/oshiho3/growstuff/commit/d12ec968c41233a51aec6307cb901400675eff44 and https://github.com/oshiho3/growstuff/commit/018b2b4711a8aa737deb7ca69f14aaa595f41440 which have the same commit message) but those commits seem to include multiple things, which the commit messages don't explain (eg. those ones include not only turning off elasticsearch in development, but also a bunch of other work).

http://wiki.growstuff.org/index.php/Development/Coding_session (especially [committing your work](http://wiki.growstuff.org/index.php/Development/Coding_session#Committing_your_work) and [finishing your coding session](http://wiki.growstuff.org/index.php/Development/Coding_session#Committing_your_work) describes the usual process, in terms of what merges/commits/pushes you'd normally be making.  As for merging in the latest stuff from dev, I would usually only bother doing that right near the end of your work, unless you think you're likely to be working at cross-purposes with someone else and touching the same code (which you're not in this case).

oshiho3:
I suspect it's probably caused that I keep forgetting pull before pushing,
and also me being careless about comment :( I'll try better in future.

pozorvlak:
Ah, are you developing this branch on several different machines? I find `git pull --rebase` useful in those situations; it pulls down the commits from the remote branch and then replays your local commits on top of them, which results in no merge commits. But yeah, if you can remember to pull at the start of your coding session it makes life easier all-round.

Clear commit messages are massively helpful to code reviewers, but I also find them helpful to me while I'm developing: writing a specific commit message helps me to keep track of what I'm doing at any given time and why. I have ADHD, though, so YMMV :-)

oshiho3:
Thanks for the info pozorvlak! I also think I've been using <code>git commit --amend</code> when I shouldn't have been or didn't update the comment properly for it. Yeah, I sometimes forget what I was thinking or doing a few seconds before and it scares me lol

pozorvlak:
I think the only thing now missing is a comment explaining the tokenisation strategy. It would be good if you could squash the history into a more meaningful set of commits, but that's IMHO not a blocker.

Skud:
This is looking good to me. Are we ready to merge and test, do you think?

oshiho3:
OK, I added a few more feature tests and found that I need a little more code to trigger indexing when scientific or alternate name has been added/updated/removed etc. I made it to work in dev when tested manually, but unfortunately rspec tests fail. It seems to do with after_commit callback not being called in rspec.

I tried this gem called test_after_commit, but a few other existing tests ended up failing :(

In summary, give me a couple of more days in order to sort this out and I'll be happy to hand over the code!

Skud:
It's probably worth noting that the crops database isn't very fast changing. If something isn't reindexed immediately it's not going to have much of an effect at least for now.  But it seems like it works anyway, right?  It's just the testing that's having problems?  Please don't spend *too* long on this -- if it's really hard to test, just leave a comment somewhere in amongst the relevant tests saying that we can't do it because X and Y reasons, and we can revisit it later.

oshiho3:
I'm handing over the code finally.

To install elasticsearch on Heroku,
<code>$heroku addons:add bonsai</code>

Then to create index for existing database, execute
<code>rake growstuff:oneoff:elasticsearch_create_index</code>

Skud:
@oshiho3 wrote:

> Then to create index for existing database, execute
rake growstuff:oneoff:elasticsearch_create_index

I see this is included in scripts/deploy-tasks.sh which is great.  I'm just wondering whether we should include this somewhere in the process that is run when you create a new app instance, eg. in development or when deploying to a nonce heroku instance.  Is there documentation for setting up elasticsearch in your dev environment?  Can we add that to the wiki?

This is not a blocker on the pull request however. I think we're ready to merge, at long last!

oshiho3:
I know I let this one sleep for too long :(
Here are the steps if anyone want to test elasticsearch in dev.

1. Install and start Elasticsearch.
http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/setup.html

2. Then to create index for existing database, execute
<code>rake growstuff:oneoff:elasticsearch_create_index</code>

3. In application.yml, update <code>GROWSTUFF_ELASTICSEARCH</code> option under the default section to <code>true</code>
<code>GROWSTUFF_ELASTICSEARCH: "true"</code>

Issue #491, opened 2014-12-12
Set up continuous deployment to staging


Skud:
In order to ease testing of the staging environment, we should enable continuous deployment to staging through Travis CI.

PR #659, opened 2015-01-19
Deployment from Travis builds

pozorvlak:
A stab at #491. If you click "merge" on this PR then it *should* deploy recent changes to growstuff-staging and run migrations and the current deploy-tasks.sh. I haven't tested it on the dev branch, but [here's an example of it successfully deploying to my sandbox app](https://travis-ci.org/Growstuff/growstuff/builds/47582713).

This doesn't totally meet the spec, unfortunately:

 - it doesn't turn maintenance mode on for the duration of deployments (see https://travis-ci.org/Growstuff/growstuff/builds/47577519 and https://travis-ci.org/Growstuff/growstuff/builds/47580792 for two failed attempts at doing this). http://xseignard.github.io/2013/02/18/continuous-deployement-with-github-travis-and-heroku-for-node.js/ suggests a possible approach to turning on maintenance mode, but that looks unpleasantly hand-rolled.
 - it doesn't set environment variables on the heroku dynos (the `figaro` step in [the release management guidelines](http://wiki.growstuff.org/index.php/Release_management#Deploying_new_features_to_staging)), because I'm not sure how to expose them to Travis securely. Perhaps we could use [Travis' secure variables](http://docs.travis-ci.com/user/environment-variables/#Secure-Variables)? Not sure how that would work, TBH.

I've left the code in for deploying the `travis_deploy` branch to http://tranquil-basin-3130.herokuapp.com, which is an instance of Growstuff running on my Heroku account. We'll want to remove it eventually, but I suspect we'll need a deployment sandbox while we work through the issues on this PR :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ deploy:
   app: tranquil-basin-3130
   on:
     repo: Growstuff/growstuff
-    branch: travis_deploy
+  app:
+    master: growstuff-prod
+    dev: growstuff-staging
+    travis_deploy: tranquil-basin-3130
   run:
     - "rake db:migrate"
     - "script/deploy-tasks.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,18 @@
----
 language: ruby
-
 env: GROWSTUFF_SITE_NAME="Growstuff (travis)" RAILS_SECRET_TOKEN='xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
-bundler_args: --without development production staging
+bundler_args: "--without development production staging"
 rvm:
-  - 2.1.5
+- 2.1.5
 before_script:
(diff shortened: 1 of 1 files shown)


tygriffin:
1. Being able to turn maintenance mode on would be nice, but indeed looks like a headache. I wouldn't worry about it.

2. Looks like Travis Secure Variables wouldn't work with PR's coming from a fork. This means that someone with greater privileges must manually do the deploy when something in the "secrets" (soon to be secrets.yml, I believe) changes just like we're doing now, which seems right to me. I'd leave this one be as well.

Skud:
Maintenance mode doesn't really matter for staging, but I would definitely like to use it on production.

pozorvlak:
Oops, looks like I opened this from the wrong place. I'll close it and re-open it from my fork.

pozorvlak:
It would be nice if we could at least add an automatic "has the list of
environment variables changed" check.

PR #661, opened 2015-01-20
Travis deploy

pozorvlak:
Supersedes #659. As I said on that PR:

A stab at #491. If you click "merge" on this PR then it *should* deploy recent changes to growstuff-staging and run migrations and the current deploy-tasks.sh. I haven't tested it on the dev branch, but [here's an example of it successfully deploying to my sandbox app](https://travis-ci.org/Growstuff/growstuff/builds/47582713).

This doesn't totally meet the spec, unfortunately:

 - it doesn't turn maintenance mode on for the duration of deployments (see https://travis-ci.org/Growstuff/growstuff/builds/47577519 and https://travis-ci.org/Growstuff/growstuff/builds/47580792 for two failed attempts at doing this). http://xseignard.github.io/2013/02/18/continuous-deployement-with-github-travis-and-heroku-for-node.js/ suggests a possible approach to turning on maintenance mode, but that looks unpleasantly hand-rolled.
 - it doesn't set environment variables on the heroku dynos (the `figaro` step in [the release management guidelines](http://wiki.growstuff.org/index.php/Release_management#Deploying_new_features_to_staging)), because I'm not sure how to expose them to Travis securely. Perhaps we could use [Travis' secure variables](http://docs.travis-ci.com/user/environment-variables/#Secure-Variables)? Not sure how that would work, TBH.

I've left the code in for deploying the `travis_deploy` branch to http://tranquil-basin-3130.herokuapp.com, which is an instance of Growstuff running on my Heroku account. We'll want to remove it eventually, but I suspect we'll need a deployment sandbox while we work through the issues on this PR :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ deploy:
   app: tranquil-basin-3130
   on:
     repo: Growstuff/growstuff
-    branch: travis_deploy
+  app:
+    master: growstuff-prod
+    dev: growstuff-staging
+    travis_deploy: tranquil-basin-3130
   run:
     - "rake db:migrate"
     - "script/deploy-tasks.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,18 @@
----
 language: ruby
-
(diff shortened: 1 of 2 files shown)


pozorvlak:
Probably superseded by #664.

PR #664, opened 2015-01-21
Deployment from Travis containers

pozorvlak:
This is (I think) an improvement over #661 in two ways:
 - it caches the results of `bundle install`, speeding up builds by a factor of around 2.5
 - it doesn't install the whole Heroku toolbelt, which felt like overkill.

It turns maintenance mode on for deployment (and off again afterwards!), but doesn't try to update the environment variables on the Heroku dyno. It also duplicates the branch => app name logic, which is ugly. Perhaps we could parse .travis.yml in `heroku_maintenance.rb`? **Edit** fixed in 5755245.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ deploy:
   app: tranquil-basin-3130
   on:
     repo: Growstuff/growstuff
-    branch: travis_deploy
+  app:
+    master: growstuff-prod
+    dev: growstuff-staging
+    travis_deploy: tranquil-basin-3130
   run:
     - "rake db:migrate"
     - "script/deploy-tasks.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,18 @@
----
 language: ruby
-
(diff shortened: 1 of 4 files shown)


Skud:
Should #661 be closed?

pozorvlak:
If you agree that this approach is better, then yes. I wanted to allow
reviewers to compare the two.

Skud:
Ah, ok! I didn't actually have a firm opinion, wasn't quite sure what 
was going on.  I think this is probably better though.

On 21/01/2015 7:07 pm, pozorvlak wrote:
> If you agree that this approach is better, then yes. I wanted to allow
> reviewers to compare the two.
>
> —
> <https://github.com/Growstuff/growstuff/pull/664#issuecomment-70798909>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

pozorvlak:
The tradeoff was "not reinventing the wheel" versus "faster builds", but actually, it's a pretty tiny wheel I've reinvented here. It would have been nice if we could have done this via the official Travis deploy infrastructure, but I can't see any way of doing that.

Skud:
@pozorvlak, if you're happy with this as it stands, then it LGTM and we can merge.

PR #662, opened 2015-01-20
WIP: Issue #468 deletion

cesy:
Draft, working towards fixing issue #468. Please do not yet merge, just leaving it here for easy reference and any early review of the test cases.

PR #665, opened 2015-01-21
Fix the 500 server error when you try to follow another user

pozorvlak:
See discussion at http://talk.growstuff.org/t/please-test-footer-height-authorisation-on-follows

### DIFF
--- a/app/controllers/follows_controller.rb
+++ b/app/controllers/follows_controller.rb
@@ -1,6 +1,7 @@
 class FollowsController < ApplicationController
   before_filter :authenticate_member!
   load_and_authorize_resource
+  skip_load_resource :only => :create
 
   # POST /follows
   def create
--- a/spec/features/following_spec.rb
+++ b/spec/features/following_spec.rb
@@ -47,6 +47,13 @@ feature "follows" do
         expect(page).to have_content "#{other_member.login_name}"
       end
 
+      scenario "does not die when passed an authenticity_token" do
+        visit member_follows_path(
+          member,
+          :params => {:authenticity_token => "Ultima ratio regum"})
+        expect(page.status_code).to equal 200
+      end
+
       scenario "has correct message and follow button after unfollow" do
         click_link 'Follow'
         click_link 'Unfollow'
--- a/spec/features/following_spec.rb
+++ b/spec/features/following_spec.rb
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-feature "follows" do
+feature "follows", :js => true do
 
   context "when signed out" do
     let(:member) { FactoryGirl.create(:member) }


pozorvlak:
OK, turns out this *doesn't* actually fix the problem. Bah. Back to the drawing board...

pozorvlak:
More precisely, it fixes the `ActiveRecord::UnknownAttributeError (unknown attribute: authenticity_token)`, caused by passing the XSRF-defeater through to the Follow model's constructor, but introduces an `found unpermitted parameters: authenticity_token` error in `FollowController#follow_params`.

Skud:
The bit I'm wondering about is how our feature tests passed at all, with 
this problem?

On 22/01/2015 8:16 am, pozorvlak wrote:
>
> More precisely, it fixes the |ActiveRecord::UnknownAttributeError 
> (unknown attribute: authenticity_token)|, caused by passing the 
> XSRF-defeater through to the Follow model's constructor, but 
> introduces an |found unpermitted parameters: authenticity_token| error 
> in |FollowController#follow_params|.
>
> —
> <https://github.com/Growstuff/growstuff/pull/665#issuecomment-70923218>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

pozorvlak:
@Skud: yes, that is worrying. I had a look at where the `authenticity_token` came from, and couldn't work it out - I think it's being generated somewhere deep within the guts of a rails/jquery event handler. But PhantomJS ought to be running that handler. Most mysterious.

pozorvlak:
Aha, @tygriffin [has the answer](https://github.com/Growstuff/growstuff/pull/662/files#r23291413) - we need to use give Capybara the `:js => true` option to use PhantomJS. Doing this makes most of the feature tests for Follows fail! Unfortunately it makes them about 3x slower :-(

spec/controllers/follows_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
```
cesy: I thought we were just doing feature tests these days, not the others? Or was it just view tests that we decided to stop doing?

pozorvlak: Yeah, I wrote a controller test because I couldn't write a failing feature test. OTOH, now I've managed to make a feature test fail, I can rip this out. Thanks for the reminder!

pozorvlak: I have now removed the controller test.

Skud:
LGTM. @pozorvlak if this is good to go, please merge. (I'm commenting in this form because there have been so many WIPs/experiments lately that I'm no longer quite sure what's what!)

PR #667, opened 2015-01-22
More tests

cesy:
I can't find an issue to link it to, but this should increase our coverage slightly with some useful tests. Coveralls shows an increase of 12 hits, too small to impact the percentage. I have now figured out the right bit of Coveralls to add stuff that it thinks is missing, and will do more prioritised ones now.


cesy:
I have a mix of single and double-quotes here. What's our official coding standard on that?

pozorvlak:
I don't think we have one (and there [doesn't appear to be a widespread community standard](https://github.com/bbatsov/ruby-style-guide), but we probably ought to.

tygriffin:
I generally prefer double quotes in any language with interpolation like ruby, eg. "#{ruby}"

iPhone$B$+$iAw?.(B

2015/01/23 0:31$B!"(BCesy <notifications@github.com> $B$N%a%C%;!<%8(B:

Issue #666, opened 2015-01-22
Adding a new crop without scientific name gives an uncaught error


cesy:
If a crop wrangler adds a new crop and only fills in the name, not the scientific names (contrary to our wrangling guidelines, but possible), instead of being caught smoothly, you get an ActiveRecord error.

ActionController::UnpermittedParameters in CropsController#create
found unpermitted parameters: scientific_names_attributes

Skud:
FWIW, this error doesn't occur on staging (or production) at present.

On 22/01/2015 9:39 pm, Cesy wrote:
>
> If a crop wrangler adds a new crop and only fills in the name, not the 
> scientific names (contrary to our wrangling guidelines, but possible), 
> instead of being caught smoothly, you get an ActiveRecord error.
>
> ActionController::UnpermittedParameters in CropsController#create
> found unpermitted parameters: scientific_names_attributes
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/Growstuff/growstuff/issues/666>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

tygriffin:
I know exactly what this is. We need to pass a hash of permitted params for scientific names when setting permitted params for crops. I'll fix when I'm at my computer in a couple hours.

iPhone$B$+$iAw?.(B

2015/01/22 20:39$B!"(BCesy <notifications@github.com> $B$N%a%C%;!<%8(B:

tygriffin:
@Skud That's probably because permitted params are configed to only raise exceptions in dev. Should probably make it raise in staging too, I think.

iPhone$B$+$iAw?.(B

2015/01/22 20:43$B!"(BSkud <notifications@github.com> $B$N%a%C%;!<%8(B:

PR #669, opened 2015-01-22
Sci names permitted params

tygriffin:
Fixes [666](https://github.com/Growstuff/growstuff/issues/666#issuecomment-71002122)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20141119130555) do
+ActiveRecord::Schema.define(version: 20141209074953) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


pozorvlak:
Hmmm, you seem to have accidentally committed some unrelated changes to db/schema.rb. I *think* my merging it was harmless...

db/schema.rb
```diff
@@ -116,7 +117,7 @@
     t.string   "area_unit"
   end
 
-  add_index "gardens", ["owner_id"], name: "index_gardens_on_owner_id", using: :btree
+  add_index "gardens", ["owner_id"], name: "index_gardens_on_user_id", using: :btree
```
Skud: Err, we don't use "user" anywhere. We say "member" and the actual field here is owner_id. What is going on here?

db/schema.rb
```diff
@@ -175,11 +176,11 @@
     t.boolean  "send_planting_reminder",  default: true
   end
 
-  add_index "members", ["confirmation_token"], name: "index_members_on_confirmation_token", unique: true, using: :btree
-  add_index "members", ["email"], name: "index_members_on_email", unique: true, using: :btree
-  add_index "members", ["reset_password_token"], name: "index_members_on_reset_password_token", unique: true, using: :btree
-  add_index "members", ["slug"], name: "index_members_on_slug", unique: true, using: :btree
-  add_index "members", ["unlock_token"], name: "index_members_on_unlock_token", unique: true, using: :btree
+  add_index "members", ["confirmation_token"], name: "index_users_on_confirmation_token", unique: true, using: :btree
```
Skud: Again with the "user" thing. We don't use that terminology. What is the intent of this change?

db/schema.rb
```diff
@@ -20,24 +20,24 @@
     t.string   "name",              null: false
     t.boolean  "is_paid"
     t.boolean  "is_permanent_paid"
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.datetime "created_at",        null: false
```
Skud: What caused all these schema changes?

tygriffin: No clue. Just did a quick search on "user" in the migrations and nothing comes up before 2013 when the users table was renamed to members.

PR #670, opened 2015-01-22
WIP: Turn on PhantomJS for all feature tests

pozorvlak:
Since turning on PhantomJS made the bug in the FollowsController apparent (see #665), I thought it might be interesting to turn it on everywhere and see what happened. The result (on my machine) is 17 test failures (of which six should be fixed by #665) and an approximate 3x slowdown on the tests in spec/features.


Skud:
I think this is an excellent idea and my only question is, is there a way to do this at a config level rather than having to do it in each file?

Also, we should obviously fix all the breaking tests before merging this.

pozorvlak:
It [looks like](https://blog.codecentric.de/en/2013/08/cucumber-capybara-poltergeist/) we can set `Capybara.default_driver = :poltergeist` in our test environment.

I was actually thinking we could run all the feature tests with and without JavaScript, by setting the Capybara config from an environment variable and using Travis' build matrix. How many tests rely on JavaScript to be meaningful?

pozorvlak:
And yes, we should definitely fix the breaking tests before merging this :-) This is mostly up here for visibility. I'll mark it as a work-in-progress.

cesy:
If you can update this, I think it will help my deletion testing to see what's a genuine existing failure and what's not.

pozorvlak:
@cesy: done.

cesy:
This is still failing, @pozorvlak ?

pozorvlak:
Yeah, sorry, attacks of life. Someone needs to go through the test failures
and work out if they're bugs or problems in the test, and I haven't got
round to doing that yet.

cesy:
Superseded by #810.

PR #676, opened 2015-01-24
New seed properties: days until maturity, organic, GMO, heirloom

### LINKED ISSUE #493: Add "days until harvest" as a property of seeds
Since this often appears on seed packets, let's collect the information in our database.  It should be validated as a positive numeric value.

On the crop detail page, under how to grow it (eg. near the sun/shade etc), we should show the median(?) of this number based on all seeds for that crop. (Should be a separate story, and we should do it after we've gathered a bit of data.)

This figure should also be added to the crop CSV file (separate story: #677)

### LINKED ISSUE #523: Add organic, heirloom, and GMO attributes for seeds
When recording seeds in our database, we should allow people to select data on whether the seeds are organic, GMO, heirloom, etc.

There should be three dropdowns as follows:

Organic: 

    certified organic
    non-certified organic
    conventional/non-organic
    unknown

GMO: 

    certified GMO-free
    non-certified GMO-free
    GMO
    unknown

Heirloom or hybrid?

    heirloom
    hybrid
    unknown

All should default to "unknown".

This data should obviously be shown on the seeds/show page!  Don't bother modifying the index; there's already #518 which is probably a better way to show this stuff on the index page.

### LINKED ISSUE #614: Bug: can specify a negative number of seeds
I seem to be able to specify that I have a negative number of seeds.  Expected: this should require a number > 0 (or blank).  I believe a fix for this was recently made for plantings or harvests, so we should copy that across.

pozorvlak:
Fixes #493 
Fixes #523 
Fixes #614

### DIFF
--- a/spec/features/seeds/adding_seeds_spec.rb
+++ b/spec/features/seeds/adding_seeds_spec.rb
@@ -17,6 +17,10 @@ feature "Harvesting a crop", :js => true do
     within "form#new_seed" do
       fill_in "Quantity:", :with => 42
       fill_in "Plant before:", :with => "2014-06-15"
+      fill_in "Days until maturity:", :with => 999
+      select "certified organic", :from => "Organic?"
+      select "non-certified GMO-free", :from => "GMO?"
+      select "heirloom", :from => "Heirloom?"
       fill_in "Description", :with => "It's killer."
       select "internationally", :from => "Will trade:"
       click_button "Save"
--- a/spec/features/seeds/adding_seeds_spec.rb
+++ b/spec/features/seeds/adding_seeds_spec.rb
@@ -17,8 +17,8 @@ feature "Seeds", :js => true do
     within "form#new_seed" do
       fill_in "Quantity:", :with => 42
       fill_in "Plant before:", :with => "2014-06-15"
-      fill_in "Days until maturity, min:", :with => 999
-      fill_in "Days until maturity, max:", :with => 1999
+      fill_in "Days until maturity:", :with => 999
+      fill_in "to", :with => 1999
       select "certified organic", :from => "Organic?"
       select "non-certified GMO-free", :from => "GMO?"
       select "heirloom", :from => "Heirloom?"
--- a/spec/features/seeds/adding_seeds_spec.rb
+++ b/spec/features/seeds/adding_seeds_spec.rb
@@ -27,6 +27,12 @@ feature "Harvesting a crop", :js => true do
     end
 
     expect(page).to have_content "Successfully added maize seed to your stash"
+    expect(page).to have_content "Quantity: 42"
+    expect(page).to have_content "Days until maturity: 999"
+    expect(page).to have_content "certified organic"
+    expect(page).to have_content "non-certified GMO-free"
+    expect(page).to have_content "Heirloom? heirloom"
+    expect(page).to have_content "It's killer."
   end
 
   scenario "Adding a seed from crop page" do
--- a/spec/features/seeds/adding_seeds_spec.rb
+++ b/spec/features/seeds/adding_seeds_spec.rb
@@ -29,7 +29,7 @@ feature "Harvesting a crop", :js => true do
 
     expect(page).to have_content "Successfully added maize seed to your stash"
     expect(page).to have_content "Quantity: 42"
-    expect(page).to have_content "Days until maturity: 999"
+    expect(page).to have_content "Days until maturity: 999–1999"
     expect(page).to have_content "certified organic"
     expect(page).to have_content "non-certified GMO-free"
     expect(page).to have_content "Heirloom? heirloom"
--- a/spec/features/seeds/adding_seeds_spec.rb
+++ b/spec/features/seeds/adding_seeds_spec.rb
@@ -17,7 +17,8 @@ feature "Harvesting a crop", :js => true do
     within "form#new_seed" do
       fill_in "Quantity:", :with => 42
       fill_in "Plant before:", :with => "2014-06-15"
-      fill_in "Days until maturity:", :with => 999
+      fill_in "Days until maturity, min:", :with => 999
+      fill_in "Days until maturity, max:", :with => 1999
       select "certified organic", :from => "Organic?"
       select "non-certified GMO-free", :from => "GMO?"
       select "heirloom", :from => "Heirloom?"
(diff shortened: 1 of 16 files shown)


Skud:
Re: #493 something just occurred to me.  I've been looking at actual seed packets a bit lately and they tend to actually have a range for days to maturity, eg. 55-65 days.  Should we have days_to_maturity_min and days_to_maturity_max?

pozorvlak:
Good idea!

pozorvlak:
I've split the days_until_maturity field into two. I'm not terribly happy with how the form looks (see attached screenshot) , and would appreciate input from someone more frontendy!
screenshot from 2015-01-26 19 09 46

Skud:
@pozorvlak this looks good. Wrt the form, I'd simply do "days until maturity: \[    ] to \[    ] days".  We should probably also validate that max > min.

Can someone like @maco @tygriffin @oshiho3 @cesy look over this code, too, and give it a review? We need someone who's not me or Miles to say it looks OK before merging.

spec/features/seeds/adding_seeds_spec.rb
```diff
@@ -17,12 +17,23 @@
...
+      select "non-certified GMO-free", :from => "GMO?"
+      select "heirloom", :from => "Heirloom?"
       fill_in "Description", :with => "It's killer."
       select "internationally", :from => "Will trade:"
       click_button "Save"
     end
 
     expect(page).to have_content "Successfully added maize seed to your stash"
+    expect(page).to have_content "Quantity: 42"
+    expect(page).to have_content "Days until maturity: 999"
```
maco: Should this check for it displaying as a date range, since there's a partial that does different things depending on whether both min & max are filled in?

spec/rails_helper.rb
```diff
@@ -25,7 +25,7 @@
 require 'capybara/poltergeist'
 Capybara.javascript_driver = :poltergeist
 Capybara.app_host = 'http://localhost'
-Capybara.server_port = 8080
+Capybara.server_port = 8081
```
maco: This looks like it's part of something else besides seed properties

spec/features/seeds/adding_seeds_spec.rb
Skud: Err, no, because it's not a date, it's an integer.  You know on a seed packet it says "plant in full sun, early spring to late summer, days to maturity 55-65" or something?  It's that.

spec/rails_helper.rb
Skud: It's just a "fix up tests that are behaving annoyingly" thing. It was trying to run them on the same port as we typically run unicorn for our dev server, which caused weird behaviour at times.

spec/features/seeds/adding_seeds_spec.rb
maco: Sorry, shouldn't have said "date range" but an integer range? It's only testing for "days to maturity 55" not "days to maturity 55-65", see? But the partial he wrote displays only 1 number when min or max is blank and displays "x-y" when they're both filled in.

Skud: Ah, I see what you're getting at! Yes, we should test for the cases where min, max, min or max, or neither are set.

pozorvlak: Yep, absolutely: I'll add tests for those.

pozorvlak:
I added some tests, but as you can see they failed with the error `undefined method `days_until_harvest_min=' for #<Seed:0x00000010bd7508>`. Anyone know what I'm doing wrong?

maco:
Your variable is called days_until_maturity_min not days_until_harvest_min

pozorvlak:
...aaaand this is what I get for coding late at night. *facepalm*

pozorvlak:
I've altered the wording on the "add seeds" form as suggested by @Skud, but I'm not very happy with the look of it:

screenshot from 2015-02-01 17 52 55

Also, I don't think you're meant to [use input-groups that way](http://getbootstrap.com/components/#input-groups).

Skud:
Where's this at?  It looks ok to me except that @pozorvlak thinks he's misusing the input group bootstrap class.

pozorvlak:
> What happens if you just make this a col-md-8 and then put f.number_field, a span for the text, and f.number_field?

screenshot from 2015-02-05 16 03 28

> Failing that, why can't you just put the cols (as below) without grouping them in a .input-group? I would suggest, though, making sure they equal 8 in total (even if you just make the last one wider).

screenshot from 2015-02-05 16 06 12

pozorvlak:
Multiple `form-control`s within a single `input-group` are apparently "not supported": I don't know whether we should worry about that or not. I think the version with the `input-group` is the least ugly of the three screenshots I've posted, though it's far from beautiful.

maco:
Are you possibly looking for the fieldset tag?

pozorvlak:
...maybe? Replacing `.input-group` with `%fieldset`, ie using the code

```haml
  .form-group
    = f.label :days_until_maturity_min, 'Days until maturity:', :class => 'control-label col-md-2'
    %fieldset
      .col-md-2
        = f.number_field :days_until_maturity_min, :class => 'form-control'
      .col-md-1
        = f.label :days_until_maturity_max, 'to', :class => 'control-label'
      .col-md-2
        = f.number_field :days_until_maturity_max, :class => 'form-control'
      .col-md-1
        = f.label :dummy, 'days', :class => 'control-label'
```
results in

screenshot from 2015-02-05 16 36 54

ie, the same as my third screenshot above. Or do you mean something else?

maco:
Well since you're not "suppose" to put multiple form controls in an
input-group, but that's actually what fieldset is *meant* for...I figured
I'd toss it out there.

pozorvlak:
Oh, right. Yeah, it was definitely worth a shot, and I've learned something about HTML today. Does `fieldset` work better in screenreaders?

maco:
No idea

Skud:
Yes, fieldsets are better for a11y, according to http://www.coolfields.co.uk/2011/10/accessible-forms-grouping-elements-together-with-fieldset/ and a bunch of other pages I found.

Also just a general "argh!" to this whole thing.

I say we go with the fieldset solution even if it's a bit clunky-looking. The *real* fix for this, in fact, is the "edit in place" screen, as I'm sure under that regime this would not be one of the key fields we require on the "new seeds" form!

pozorvlak:
Agreed: I'll commit the fieldset version and push it up tomorrow morning.

PR #680, opened 2015-01-27
Garden photos

### LINKED ISSUE #600: Garden photos
You should be able to attach photos to gardens, much as we currently do for plantings and harvests. These photos should show on the garden detail page for now (probably at the top of the page, similar to how photos show on the crop detail page).

maco:
This should get the ability to add photos to gardens. The photos will then display both on the garden detail page and in the garden tabs on the member page.

Fixes #600

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20141119130555) do
+ActiveRecord::Schema.define(version: 20150127043022) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -20,16 +20,16 @@
     t.string   "name",              null: false
     t.boolean  "is_paid"
     t.boolean  "is_permanent_paid"
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.datetime "created_at",        null: false
```
Skud: augh WHAT is this weird schema thing that keeps happening? Obviously something (recent version of rake, maybe?) wants to make these changes. Wish I understood what :(

Skud:
Looking pretty good!  FYI to @maco: when doing a PR for a github issue, if you say "Fixes #600" in your pull request message, it will link the two together, and close the issue when the PR is merged.  I'll add it to your PR now, but you should keep it in mind for next time.

maco:
I couldn't remember if that was supposed to happen this point or in merge
to master

db/schema.rb
pozorvlak: @Skud do you think this should block merging the PR? Otherwise, it looks good to me.

pozorvlak:
@maco Good question! Our pre-GitHub process didn't mark issues as closed until they were deployed on master, but since we moved to GitHub Issues people have been using "Fixed #XXX" on the pull requests into dev. I think the latter is probably more useful from a workflow-tracking point of view, TBH.

db/schema.rb
maco: @skud i'm glad to hear it's a "keeps happening" and not a "i did the migration wrong"

Skud: Yeah. I don't know what causes it but I don't think it'll do any harm. I'm going to just shrug and move with it.

PR #682, opened 2015-01-29
fixing harvest csv so it doesn't error

### LINKED ISSUE #681: Harvest CSV throws an error
I hit this on my local branch while testing to see if I got the harvest csv to include standardized weights, and then I reproduced it on the live site.

NoMethodError at /harvests/owner/test1.csv
undefined method `name' for #<Member:0x007fe8bdfe4838>

```
    respond_to do |format|
      format.html { @harvests = @harvests.paginate(:page => params[:page]) }
      format.json { render json: @harvests }
      format.csv do
        specifics = (@owner ? "#{@owner.name}-" : @crop ? "#{@crop.name}-" : nil)
        @filename = "Growstuff-#{specifics}Harvests-#{Time.zone.now.to_s(:number)}.csv"
        render :csv => @harvests
      end 
    end 
```

It's highlighting the ```specifics = (@owner ? "#{@owner.name}-" : @crop ? "#{@crop.name}-" : nil)``` line.

maco:
owner.name doesn't exist, but owner.login_name does

This is making the harvest CSV error out when you try to see it. I'm not really sure what kind of test I'd add for this though.

Fixes #681

### DIFF
--- a/app/controllers/harvests_controller.rb
+++ b/app/controllers/harvests_controller.rb
@@ -20,7 +20,7 @@ class HarvestsController < ApplicationController
       format.html { @harvests = @harvests.paginate(:page => params[:page]) }
       format.json { render json: @harvests }
       format.csv do
-        specifics = (@owner ? "#{@owner.name}-" : @crop ? "#{@crop.name}-" : nil)
+        specifics = (@owner ? "#{@owner.login_name}-" : @crop ? "#{@crop.name}-" : nil)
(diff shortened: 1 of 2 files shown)


Skud:
To test this: I'd suggest a feature test where you try to fetch the CSV, and then check that you got a 200 status. That would be a minimal regression test for this!

I think there are some similar kinds of tests for RSS in spec/features if you want to copypasta.

maco:
Thanks. I have people over tonight, but I might be able to do that after
they leave.

PR #684, opened 2015-01-31
Standard weights for harvests

### LINKED ISSUE #601: Standardised harvest weights
When a harvest is entered, the weight should be recorded in the user's preferred units but also stored as a conversion to a standardized unit (kg). This will make future statistical analysis of harvests easier.

This will also require a rake task to set the standardized weight of pre-existing harvests.

Task breakdown (from PT):

* add column to track standardised weight
* when a harvest is saved, fill in the standardised weight (probably before_save or something)
* write rake task to fill in standardised weights for older harvests
* include standardised weight in CSV downloads for harvests

maco:
Parts to this feature:
- [x] add column to track standardised weight
- [x] when a harvest is saved, fill in the standardised weight (probably before_save or something)
- [x] test that
- [x] write rake task to fill in standardised weights for older harvests
- [x] include standardised weight in CSV downloads for harvests

This includes tests of converting from oz, lb, and kg to kg. 

Fixes #601


maco:
Given what the tests did after I updated to use Unit(), I might know what's wrong. I think it's trying to interpret our :unit stuff

maco:
oh, I think I know what's wrong

pozorvlak:
@Skud: the base unit of mass in SI is the kilogram - see http://en.wikipedia.org/wiki/SI_base_unit (though perhaps you knew that and were using "base" to mean "unprefixed"?). I'm in favour of using kg as our standard unit, mostly because it's the base SI unit; if we switch to using grams instead then we shouldn't call the field `si_weight`.

Skud:
I did mean unprefixed I guess and also "smallest practical". I was thinking partly about how it's hood practice to store money as cents not dollars to avoid floating point trouble. But tbh you're guaranteed to have floating point trouble here since we allow floating point input anyway.

Don't mind me! Ignore my previous comment, I retract it!

maco:
Ok, ready for final reviews!

app/models/harvest.rb
```diff
@@ -59,6 +59,17 @@ class Harvest < ActiveRecord::Base
...
   after_validation :cleanup_quantities
 
+  before_save :set_si_weight
+
+  # we're storing the harvest weight in kilograms in the db too 
+  # to make data manipulation easier
+  def set_si_weight
+    if self.weight_unit != nil
+      weight_string = "#{self.weight_quantity} #{self.weight_unit}"
+      self.si_weight = Unit(weight_string).convert_to("kg").to_s("%0.2f").delete(" kg").to_f
```
Skud: I think we probably want "%0.3f" which rounds to the nearest gram.

lib/tasks/growstuff.rake
```diff
@@ -314,6 +314,18 @@ namespace :growstuff do
         end
       end
     end
+
+    desc "January 2015: fill in si_weight column"
+    task :populate_si_weight => :environment do
+      Harvest.find_each do |h|
+        if h.weight_unit != nil
+          weight_string = "#{h.weight_quantity} #{h.weight_unit}"
+          h.si_weight = Unit(weight_string).convert_to("kg").to_s("%0.2f").delete(" kg").to_f
```
Skud: This code is repeated here and in the model. Why not just call h.set_si_weight?

Skud:
I left two tiny comments!  Do those, and merge the latest dev, and I reckon this looks good to go.

lib/tasks/growstuff.rake
maco: I feel silly for not realizing I could do that.

maco:
done!

maco:
done

PR #686, opened 2015-02-01
Allow the planting reminder spec to test links in emails properly

yoongkang:
In #420 we found that we weren't able to test links due to the host defaulting to www.example.com. Turns out that since this spec isn't a mailer spec, but a feature spec, we aren't using the default host as configured for ActionMailer in test.rb. 

Here's the bit in the RSpec code that retrieves the default host: 

https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/example/feature_example_group.rb#L18 

Unfortunately, I wasn't able to find a way to configure it in test.rb. Any other option seemed to affect everything else. So, I decided to override the feature spec instead, since it will only affect the spec.

### DIFF
--- a/spec/features/planting_reminder_spec.rb
+++ b/spec/features/planting_reminder_spec.rb
@@ -40,10 +40,8 @@ feature "Planting reminder email", :js => true do
 
     scenario "lists plantings" do
       expect(mail).to have_content "most recent plantings you've told us about"
-      expect(mail).to have_content @p1.to_s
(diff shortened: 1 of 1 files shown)


pozorvlak:
Do we still need the `have_content` expectations now we have `have_link` ones?

Skud:
Agreed with @pozorvlak -- the have_content tests would be better as :text params to the have_link test, eg. have_link :href => foo, :text => bar

PR #687, opened 2015-02-01
Crop request

### LINKED ISSUE #602: Improve "request new crops" process
As an ordinary Growstuff member, I want to be able to request crops by some mechanism other than posting in a forum.

Specifically, I should be able to use a variant of the "new crop" form, which asks me for as much detail as I have, then records that crop as "requested" or something but doesn't make it live until reviewed by a crop wrangler.

When a crop is requested, email should be sent to all crop wranglers letting them know about it.

As a crop wrangler, I should be able to see these pending crops on the crop wrangler page, and approve them (making edits if necessary).  Alternately, I should be able to reject a crop, providing a reason (eg. "not edible").  It would be good to have a list of standard rejection reasons to select from plus "other".

When the crop is approved, an email should be sent to the requester letting them know it's been added and providing links to plant, harvest, or stash seeds.  If rejected, an email should be sent with the reason.

Some notes:

1) We might want to ease up on validation a bit, eg. to allow them to request it even if they don't know the Wikipedia link or whatever

2) Be careful of race conditions if multiple crop wranglers both come to deal with a new crop simultaneously.

3) We'll need some discussion about how to actually track the state of the crop. I'd suggest that we simply have a "status" field which can be "requested", "approved", and "rejected". All current crops should be set to "approved".

4) We'll also need a requested_by field to track who asked for a crop so we can email them.

5) Keep in mind that scientific names will need to be managed in similar ways.

6) We'll want to modify the help text on crops/new to be targetted at the general public vs wranglers. OTOH, crops/edit should be wrangler-specific.

tygriffin:
Fixes #602

**WIP: Needs feedback**

1. Is the language natural enough for regular members on the new crops form page?

2. Should we use Javascript to pop up the reason for rejection only when `approval_status` is "rejected" on the edit crops form?

3. Is the `reason_for_rejection` required if rejected?

4. Should the reasons for rejection be hardcoded in the model or should they be in a database table? (Database seems to make more sense. That way, if a wrangler selects "other" as a reason, wranglers can be presented the option to create a new reason.)

5. How should we avoid race conditions? Should we simply prevent a wrangler from marking a crop rejected if it's already approved and vice versa - with a descriptive error message?

6. Should it be impossible to navigate to rejected crop by going to it's url directly? Should rejected crops be listed anywhere?

### DIFF
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -169,6 +169,6 @@ class CropsController < ApplicationController
   private
 
   def crop_params
-    params.require(:crop).permit(:en_wikipedia_url, :name, :parent_id, :creator_id, :approval_status, :request_notes, :reason_for_rejection, :scientific_names_attributes => [:scientific_name])
+    params.require(:crop).permit(:en_wikipedia_url, :name, :parent_id, :creator_id, :approval_status, :request_notes, :reason_for_rejection, :scientific_names_attributes => [:scientific_name, :_destroy, :id])
   end
 end
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -32,7 +32,8 @@ class CropsController < ApplicationController
 
   # GET /crops/wrangle
   def wrangle
-    @crops = Crop.recent.paginate(:page => params[:page])
+    @recent_crops = Crop.recent.paginate(:page => params[:page])
+    @pending_approval = Crop.pending_approval
     @crop_wranglers = Role.crop_wranglers
     respond_to do |format|
       format.html
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -106,9 +106,11 @@ class CropsController < ApplicationController
 
     if current_member.has_role? :crop_wrangler
       @crop.creator = current_member
+      success_msg = "Crop was successfully created."
     else
       @crop.requester = current_member
       @crop.approval_status = "pending"
+      success_msg = "Crop was successfully requested."
     end
 
     respond_to do |format|
@@ -119,7 +121,7 @@ class CropsController < ApplicationController
           end
         end
 
-        format.html { redirect_to @crop, notice: 'Crop was successfully created.' }
+        format.html { redirect_to @crop, notice: success_msg }
         format.json { render json: @crop, status: :created, location: @crop }
       else
         format.html { render action: "new" }
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -147,6 +147,8 @@ class CropsController < ApplicationController
 
     previous_status = @crop.approval_status
 
+    @crop.creator = current_member if previous_status == "pending"
+
     respond_to do |format|
       if @crop.update(crop_params)
         if previous_status == "pending"
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -101,8 +101,14 @@ class CropsController < ApplicationController
   # POST /crops
   # POST /crops.json
   def create
-    params[:crop][:creator_id] = current_member.id
     @crop = Crop.new(crop_params)
+
+    if current_member.has_role? :crop_wrangler
+      @crop.creator = current_member
+    else
+      @crop.requester = current_member
+      @crop.approved = false
+    end
 
     respond_to do |format|
       if @crop.save
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -133,8 +133,15 @@ class CropsController < ApplicationController
   def update
     @crop = Crop.find(params[:id])
 
+    previous_status = @crop.approval_status
+
     respond_to do |format|
       if @crop.update(crop_params)
+        if previous_status == "pending"
+          requester = @crop.requester
+          new_status = @crop.approval_status
+          Notifier.crop_request_approved(requester, @crop).deliver! if new_status == "approved"
+        end
         format.html { redirect_to @crop, notice: 'Crop was successfully updated.' }
         format.json { head :no_content }
       else
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -113,6 +113,12 @@ class CropsController < ApplicationController
 
     respond_to do |format|
       if @crop.save
+        if current_member.has_role? :crop_wrangler
+          Role.crop_wranglers.each do |w|
+            Notifier.new_crop_request(w, @crop).deliver!
+          end
+        end
+
         format.html { redirect_to @crop, notice: 'Crop was successfully created.' }
         format.json { render json: @crop, status: :created, location: @crop }
       else
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -108,7 +108,7 @@ class CropsController < ApplicationController
       @crop.creator = current_member
     else
       @crop.requester = current_member
-      @crop.approved = false
+      @crop.approval_status = "pending"
     end
 
     respond_to do |format|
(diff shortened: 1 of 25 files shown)


app/controllers/crops_controller.rb
```diff
@@ -32,7 +32,8 @@ def index
 
   # GET /crops/wrangle
   def wrangle
-    @crops = Crop.recent.paginate(:page => params[:page])
+    @recent_crops = Crop.recent.paginate(:page => params[:page])
+    @pending_approval = Crop.pending_approval
```
tygriffin: Note to self: This should be paginated as well.

app/controllers/crops_controller.rb
```diff
@@ -120,8 +135,16 @@ def create
...
 
+    previous_status = @crop.approval_status
+
     respond_to do |format|
       if @crop.update(crop_params)
+        if previous_status == "pending"
+          requester = @crop.requester
+          new_status = @crop.approval_status
+          Notifier.crop_request_approved(requester, @crop).deliver! if new_status == "approved"
+          Notifier.crop_request_rejected(requester, @crop).deliver! if new_status == "rejected"
```
tygriffin: Note to self: If I refactor this using `send` to get rid duplication will that be annoying and less explicit?

spec/mailers/notifier_spec.rb
```diff
@@ -44,4 +44,84 @@
...
+    end
+
+    it 'includes the approved crop URL' do
+      expect(mail.body.encoded).to match crop_url(crop)
+    end
+
+    it 'includes links to plant, harvest and stash seeds for the new crop' do
+      expect(mail.body.encoded).to match new_planting_url
+      expect(mail.body.encoded).to match new_harvest_url
+      expect(mail.body.encoded).to match new_seed_url
```
tygriffin: How can I check that the query parameter is included as well? The following wasn't working for some reason:

`expect(mail.body.encoded).to match new_planting_url(crop_id: crop.id)`

Can we just use `have_link` in this test?

app/controllers/crops_controller.rb
tygriffin: Note to self: If approved set the currently logged in wrangler as the creator.

app/models/crop.rb
```diff
@@ -21,16 +22,18 @@ class Crop < ActiveRecord::Base
   before_destroy {|crop| crop.posts.clear}
 
   default_scope { order("lower(name) asc") }
```
Skud: Can we put the where(:approval_status => "approved") into the default scope and save ourselves having to repeat it many times below?  Then when we want the pending approval ones, we just ask for those.

app/models/crop.rb
```diff
@@ -21,16 +22,18 @@ class Crop < ActiveRecord::Base
...
+  scope :randomized, -> { where(:approval_status => "approved").reorder('random()') } # ok on sqlite and psql, but not on mysql
+  scope :pending_approval, -> { where(:approval_status => "pending") }
 
   validates :en_wikipedia_url,
     :format => {
       :with => /\Ahttps?:\/\/en\.wikipedia\.org\/wiki/,
       :message => 'is not a valid English Wikipedia URL'
-    }
+    },
+    :if => :approved?
```
Skud: maybe leave a comment here explaining the :if => approved thing

Skud:
Feedback:

1. Is the language natural enough for regular members on the new crops form page?

Can we see a screenshot?  I think it's probably ok though.

2. Should we use Javascript to pop up the reason for rejection only when approval_status is "rejected" on the edit crops form?

If you feel like it. I wouldn't say it's required though. We can expect crop wranglers to handle it okay, I think.

3. Is the reason_for_rejection required if rejected?

Yes.

4. Should the reasons for rejection be hardcoded in the model or should they be in a database table? (Database seems to make more sense. That way, if a wrangler selects "other" as a reason, wranglers can be presented the option to create a new reason.)

Would we ever want to look up all the crops that were denied for X reason? If so, database table. I can't imagine wanting it though.

Here are some suggestions for reasons, to start with:

* already in database
* not edible
* not enough information
* other

5.  How should we avoid race conditions? Should we simply prevent a wrangler from marking a crop rejected if it's already approved and vice versa - with a descriptive error message?

Yes, I think so. Also if they visit the edit page and it's already been approved they should clearly see that that has happened.  I think this implies we might want to include some metadata for wranglers on the edit page, eg. "requested by X on date. approved/rejected by Y on date."

6. Should it be impossible to navigate to rejected crop by going to it's url directly? Should rejected crops be listed anywhere?

Yes, and yes -- they should be listed on the crop wrangler page somewhere (or on a subpage). Nowhere else for now, though, I think.

app/models/crop.rb
tygriffin: I wasn't sure exactly how to do this. If I do this 
```
default_scope { where(:approval_status => "approved").reorder("created_at desc")  }
scope :pending_approval, -> { where(:approval_status => "pending") }
```

I get queries like this:

`SELECT "crops".* FROM "crops"  WHERE "crops"."approval_status" = 'approved' AND "crops"."approval_status" = 'pending'  ORDER BY lower(name) asc`

tygriffin:
**Ready for review**

A couple things may need some additional discussion:

1. Is this language good for non-wrangler members?
screen shot 2015-02-04 at 2 44 48 pm

2. Do we want to include any more standard reasons for rejection other than "already in database", "not edible", "not enough information", and "other"?

Skud:
1. Language to use: I would suggest showing different language to crop wranglers vs non-wranglers.  The wranglers can see what we already have.  For non-wranglers, I'd put something like 

* Please provide the common name for the crop, in English, if you know it.
* Please provide a link to the crop's page on the English language Wikipedia. Crops without Wikipedia pages cannot be added to our database at this time.

2. I think that'll do for now and we can see what people put in the comments for "other" if there are a lot of them.  If you go through the discussion thread at http://growstuff.org/posts/skud-20130319-requests-for-new-crops you can see the reasons given in the past. The ones I see there are:

* silverbeet - already in db
* kumara - already in db
* rhubarb - already in db
* spider plant - not edible
* rabbit foot fern - not edible
* madder - not edible
* longish list of cover crops etc - not edible
* daisy - other (insufficiently specific; some are edible and some aren't)
* rocket - already in db
* cerejeira - already in db
* peace lily - not edible

And lots and lots that were "there's no wikipedia page for this so we can't add it" which is basically "not enough information" I guess?

Skud:
Just thinking about this a bit more and I think the page title should vary too:

* Non-wrangler: "Suggest a crop"
* Wrangler, reviewing a suggested crop: "Review crop: $cropname"
* Wrangler, adding a brand new crop: "New crop"

Also there should be some information for non-wranglers, to help them understand the process.  At the top of the page:

> Thanks for taking the time to suggest a crop! Our crop database is managed by volunteers, and we appreciate your help.  Here are some things to consider when suggesting a new crop:
> 
> * First, you might want to [search our crops](http://growstuff.org/crops/search) to make sure we don't have it already, perhaps under an alternate name.
> * The Growstuff database only contains edible crops.  In future we hope to support other crops, but for now, if your suggestion is not edible we won't be able to add it.
> * At this time, we are only adding crops which have a Wikipedia page. If you want to add a specific variety of crop that doesn't have its own Wikipedia entry, please use the more general form of the crop instead and put the name of your variety in the notes/description.

(Note: that crop search link doesn't actually include a search form, except in the navbar. We might want to include one there just so we can point people at it.)

After the form but before the submit button:

> When you submit this form, your suggestion will be sent to our team of [volunteer crop wranglers](http://talk.growstuff.org/c/crop-wrangling) for review.  We'll let you know the outcome as soon as we can.

tygriffin:
Good ideas! 

I didn't add the second crop search field to `/crops/search` because I thought the one in the nav bar was pretty obvious. What do you think though? Am I wrong on that one?

tygriffin:
Also, we're starting to accumulate a lot of logic in the views. What do you think about making a new story to add decorators to Growstuff a la [Draper](https://github.com/drapergem/draper)

(PS: That gem is looking for a new maintainer so maybe not so great, but we could make our own light-weight decorator class.)

Skud:
Draper looks awesome but let's hold off on that for now.

Skud:
@tygriffin re: search, the problem is that http://staging.growstuff.org/crops/search (linking to staging since that has elasticsearch) is a really unfriendly, weird page to link someone to.  The heading is weird and ungrammatical (I mentioned this in the "please test" thread for elasticsearch) and then it says "no results found, please do this other thing".  I am not thrilled about linking to that page.  It's not the end of the world but I would like to fix it up.  I guess we could fork a separate story.

PR #691, opened 2015-02-02
Elasticsearch tests

Skud:
If you don't have elasticsearch installed, tests fail whenever you add a crop/sciname/altname, as it tries to index them.  This fixes that problem.


Skud:
Yeesh. I have something wrong with my dev branch, I think. I seem to have pulled @tygriffin's stuff into there by accident. I'm going to close this and then try again.

PR #692, opened 2015-02-02
Fixed test failure if elasticsearch isn't present in test env

Skud:
Fix for tests that were failing if you didn't have elasticsearch running on your dev machine.


pozorvlak:
Works for me - tests fail without this PR, and pass with it. Merging.

PR #693, opened 2015-02-03
make the one-off save the si weight

maco:
fix the deploy task for standard weights feature


Skud:
Short and sweet! Should be auto-deployed within 10 mins.

PR #695, opened 2015-02-03
Functionality to sort members by recently joined : Issue #550

### LINKED ISSUE #550: Sort members by recently joined
It would be nice if we could easily see the newest/most recently joined members on the site.  I'm imagining a dropdown for sort order, or something like that, on the members/index page.  We could copy the implementation from the "Browse crops" sort options.

sksavant:
I've mostly imitated the crops sort thing.
Not quite sure how to write automated tests for this, if any. I'd appreciate help regarding that.

Fixes #550

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -32,6 +32,7 @@ class Member < ActiveRecord::Base
   scope :confirmed, -> { where('confirmed_at IS NOT NULL') }
   scope :located, -> { where("location <> '' and latitude IS NOT NULL and longitude IS NOT NULL") }
   scope :recently_signed_in, -> { reorder('updated_at DESC') }
+  scope :recently_joined, -> { reorder("confirmed_at desc") }
   scope :wants_newsletter, -> { where(:newsletter => true) }
 
   has_many :follows, :class_name => "Follow", :foreign_key => "follower_id"
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -53,3 +53,5 @@ submit the change with your pull request.
 - Justin Hamman / [juzham](https://github.com/juzham)
 - Rocky Jaiswal / [rocky-jaiswal](https://github.com/rocky-jaiswal)
 - Robert Landreaux / [robertlandreaux](https://github.com/robertlandreaux)
+- Savant Krishna / [sksavant](https://github.com/sksavant)
+
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -6,7 +6,12 @@ class MembersController < ApplicationController
   after_action :expire_cache_fragments, :only => :create
 
   def index
-    @members = Member.confirmed.paginate(:page => params[:page])
+    @sort = params[:sort]
+    if @sort == 'recently_joined'
+      @members = Member.confirmed.recently_joined.paginate(:page => params[:page])
+    else
+      @members = Member.confirmed.paginate(:page => params[:page])
+    end
 
     respond_to do |format|
       format.html # index.html.haml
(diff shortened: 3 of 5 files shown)


Skud:
Fantastic! Thanks so much for this and welcome to Growstuff :)

I'd recommend creating tests in spec/features/crops that goes through the process of visiting the members page, clicking on the form, submitting, looking at the resulting sort order, etc.

There's some equivalent tests for the crop sort in spec/views/crops/index.html.haml_spec.rb but we are moving away from view tests to feature tests, so please don't write anything in spec/views but do it in spec/features instead.  (For bonus points, move the crops stuff over too!)

Btw I see you're in Melbourne -- do you know about the [Open Knowledge workshop](http://www.meetup.com/Open-Knowledge-Melbourne/events/219627090/) tomorrow night? I'll be giving a talk about Growstuff and a few other Growstuff folks will be there too.  We also meet a couple of times a month to work on Growstuff together -- see [here](http://talk.growstuff.org/c/events) for details.

app/models/member.rb
```diff
@@ -32,6 +32,7 @@ class Member < ActiveRecord::Base
   scope :confirmed, -> { where('confirmed_at IS NOT NULL') }
   scope :located, -> { where("location <> '' and latitude IS NOT NULL and longitude IS NOT NULL") }
   scope :recently_signed_in, -> { reorder('updated_at DESC') }
+  scope :recently_joined, -> { reorder("created_at desc") }
```
Skud: I think this should probably be confirmed_at so we get the time when they actually click the confirmation link rather than when they first sign up. I'm open to argument on this though!

sksavant:
Thanks @Skud , I'll have a look.

Yes, I've got on to the website from the workshop page although @oshiho3 mentioned about this project some time ago. Looks like an interesting project :)

I'll see you tomorrow, most probably.

app/models/member.rb
sksavant: You may be right but technically a user could be considered as joined as soon as they registered.

Skud: My concern is this.  Say that you have two members, Alice and Bob.

* Alice signs up.
* Bob signs up.
* Some time passes.
* Bob confirms.
* Now someone looks at the member page (or RSS feed, which should come next) and sees Bob's the most recently joined.
* More time passes
* Alice confirms.
* Now you look at the most recently signed up page/RSS feed and Bob is still the "most recent" and you might not notice Alice.  

Although Alice *signed up* first, she's not visible on the "most recent* page until she's confirmed, and she might not get noticed if she confirms out of order.

FWIW, I'm intending to subscribe to the RSS feed once we get it, so this is pretty relevant to me :)

sksavant: Sure. Which one do you think should be the default view? Alphabetical sort or Recently joined?

Skud: Hmmm. Unlike crops, I don't think there's much benefit to putting 
recently joined first.  I'd say alpha sort as default, for now.  In 
future we might want to have "most active" or something as the default, 
which would give you a more interesting first page.


-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

tygriffin:
Hi @sksavant! I think we sat next to each other at a recent Open Knowledge Workshop. Thanks for this PR and welcome!

yoongkang:
Would it be worth putting in a 'Most followed' filter as well?

Skud:
@yoongkang maybe? I think it would be better to encourage people to follow people *near* them, or who have similar growing habits, rather than just the most popular.  But maybe worth bringing up under "Ideas" on Growstuff Talk.

Skud:
Hi @sksavant, just wanted to check on the status of this PR?  Do you think you'll be able to write the tests discussed above?

sksavant:
Hey @Skud, I've just pushed the tests. Let me know if it's fine.

Skud:
Looks great, thanks!

Skud:
Wait, one more thing! Please add yourself to CONTRIBUTORS.md.

sksavant:
Done and rebased on latest dev.

sksavant:
Changed to use let instead of begin.

Skud:
Thanks! FYI we don't usually rebase as we like to see people's process, so no need to do that in future.

Issue #541, opened 2014-12-12
"like" posts


Skud:
I should be able to "like" a post to indicate that I think it's helpful/interesting/whatever.

When I click "like" there should be a secondary thing to click that lets me say why I liked it:
- helpful
- interesting
- beautiful
- funny
- sympathy
- ???

(We probably want to discuss this list before implementing it. We might also want to let people choose multiples?  See Ravelry and Slashdot for design inspiration.)

We can then use this info to show eg. the most helpful posts related to a given crop.

When member A "likes" member B's post, member B should get a notification.

Note: we probably also want to apply this "why I liked it" thing to #540 (photo likes).

maco:
I'm a fan of how Ravelry does it, which doesn't include a notification. Those might get to be excessive as the site grows. However if the helpful/interesting/etc-counts show up on the posts when viewing "My Posts" that'd be nice. You have to click through to the thread to see that info on Ravelry.

Skud:
Agreed that it could be spammy. On the other hand we are having a 
problem with getting people with accounts to visit GS regularly, and 
occasional email reminding them of things going on on the site would 
probably help!

On 30/01/2015 2:50 pm, Mackenzie wrote:
>
> I'm a fan of how Ravelry does it, which doesn't include a 
> notification. Those might get to be excessive as the site grows. 
> However if the helpful/interesting/etc-counts show up on the posts 
> when viewing "My Posts" that'd be nice. You have to click through to 
> the thread to see that info on Ravelry.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/Growstuff/growstuff/issues/541#issuecomment-72150174>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

cesy:
Making the list configurable by an admin and allowing you to choose multiple items sounds best. And notifications are a good optional extra that could be added once this is working, or included in the existing activity emails.

Br3nda:
> I should be able to "like" a post to indicate that I think it's helpful/interesting/whatever.

This portion is done

CloCkWeRX:
#2079

CloCkWeRX:
Likes for posts exists.

PR #696, opened 2015-02-04
WIP like feature for posts

yoongkang:
Hi, I've been looking at #541 and started some preliminary hacking. Before I go much further I'd like to open a discussion with regards to the specifics of the implementations (design, categories of 'likes', etc.), and also do a code review.

I have a module called "Likeable" which you can now just include in any model and it will become a likeable model. Should be extensible to photos, crops, etc.

At the moment I've only added it to posts, and it's still ugly as. No AJAX yet either. 

Hopefully I'll get to drop by one of the hack nights so we can iron this out.


app/models/likeable.rb
```diff
@@ -0,0 +1,9 @@
+module Likeable
+  extend ActiveSupport::Concern
```
Skud: Love that we're doing this with concerns. I'd love to make our photos code be a concern too, in future.

Skud:
This looks fantastic. I particularly love the use of concerns (which I think we should extend to photos, too, in future).

Agreed that we should talk about design (in the visual sense) and categories for likes.  I think I should raise the latter, at least, on Growstuff Talk.  I'll start a thread there.

tygriffin:
Like!

cesy:
I'd love to get this moving again. @yoongkang, are you still keen for hacking on it? We've got a great start here. The other thing mentioned in the issue is having reasons for liking a post, and for those reasons to be configurable by an admin, I think, but that's not necessarily required for minimum viable product and can be added in a future release. Similarly, we can spin off a separate issue for AJAX-ifying it.

So what else is needed to make this usable? Next steps are to fix the merge conflicts and share a couple of screenshots for input on the design, I think, to see how much CSS change is needed before we go live with it.

yoongkang:
@cesy Yep, I'm happy to continue with this. Hopefully I'll have a bit of time this weekend to hack on it. 

If you're happy right now to do a page reload every time a 'Like' button is clicked, I'll just have to make it not ugly and it should be good to go for another review.

AJAX will require a unique id for every element containing the Like link for lists of likeable items. I can't remember what stopped me from implementing that but I'll take a look again.

cesy:
Have you had a chance to look at this, or are you interested in picking it up again? We've just started deploying more often again.

app/models/likeable.rb
maco: I think Rails convention is to put model concerns in app/models/concerns/

maco:
A "type" field on the Like would be a good thing to pop in. The full feature includes having many types of reactions (hmm Facebook is calling them "reactions" now that they added things other than just "like" and that might be a better class name), and if we create `ReactionType` or `LikeType` now we can avoid having to have a migration later that shifts all uncategorized Likes into a category.

yoongkang:
Sorry, I've been way too busy. Happy to free it up for anyone else who's interested.

PR #697, opened 2015-02-06
Needs Feedback: Homepage redesign redux

### LINKED ISSUE #628: Homepage redesign, 2015

tygriffin:
Fixes #628

Replaces #674

~~@Skud :~~
~~> I think after you select something from that search UI it should take you direct to the crop page in question (eg. /crops/eggplant) not to search results (unless there's not an exact match).~~

~~This should work in production where we're using Elasticsearch but it doesn't seem to work in development. See inline note.~~

@Skud :

> Why is the crop search not an autosuggest? If I type "pea" or "berry" for instance I don't see the list.

Good call. I need to refactor autosuggest a bit. Should I do that now or break into a new story?

**Question**: Should the [signed in user homepage redesign](http://talk.growstuff.org/t/major-homepage-redesign/168) be part of this PR?

**Screenshot** (updated March 1, 2015)
screen shot 2015-03-01 at 5 07 09 pm


app/controllers/crops_controller.rb
```diff
@@ -61,14 +61,20 @@ def hierarchy
   # GET /crops/search
   def search
     @all_matches = Crop.search(params[:search])
-    exact_match = Crop.find_by_name(params[:search])
+    exact_match = Crop.find_by(name: params[:search])
+    exact_search = params[:exact_search] || false
+
     if exact_match
       @all_matches.delete(exact_match)
```
tygriffin: This seems to delete from the actual database in development. I think this is because we're not using Elasticsearch in the dev env and so `search` is returning an `ActiveRecord::Relation`, not a regular array. It looks like calling search results in queries like this for example:

`DELETE FROM "crops" WHERE (name ILIKE '%corn%') AND "crops"."id" = 61`

Not sure how to debug this.

cesy:
Does this still need a review?

tygriffin:
@cesy : I believe it does.

app/assets/stylesheets/overrides.css.less
```diff
@@ -214,6 +206,77 @@ li.crop-hierarchy {
...
+  min-height: 280px;
+  padding: 30px;
+  margin: 0 -15px 0 -15px;
+
+  @media (max-width: @screen-md-min) {
+    background-size: @screen-md-min;
+  }
+
+  h1 {
+    font-size: 60px;
```
cesy: Can we make this relative in ems instead of absolute, for accessibility?

app/assets/stylesheets/overrides.css.less
```diff
@@ -214,6 +206,77 @@ li.crop-hierarchy {
...
+    margin: 60px;
+  }
+}
+
+.image-attribution {
+  position: absolute;
+  bottom: 0; right: 5px;
+  font-size: 12px;
```
cesy: Again, shouldn't be absolute.

cesy:
Is it easy to split out the search changes into a separate PR?

cesy:
Can we also add a feature test for the new homepage, to check some of the values and features are coming out correctly?

I've had a first review of this and most of it looks good. It will need someone with more Javascript expertise to review those parts.

app/assets/stylesheets/overrides.css.less
tygriffin: Totally. Shouldn't be a problem.

Skud: Actually, can we use Bootstrap less variables please!

Bootstrap itself is based on a base pixel size, and then has things like 
@headingblah (I forget the exact names) to set the sizes of everything else.

On 27/03/2015 12:09 am, Taylor Griffin wrote:
>
> In app/assets/stylesheets/overrides.css.less 
> <https://github.com/Growstuff/growstuff/pull/697#discussion_r27210396>:
>
> > +
> > +.title-banner {
> > +  position: relative;
> > +  background-image: image-url('veggies.jpg');
> > +  background-repeat: no-repeat;
> > +  background-size: 100%;
> > +  min-height: 280px;
> > +  padding: 30px;
> > +  margin: 0 -15px 0 -15px;
> > +
> > +  @media (max-width: @screen-md-min) {
> > +    background-size: @screen-md-min;
> > +  }
> > +
> > +  h1 {
> > +    font-size: 60px;
>
> Totally. Shouldn't be a problem.
>
> —
> <https://github.com/Growstuff/growstuff/pull/697/files#r27210396>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

tygriffin: So glad you made this suggestion. I've been using variables for all kinds of padding and margins at work, and it's been a real game changer. I'll get caught up on this PR this weekend, btw.

cesy:
Hi @tygriffin, we've now got Release 8 out the way and are keen to get this moving again if you're around.

tygriffin:
Hi @cesy! Great to hear about Release 8. Unfortunately, I don't think I'll have time to do much development on Growstuff in the foreseeable future. Do you think someone might be able to take over this story?

CloCkWeRX:
I'd be happy to take this one over and redo against current dev

cesy:
Thank you, that would be great!

CloCkWeRX:
Shifted to #1113

PR #698, opened 2015-02-07
Added Helper for See Who's Planted Link Text

### LINKED ISSUE #633: Problems with crop page sidebar display
The display of plantings and harvests in the crop detail page sidebar is a bit off.  If some details aren't filled in, it comes out as "Skud planted ." or similar.

screenshot 2015-01-15 10 34 50

This seems to be a result of [this work](https://github.com/Growstuff/growstuff/commit/14d039ee018b4d877df10c17285d7a0f471fecd4) which means the problem's been there for a few months.  However I think it's less likely to show up on production (the screenshot is from staging) because people tend to fill in more details there.  Here's a screenshot from production:

screenshot 2015-01-15 10 44 59

I've marked it as "severity: major" because although happens less on production, the crop detail page is our showcase, and this is definitely higher priority than most of the "severity: minor" bugs.

I would suggest looking at Harvest.to_s and seeing whether we can use that for harvests at least, and whether we can do something similar for plantings, to standardise this kind of display.

gabrielle27:
Quantity and planted_from are not required
to add a planting. This change handles
the cases where one or both attributes are
missing.

Fixes #633

### DIFF
--- a/app/helpers/plantings_helper.rb
+++ b/app/helpers/plantings_helper.rb
@@ -1,0 +1,15 @@
+module PlantingsHelper
+
+  def display_planting(planting)
+    if planting.quantity.to_i > 0 && planting.planted_from.present?
+      return "#{planting.owner} planted #{pluralize(planting.quantity, planting.planted_from)}."
+    elsif planting.quantity.to_i > 0
+      return "#{planting.owner} planted #{pluralize(planting.quantity, 'unit')}."
+    elsif planting.planted_from.present?
+      return "#{planting.owner} planted #{planting.planted_from.pluralize}."
+    else
+      return "#{planting.owner}."
+    end
+  end
+
+end
--- a/app/views/crops/_plantings.html.haml
+++ b/app/views/crops/_plantings.html.haml
@@ -6,7 +6,7 @@
   %ul
     - crop.plantings.take(3).each do |planting|
       %li
-        = link_to "#{planting.owner} planted #{planting.quantity} #{planting.planted_from}.", planting_path(planting)
+        = link_to display_planting(planting), planting_path(planting)
         = render :partial => 'members/location', :locals => { :member => planting.owner }
         %small
           = distance_of_time_in_words(planting.created_at, Time.zone.now)
--- a/spec/helpers/plantings_helper_spec.rb
+++ b/spec/helpers/plantings_helper_spec.rb
@@ -1,0 +1,78 @@
+require 'rails_helper'
+
+describe PlantingsHelper do
+  describe "display_planting" do
+
+    before(:each) do
+      @member = FactoryGirl.build(:member,
+        :login_name => 'crop_lady'
+      )
+    end
+
+    it "neither quantity nor planted from provided" do
+      planting = FactoryGirl.build(:planting,
+        :quantity => nil,
+        :planted_from => '',
+        :owner => @member
+      )
+      result = helper.display_planting(planting)
+      result.should eq "crop_lady."
+    end
+
+    it "quantity not provided" do
+      planting = FactoryGirl.build(:planting,
+        :quantity => nil,
+        :planted_from => 'seed',
+        :owner => @member
+      )
+      result = helper.display_planting(planting)
+      result.should eq "crop_lady planted seeds."
+    end
+
+    context "multiple quantity" do
+      it "planted_from not provided" do
+        planting = FactoryGirl.build(:planting,
+          :quantity => 10,
+          :planted_from => '',
+          :owner => @member
+        )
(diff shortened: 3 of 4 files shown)


spec/helpers/plantings_helper_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
+
+describe PlantingsHelper do
+  describe "display_planting" do
+
+    before(:each) do
+      @member = FactoryGirl.build(:member,
```
Skud: Can you please use "let" syntax here?

spec/helpers/plantings_helper_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    it "quantity not provided" do
+      planting = FactoryGirl.build(:planting,
+        :quantity => nil,
+        :planted_from => 'seed',
+        :owner => @member
+      )
+      result = helper.display_planting(planting)
+      result.should eq "crop_lady planted seeds."
```
Skud: Please use expect(foo).to syntax for these tests!

Skud:
This looks good except that we're trying to modernise our rspec syntax to use "let" and "expect" rather than "before" and "should", wherever possible.  If you could tweak those that would be great!

Also you might be interested in the discussion about decorators in #687 -- something we're looking at doing in future that would have been relevant here!  No action required at this time, however.

Skud:
Oh, one more thing! Please add yourself to CONTRIBUTORS.md

app/helpers/plantings_helper.rb
```diff
@@ -0,0 +1,15 @@
+module PlantingsHelper
+
+  def display_planting(planting)
+    if planting.quantity.to_i > 0 && planting.planted_from.present?
```
tygriffin: I think `quantity` is already an integer so we shouldn't need `to_i` here or below.

app/views/crops/_plantings.html.haml
```diff
@@ -6,7 +6,7 @@
   %ul
     - crop.plantings.take(3).each do |planting|
       %li
-        = link_to "#{planting.owner} planted #{planting.quantity} #{planting.planted_from}.", planting_path(planting)
+        = link_to display_planting(planting), planting_path(planting)
```
tygriffin: This may be my personal preference, but I would probably refactor this to look like:
```
= link_to planting.to_sentence, planting_path(planting)
```

spec/helpers/plantings_helper_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+describe PlantingsHelper do
+  describe "display_planting" do
+
+    before(:each) do
+      @member = FactoryGirl.build(:member,
+        :login_name => 'crop_lady'
+      )
+    end
+
+    it "neither quantity nor planted from provided" do
```
tygriffin: I don't think we've ever talked about this before, but in situations where the "it" syntax doesn't make grammatical sense, should be using the alternative `specify` syntax?

tygriffin:
Thanks for the PR and welcome! :)

app/helpers/plantings_helper.rb
gabrielle27: quantity can be nil.  I was using to_i here to force the nil to 0.

gabrielle27:
Thanks @Skud and @tygriffin for the feedback.  For my first go-round I was trying to maintain consistency, but you've suggested better practices.  I'm going to close this out and submit a new pull with the revisions.

app/helpers/plantings_helper.rb
tygriffin: Oh, good call!

gabrielle27:
I made the changes previously discussed in this pr.  fwiw I like the idea of moving to a decorator pattern.  I'm a fan of the draper gem.  I'm not sure what the protocol is for merge conflict commits like I ran into with the contributors file.  Let me know if I should cut a new branch.

cesy:
Can you sort out the merge conflict or rebase off a newer version of the dev branch so that we have something clean we can merge, showing just your changes and not other pull requests as well, please?

gabrielle27:
I apologize for taking so long to revisit this. The merge conflicts are resolved now.

cesy:
@gabrielle27 sorry for the delay, it's come up with more merge conflicts now - can you have another look, please? We've now got release 8 live so can merge this promptly.

pozorvlak:
The conflicts were pretty trivial, so I fixed them myself and checked that the tests passed locally. @gabrielle27, thanks for your contribution!

gabrielle27:
@pozorvlak++ thanks!

PR #702, opened 2015-02-10
Request crop links

tygriffin:
Remove links to previous new crop request implementation (sending a post) and replaces it with new crop request form.


spec/views/plantings/edit.html.haml_spec.rb
```diff
@@ -40,7 +40,6 @@
 
     it 'includes helpful links for crops and gardens' do
```
maco: if its only including a helpful link for garden now, maybe this name should be changed. or the assert_select should find the new request link.

tygriffin: You're probably right about both things, but because view tests are deprecated I generally only change them just enough to prevent failing.

tygriffin:
Self merging so we can get on with testing.

PR #706, opened 2015-02-12
add deprecation notice to all controller and view specs

### LINKED ISSUE #622: Add deprecation notes to top of all controller and view tests
It would be great if someone could go through all the controller and view tests and leave a block comment at the top, where it can't be missed, saying that they're deprecated and that people should be writing feature tests instead.

tygriffin:
Fixes #622

Additionally, we can put all controller and view specs under a "deprecated" directory, so it's extra obvious. The paths would look like this: `spec/views/deprecated/`. Probably not necessary though.

### DIFF
--- a/spec/controllers/account_types_controller_spec.rb
+++ b/spec/controllers/account_types_controller_spec.rb
@@ -1,3 +1,19 @@
+## DEPRECATION NOTICE: Do not add new tests to this file!
+##
+## View and controller tests are deprecated in the Growstuff project. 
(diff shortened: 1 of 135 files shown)


pozorvlak:
Or `spec/deprecated/views`?

tygriffin:
@pozorvlak : Yeah, that would make the most sense. However, RSpec won't recognise those specs as view specs (therefore, you wouldn't have access to controller stubs, etc) and they'll all start failing.

PR #710, opened 2015-02-16
Switching to new secrets format

cesy:
I found that rake was failing in my dev environment. The cause seems to be http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#config-secrets-yml

Is this an issue for anyone else?


config/secrets.yml
```diff
@@ -0,0 +1,17 @@
...
+development:
+  secret_key_base: 'b1b67abb399261478f4721e704eb3851466daf60d9cd2b53a1839b056d641c4c1c2a476bcaf7addc6d6548926cfd32fa5a00a8de258880257ebb5a6fd86cb08f'
+  # run 'rake secret' to generate your own
+ 
+test:
+  secret_key_base: 'be557aa019b181f201c9906663dbf8f22efb1b70b11f78035bfeda86aa7dcfd1efb184e2ee894a0ae0dc37fe67d311f38e7731fa16d8d595f2e1ef5447bae020'
+  # run 'rake secret' to generate your own
+ 
+production:
+  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```
cesy: I'll fix the missing line break shortly.

Skud: Where is this set? How is it pushed to Heroku? Please check application.yml (or the example one that's in github, anyway) and modify if necessary.

cesy: Ah, I meant to have it with the same name as the old one, so it would be no change in production/Heroku. Will fix.

PR #715, opened 2015-02-17
Improvements to crop requests

Skud:
As per the list of outstanding problems at http://talk.growstuff.org/t/please-test-requesting-new-crops-new-seed-attributes/223/17

I'm working through them and pushing them to this branch as I do so.

### DIFF
--- a/app/views/crops/_find_seeds.html.haml
+++ b/app/views/crops/_find_seeds.html.haml
@@ -10,7 +10,8 @@
         = render :partial => 'members/location', :locals => { :member => seed.owner }
   %p
     = link_to "View all #{crop.name} seeds", seeds_by_crop_path(crop)
-- if current_member
-  = link_to "List achiote seeds to trade", new_seed_path(:crop_id => crop.id)
-- else
-  = render :partial => 'shared/signin_signup', :locals => { :to => 'list your seeds to trade' }
+- if crop.approved?
+  - if current_member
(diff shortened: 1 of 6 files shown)


Skud:
OK, that's me done for tonight.

There are some serious complexities with removing pending/rejected crops from search results, and I don't know how to address that at all.  Maybe @oshiho3 knows?  I think we need to exclude them from the ElasticSearch index.

Meanwhile, there are several problems related to pending/rejected crops showing up in places they shouldn't, like the crop browse page.  I really want to put:

    default_scope { where(:approval_status => "approved") }

However if we do this, lots of instances of Crop.find (eg. in crop#show, crop#edit, etc) fail. It would be nice to be able to explicitly say "Use this scope that includes all approval statuses" but the various things I tried for this (unscope, rewhere, etc) didn't work.  I have found some useful-looking SO threads but haven't tried them yet:

http://stackoverflow.com/questions/1834159/overriding-a-rails-default-scope
http://stackoverflow.com/questions/25087336/why-is-using-the-rails-default-scope-often-recommend-against

(It looks like we might want to just explicitly call Crop.approved.whatever all over the place instead. Sigh.)

maco:
Shiho's PR looked ready to go, so I merged it. And this one looks good too, what with Shiho's changes being in, soooo merging this too and we can test it out.

PR #716, opened 2015-02-17
exclude unapproved crops from search

oshiho3:
This change will remove unapproved crops from elasticsearch result as mentioned in https://github.com/Growstuff/growstuff/pull/715#issuecomment-74646241.


Skud:
Thank you, this looks great :) will merge in the morning unless someone else gets to it first.

oshiho3:
Hi Alex,

That's fine but wait, it's failing Travis CI. I'm trying to fix it. I may
retreat to bed for today.

maco:
* Starting Elasticsearch Server       it: error setting limit (Operation not permitted)

^ possibly means the server's not running so falling back to sql search which lacks the filter?

Skud:
Oops I didn't notice Travis because I was on my phone!  Yeah better get that fixed.

And as maco notes, we should make sure that this also excludes unapproved crops when ES isn't being used.

oshiho3:
Hi Maco!
Thanks for looking at it. I don't remember setting fallback when elasticsearch is unavailable. I meant to purely control it by env variable (It may be a good idea to implement fallback, but the drawback for doing it is that even when elasticsearch is down it won't get noticed. What do you think @Skud?).
Maybe I'm not configuring the env variable right. I'm going to run a few tests to figure out what's causing this.

I'm going to update the sql search to check the approval_status as well. I was too sleepy last night to finish that bit :-( I'll put WIP in the PR title while I'm fixing travis issue and updating sql search.

Skud:
I don't think we need fallback when ES is down, in production.

I do think that "don't show unapproved crops" needs to work in dev/test 
(i.e. without ES) as well as on staging/production.

A.

On 18/02/2015 10:17 am, oshiho3 wrote:
>
> Hi Maco!
> Thanks for looking at it. I don't remember setting fallback when 
> elasticsearch is unavailable. I meant to purely control it by env 
> variable (It may be a good idea to implement fallback, but the 
> drawback for doing it is that even when elasticsearch is down it won't 
> get noticed. What do you think @Skud <https://github.com/Skud>?).
> Maybe I'm not configuring the env variable right. I'm going to run a 
> few tests to figure out what's causing this.
>
> I'm going to update the sql search to check the approval_status as 
> well. I was too sleepy last night to finish that bit :-( I'll put WIP 
> in the PR title while I'm fixing travis issue and updating sql search.
>
> —
> <https://github.com/Growstuff/growstuff/pull/716#issuecomment-74776647>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

oshiho3:
Got it, Alex. Thanks for the quick response.

maco:
O_o Coveralls seems to be confused. How could adding 3 lines decrease coverage by 50%?

oshiho3:
> O_o Coveralls seems to be confused. How could adding 3 lines decrease coverage by 50%?

I think it's me including <code>require 'pry'</code> accidentally and caused Travis to fail. Just removed the line in the last push.

The latest code now passes the Travis test, but hmm it must be using the sql search which I didn't mean to really. So it probably means the elasticsearch option is off on Travis. I assumed Travis uses staging or prod env variables but is actually using dev or test and elasticsearch is turned off by default on these environment.

If anyone can suggest a good way to set this <code>GROWSTUFF_ELASTICSEARCH</code> env variable when it's running on Travis, it'll appreciated. I wonder it should be in .travis.yml...

Skud:
Yes you can set environment variables on travis, via .travis.yml.  You can also say, "run this under both environments" if you want!  I think it might be good to run our test suite against both ES and postgres searches, fwiw.

oshiho3:
> We really ought to have a .approved scope and use it for this, I think!...

Make sense. updated the code as suggested.

oshiho3:
> Yes you can set environment variables on travis, via .travis.yml. You can also say, "run this under both environments" if you want! I think it might be good to run our test suite against both ES and postgres searches, fwiw.

OK. I updated .travis.yml to incorporate this.

PR #723, opened 2015-02-26
More tests

cesy:
Not linked to an issue, just trying to increase our Coveralls stats as a break from the deletion work.


cesy:
I have a Coveralls question, though. This page: https://coveralls.io/builds/2003155/source?filename=app%2Fcontrollers%2Fproducts_controller.rb shows really bad coverage of the products functionality. Yet when I look in the code, we have a feature test which covers all of these. Why is Coveralls not seeing that?

maco:
It looks like it only counts the "def" line of each function instead of the
whole function O_o

PR #726, opened 2015-03-05
Crop accept/reject notification improvements

maco:
http://talk.growstuff.org/t/please-test-requesting-new-crops-new-seed-attributes/223/17 asks for the notifications to be made a little friendlier and include the reject reason in the email beyond just the word "other"

### DIFF
--- a/app/models/crop.rb
+++ b/app/models/crop.rb
@@ -299,6 +299,14 @@ class Crop < ActiveRecord::Base
     end
   end
 
+  def rejection_explanation
+    if reason_for_rejection != "other"
+      return reason_for_rejection
+    else
+      return rejection_notes
+    end
+  end
+
   # Crop.search(string)
   def self.search(query)
     if ENV['GROWSTUFF_ELASTICSEARCH'] == "true"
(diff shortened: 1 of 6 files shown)


app/views/notifier/crop_request_rejected.html.haml
```diff
@@ -2,8 +2,11 @@
...
 
 %p
-  Your request for the new crop: #{link_to @crop.name, crop_url(@crop)} has been rejected for the following reason.
+  We're sorry, but your request for the new crop: #{link_to @crop.name, crop_url(@crop)} has been rejected for the following reason:
 
-  = @crop.reason_for_rejection
+  - if @crop.reason_for_rejection != "other"
+    = @crop.reason_for_rejection
+  - else
+    = @crop.rejection_notes
```
tygriffin: How about encapsulating this little bit of logic into a convenience method on the model called something like `rejection_explanation`?

maco:
Oh, I should add a test for the helper function.

app/models/crop.rb
```diff
@@ -299,6 +299,14 @@ def add_alternate_names_from_csv(alternate_names)
     end
   end
 
+  def rejection_explanation
+    if reason_for_rejection != "other"
```
Skud: small stylistic thing: I would reverse the clauses here to avoid using !=

Issue #575, opened 2014-12-12
One-click email unsubscribe


Skud:
At present you need to sign in to the site to unsubscribe from email notifications.  We should set up a one-click way of doing this.

oshiho3:
I'm working on this.

oshiho3:
How do we want this to behave?

<strong>1) Which type of emails do we want to include one-click-unsubscription link? eg. newsletter, planting reminders, inbox notifications</strong>

<strong>2) By clicking the link on an email, do we want</strong>

a) to turn off the particular email type only. 
eg. when clicking the "unsubscribe" link on planting reminder email, only planting reminder will be turned off.

b) to show a simple page with tick box options for disabling/enabling notifications.
eg. when clicking the link on planting reminder email, it opens up a page with tick boxes for planting reminder, inbox notification and newsletter. The user is only allowed to turn on/off email notifications but no other actions are permitted unless user manually logs in.

and/or

c) to login the user automatically and show edit email tab of user profile edit page. This will allow access to the site anywhere the user has access to.

Skud:
1. Everything that is sent as a transactional email by Growstuff.  This does *not* include the newsletter but does include all other types of emails.  (The newsletter has its own one-click unsubscribe, through Mailchimp.)

2. That particular type, I think.  It would be good to have a link that says "Unsubscribe" and then you click it and it says "You have been unsubscribed from planting reminder emails.  If you wish to unsubscribe from other Growstuff emails, you may do so via your settings page."  (with a link for them to login and go to the settings page)

We *absolutely* do not want option C.  For one thing, it's not one-click; for another, it's a security risk if any emails are forwarded.

Speaking of security... the obvious thing is to have a link something like:

http://growstuff.org/member/unsubscribe?type=planting_reminder&member=Skud

(or possibly email=skud@example.com instead of member=Skud)

However, this would allow people to switch off emails for other people if they knew their username or email address, which would be a bad thing.  We should consider using a string that is not guessable, but which we can look up in the db.  For instance, hashing their email address with one of the timestamps on their member object, eg. the created_at or confirmed_at timestamps.  That's probably not Fort Knox level security but should be enough for these purposes.

Example:

Given a URL like: http://growstuff.org/member/unsubscribe?type=planting_reminder&member=Skud&check=4gsf7d689gsfd78tgsfgydbso7fs

    who = Member.find_by(:login_name => params[:member])
    if params[:check] == some_hash_function(who.created_at)
[...]

PR #728, opened 2015-03-08
#575 One click email unsubscribe

oshiho3:
Implementation for #575. Please review.


app/controllers/members_controller.rb
```diff
@@ -49,6 +49,27 @@ def view_followers
...
+
+  def unsubscribe
+    begin
+      verifier = ActiveSupport::MessageVerifier.new(ENV['RAILS_SECRET_TOKEN'])
+      decrypted_message = verifier.verify(params[:message])
+
+      @member = Member.find(decrypted_message[:member_id])
+      @type = decrypted_message[:type]
+      @member.update_attributes(@type => false)
+
```
tygriffin: *Nitpicky note*: Instead of `update_attributes` we can just use `update`.

tygriffin:
Nice! I feel like I'm learning so much from your PR's these days :)

app/controllers/members_controller.rb
oshiho3: good suggestion! Just learned <code>update_attribetes</code> is an alias of <code>update</code>!

oshiho3:
Thanks @tygriffin for checking my code! You're very good at in motivating me :)

app/views/notifier/notify.html.haml
```diff
@@ -18,6 +18,7 @@
   %br/
   = link_to "View this message in your inbox", notification_url(@notification)
   %br/
-  = link_to "Turn off these notifications", edit_member_registration_url
+  = link_to "Turn off these notifications", unsubscribe_member_url(@signed_message)
```
Skud: I'm happy to merge this PR but I would like this string changed to "Unsubscribe from these notifications" (with the word "Unsubscribe" linked) as I think that's what people scan for when looking for a link to click.

Skud: Or even better, "Unsubscribe from direct message notifications" for this one, "... from planting reminders" for planting reminders, etc.  So it's clear you're unsubscribing from one particular kind of message.

oshiho3:
Hi @Skud, I was hoping someone's gonna say that ;-)
I modified the wording accordingly.

cesy:
This carries the security risk that if you forward an email, then someone can unsubscribe you, right? I think that's okay, though. Can I get a second opinion from @pozorvlak or @maco before merging? This looks good to me, though.

maco:
Yes, the forward unsubscribe possibility exists, but it is an extremely
common thing on the web, and as a user I'd rather someone accidentally
unsubscribe me than run into the situation (again! I've hit this with other
sites) of hitting "unsubscribe" and it saying "log in" and me going
"but...I don't...password... ugh, I'll just mark them as spam rather than
deal with figuring out how to log in and unsubscribe."

I believe @pozorvlak already said he agree that the forward unsubscribe
thing being a possibility isn't a problem for merging

pozorvlak:
Yep, looks good to me!

Issue #621, opened 2015-01-11
Limit the vertical space taken by crop varieties in sidebar


Skud:
It would be nice if we could limit the vertical space taken by the crop varieties in the sidebar of the crop detail page, eg. http://growstuff.org/crops/tomato

We could maybe do this by showing the first N and then having a "show NNN more" link which opens out the rest of them.

oshiho3:
I'm going to do this one.

maco:
this was merged, so i'm closing the bug

PR #729, opened 2015-03-12
#621 Crop varieties sidebar

oshiho3:
Please review.

Implemented #621 with javascript as I didn't realize rails-y implementation was preferred over javascript here and I wanted the toggle to look smooth instead of refreshing the whole page.
Let me know if you want me to rewrite using different technology etc.

Thanks!

### DIFF
--- a/spec/features/crops/crop_detail_page_spec.rb
+++ b/spec/features/crops/crop_detail_page_spec.rb
@@ -4,6 +4,61 @@ feature "crop detail page" do
 
   let(:crop) { FactoryGirl.create(:crop) }
 
+  context "varieties" do
+    let!(:roma1) { FactoryGirl.create(:crop, :name => 'Roma tomato 1', :parent => crop) }
+    let!(:roma2) { FactoryGirl.create(:crop, :name => 'Roma tomato 2', :parent => crop) }
+    let!(:roma3) { FactoryGirl.create(:crop, :name => 'Roma tomato 3', :parent => crop) }
+    let!(:roma4) { FactoryGirl.create(:crop, :name => 'Roma tomato 4', :parent => crop) }
+
+    scenario "The crop has 4 varieties" do
+
+      visit crop_path(crop)
+
+      # It lists all 5 items (note: including the top level item.)
+      # It DOES NOT have "Show all/less" toggle link
+      expect(page).to have_css('li', text: /tomato/i, count: 5)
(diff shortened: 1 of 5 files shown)


spec/features/crops/crop_detail_page_spec.rb
```diff
@@ -4,6 +4,61 @@
...
+    let!(:roma2) { FactoryGirl.create(:crop, :name => 'Roma tomato 2', :parent => crop) }
+    let!(:roma3) { FactoryGirl.create(:crop, :name => 'Roma tomato 3', :parent => crop) }
+    let!(:roma4) { FactoryGirl.create(:crop, :name => 'Roma tomato 4', :parent => crop) }
+
+    scenario "The crop has 4 varieties" do
+
+      visit crop_path(crop)
+
+      # It lists all 5 items (note: including the top level item.)
+      # It DOES NOT have "Show all/less" toggle link
```
tygriffin: You might use a `within` block to focus on that .crop-hierarchy class.

spec/features/crops/crop_detail_page_spec.rb
```diff
@@ -4,6 +4,61 @@
...
+      expect(page).to have_css('li', text: /tomato/i, count: 5)
+      expect(page).to have_no_css('button', text: /Show all+/i)
+      expect(page).to have_no_css('button', text: /Show less+/i)
```
tygriffin: Does it make more sense to use `have_selector` and `have_no_selector` here?

spec/features/crops/crop_detail_page_spec.rb
oshiho3: Good suggestion, thanks! Done now.

spec/features/crops/crop_detail_page_spec.rb
oshiho3: Thanks, done. I guess I was originally trying to match its classes as well so I thought I needed have_css, but decided to omit it. So you're right have_selector does the job.

oshiho3:
Code has been updated to incorporate @tygriffin's review and also
- Added a couple of more test scenarios
- Made crop/varieties partial to accept <code>max</code> param which specifies how many items can be seen without "Show all" button
- Number of crops to include the number of descendants below child as well

pozorvlak:
Looks good to me - moving into release 9.

Issue #557, opened 2014-12-12
Show roles (admin, crop wrangler) on profile page


Skud:
If someone is an admin or crop wrangler this should show on their profile page.

jacemonje:
https://github.com/Growstuff/growstuff/pull/754

PR #735, opened 2015-04-03
Added roles to profile page

cen10:
Issue #557 

I couldn't get the roles to show up directly underneath the login name when using 'content_for' so I changed it to an h1 instead (I'm not well-versed in haml). Wasn't sure if you envisioned people holding multiple roles, so the list_roles method accounts for that.

This is my first pull request, so your feedback/corrections would be appreciated!

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -107,6 +107,15 @@ class Member < ActiveRecord::Base
     roles.any? { |r| r.name.gsub(/\s+/, "_").underscore.to_sym == role_sym }
   end
 
+  def list_roles
+    roles = ""
+    self.roles.each do |role|
+      roles += role.name + ", "
+    end
+    roles.slice!(-2..-1)
+    roles
+  end
+
   def current_order
     orders.where(:completed_at => nil).first
   end
--- a/app/views/members/_account.html.haml
+++ b/app/views/members/_account.html.haml
@@ -1,6 +1,9 @@
 %h3 Account details
 
 %p
+  %strong= member.list_roles
+
+%p
   %strong Member since:
   = member.created_at.to_s(:date)
 
@@ -8,4 +11,3 @@
   %strong Account type:
   = member.account_type
   account
-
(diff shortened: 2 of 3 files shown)


maco:
What's going on with ```content_for :title``` is that its location is already set in https://github.com/Growstuff/growstuff/blob/dev/app/views/layouts/application.html.haml

I wonder if a string including a ```<br/>``` could go in the ```:subtitle```?

cen10:
Thanks, I'll look into that. I'm getting "unknown type of %string" and "unexpected end of input" syntax errors whenever I try to include anything either on the same line as `content_for :subtitle` or below it using a block.

Skud:
We definitely don't want to mess with the title blocks.  We are trying to standardise the way we display headers (title, subtitle, button bar) across the website and we shouldn't put anything else in that area without some high-level design discussion.

I'd rather these roles just showed in the sidebar somewhere (maybe near the account status and similar).

cen10:
I moved the roles directly above "Member since:" in the sidebar. Let me know what you think.

Skud:
This looks good to me now, but if this is your first commit, can you please add yourself to CONTRIBUTORS.md?  Then I think it's good to merge :)

app/models/member.rb
```diff
@@ -107,6 +107,15 @@ def has_role?(role_sym)
     roles.any? { |r| r.name.gsub(/\s+/, "_").underscore.to_sym == role_sym }
   end
 
+  def list_roles
```
Skud: Wait, sorry!  I said I think this is good to merge but I was just looking at the views.

I don't think we need this method here.  Why can't we just use join(", ") here, as we do in many other views?

pozorvlak:
Another attempt to address the same issue: #754.

cesy:
#787 is the latest one related to this issue. Does that mean this pull request should be closed as superseded by that, or is there anything from here that we need to keep/save? Sorry about the delays and confusion, @cen10.

pozorvlak:
I think we should probably close this as superseded by #787. @cen10, thanks for your contribution, and I'm sorry for the delay!

Issue #733, opened 2015-03-22
More representative Facebook thumbnail


emhoracek:
When I share the http://growstuff.org link on Facebook to invite friends to visit the page, the thumbnail that appears doesn't represent the site very well. I would think it would be the Growstuff logo or at least a picture of a plant. But instead it looks like its somebody's randomly generated avatar? Thanks Facebook. :)

![growstuff](https://cloud.githubusercontent.com/assets/5353499/6770049/cd0eb2d6-d089-11e4-9fd3-cf6fe0ef4130.png)

This looks like [a good how-to article](http://cypressnorth.com/social-network-media-marketing/how-to-customize-preview-thumbnail-images-for-pages-on-your-website/).

This is super minor, but I think a good thumbnail on Facebook links makes people more likely to share and makes friends more like to click on the shared links.

cesy:
Sadly it's still using members' avatars instead of the logo.

cesy:
http://www.limecanvas.com/how-to-make-facebook-use-a-custom-thumbnail-image-for-your-website/ gives another way to try and force it.

CloCkWeRX:
https://developers.facebook.com/tools/debug/og/object?q=growstuff.org - "The 'og:image' property should be explicitly provided, even if a value can be inferred from other tags." seems to be worth looking at

CloCkWeRX:
```
<html prefix="og: http://ogp.me/ns#">
```
Seems to be missing, which means the opengraph stuff may not kick in straight away

CloCkWeRX:
Argh, so close.
https://developers.facebook.com/tools/debug/og/object?q=http%3A%2F%2Fstaging.growstuff.org%2F

```
og:image could not be downloaded or is too small	og:image was not defined, could not be downloaded or was not big enough. Please define a chosen image using the og:image metatag, and use an image that's at least 200x200px and is accessible from Facebook. 
```
It's clearly not accurate when it says:
```
Image 'http://staging.growstuff.org/assets/growstuff-apple-touch-icon-precomposed-dc69b96d88315604421071635f3a2e36.png' will be used instead. 
```

CloCkWeRX:
I'm confused - the debugging tool shows me a share link that is quite reasonable, but the result I get seems wrong when I share just http://staging.growstuff.org/

If I share http://staging.growstuff.org/?hello ; never having seen that URI before I get a quite reasonable preview snippet with the image we want.

[...]

PR #739, opened 2015-04-28
add image to meta tag

ethulia:
Pretty new to this, but an attempt at fixing https://github.com/Growstuff/growstuff/issues/733

Let me know if the _meta.html file isn't the right place to put the tag!


pozorvlak:
Looks good, and I think `_meta.html` is exactly the right place! Could you add a test for the image's presence? We're phasing out view tests in general, but I think it would be fine in this case to add another test to `views/layouts/_meta_spec.rb` along the lines of the existing ones.

[If the answer's "no, I'm not that comfortable with Ruby yet", no problem, we'd be happy to help you out!]

ethulia:
Sure, will add the test!

pozorvlak:
Great!

ethulia:
I can't seem to get the tests to run locally, so not sure if I can write the test. Can someone else add it in?

pozorvlak:
Sure, but I'd like to work out why you can't get the tests to run. Can you copy-and-paste the command you ran and the output you got here?

Skud:
Hi @ethulia, this looks good to me :)  But as @pozovlak says, can you let us know which tests are failing for you so we can help you get that sorted out?  Also, please add yourself to CONTRIBUTORS.md!

app/views/layouts/_meta.html.haml
```diff
@@ -1,6 +1,7 @@
 %head
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta property="og:image" content="meta_image.png" />
```
Skud: this is a small thing, but perhaps we could rename this something like growstuff_logo_ogimage.png or something?  "meta_image" doesn't really tell us what it's for.

pozorvlak:
It looks like we merged someone else's fix to this issue, in #746 - sorry! Closing this as no longer needed :-(

Issue #634, opened 2015-01-15
prevent search engines from indexing staging site


Skud:
What it says on the tin :)

yez:
@Skud Glad you accepted my contribution! Is there more work to be done on this issue?

cesy:
Fixed by #740.

PR #740, opened 2015-05-05
Feature/robots dissalow on staging

yez:
This pull request addresses #634 

A per-subdomain robots.txt pattern has been introduced. The default is to load the production robots.txt file (currently empty with no guidelines for search engines). Staging version of the robots.txt file prevents all scraping from robots on entire site.


tygriffin:
This looks good to me. Note: even though we are generally deprecating controller specs, this use seems appropriate.

Issue #558, opened 2014-12-12
Cleanup tests to use config.currency rather than AUD


Skud:
Some of the tests for the shop and orders system are checking that prices are displayed in AUD, rather than checking them against the configured currency. This should be refactored.

krio:
I believe the referenced issue allows this to be closed. I don't see anymore tests that reference a currency directly without using the config.

cesy:
Yeah, this should have auto-closed but Github is fussy about the wording of the link.

PR #744, opened 2015-05-25
#558 Currency refactoring in specs.

andrba:
Issue #558
I noticed that the application_helper.rb spec file wasn't following the naming conventions. I had to rename it.


db/schema.rb
```diff
@@ -405,7 +405,6 @@
     t.datetime "updated_at"
     t.string   "slug"
     t.integer  "forum_id"
-    t.integer  "parent_id"
```
pozorvlak: This change looks like it shouldn't be here - can you remove it?

pozorvlak:
Looks good to merge once you've removed the stray change to `schema.rb`.

andrba:
@pozorvlak could you review it again? I removed the schema change

Issue #718, opened 2015-02-18
Clearly make optional fields optional on planting/harvest/seed/garden forms


Skud:
These four forms (planting, harvest, seed, and garden) are long and include many optional fields.

The simplest incremental incremental to these is to simply put "Optional" next to the optional fields on the form.

The required fields are:

* Garden - name
* Planting - crop and garden
* Harvest - crop
* Seeds - crop

A further improvement might be:

If the object is new (as opposed to editing an existing one), hide the least important/relevant fields behind a "Show more options" link, and use jquery(?) to display them only if asked for.

Fields to hide behind this link might include:

* Garden: area, active
* Planting: planted from, sun/shade, mark as finished (move "description" up higher in this case)
* Harvest: hmmm I think it's all fine (as long as optional fields are marked optional)
* Seeds: move description up under quantity, then everything below that (plant before, days until maturity, organic/gmo/etc) that can be hidden.

I'd be happy with the simpler version of this and delighted with the slightly more complex one.  I don't think it's worth doing anything super fancy as the long-term fix to this is #672.

pozorvlak:
The UI pattern I'm used to seeing is "* denotes required field", with all non-starred fields being optional.

maco:
Ditto

Skud:
I don't think that's necessarily obvious enough for these forms.

A.

On 20/02/2015 9:39 am, pozorvlak wrote:
>
> The UI pattern I'm used to seeing is "* denotes required field", with 
> all non-starred fields being optional.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/Growstuff/growstuff/issues/718#issuecomment-75154069>.
>

-- 
Alex "Skud" Bayley
skud@growstuff.org
http://growstuff.org/

pozorvlak:
OK, but I think it's more common to mark the required fields than the
optional ones.

sha1sum:
I completely overlooked the commits by @pozorvlak and just issued a pull request for the same issue. My mistake :)

pozorvlak:
Fixed in #779.

PR #749, opened 2015-06-13
Issue 718 Clearly Make Optional Fields Optional

achalc:
Added asterisks to required fields and hid requested optional fields behind a button to resolve issue #718


app/assets/javascripts/plantings.js.coffee
```diff
@@ -3,4 +3,4 @@
 # You can use CoffeeScript in this file: http://jashkenas.github.com/coffee-script/
 
 jQuery ->
-  $('.add-datepicker').datepicker('format' : 'yyyy-mm-dd')
+  $('.add-datepicker').datepicker('format' : 'yyyy-mm-dd')
```
cesy: This appears to have lost the final blank line which Github doesn't like. Very minor, but if you're making any other changes, it would be good to fix.

cesy:
How does this work if someone has Javascript turned off? It looks like since the hiding is done with CSS and the showing with JS, they won't be able to get at the fields. Could we do the hiding onpageload with JS instead?

cesy:
Welcome to the project, by the way, and sorry it's taken so long to review this.

pozorvlak:
I'm not happy about the use of `sleep` in tests - this has a tendency to make tests flaky (sometimes they pass, sometimes they don't) or slow or both. [This article suggests some alternatives](http://artsy.github.io/blog/2012/02/03/reliably-testing-asynchronous-ui-w-slash-rspec-and-capybara/).

pozorvlak:
Superseded by #779 - @achalc, thank you for your hard work, and I'm sorry we didn't end up incorporating it.

PR #750, opened 2015-06-17
Added search form on the places page.

GabrielSandoval:
places#show has a search form but places#index doesn't have. I added a search form in the index page of the 'places' so that the users may easily search the place they want to see.


pozorvlak:
Why are you gitignoring and deleting `config/database.yml`? As you can see, our continuous integration tests rely on it...

pozorvlak:
Superseded by #751.

PR #751, opened 2015-06-21
Added search form on the places index page.

GabrielSandoval:
image

places#show has a search form but places#index doesn't have. I added a search form in the index page of the 'places' so that the users may easily search the place they want to see.

@pozorvlak Sorry I messed up my last PR. Here is a new one. I left the gitignore untouched this time.

### DIFF
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -15,3 +15,16 @@ if Geocoder.config.lookup != :test
   Geocoder.configure(:lookup => :nominatim)
 end
 
+Geocoder::Lookup::Test.add_stub(
+  "Philippines", [
+    {
+      'latitude'     => 12.7503486,
+      'longitude'    => 122.7312101,
+      'address'      => 'Manila, Mnl, Philippines',
+      'state'        => 'Manila',
+      'state_code'   => 'Mnl',
+      'country'      => 'Philippines',
+      'country_code' => 'PH'
+    }
+  ]
+)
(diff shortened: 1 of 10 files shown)


pozorvlak:
Looks good! However, it should be covered by a test. I would say "just copy the existing test for the Show page", but

1. there isn't one
2. the [existing tests for the Show page](https://github.com/Growstuff/growstuff/blob/dev/spec/views/places/show.html.haml_spec.rb) are in a deprecated style.

So you'd need to write a whole new file of "feature" tests for the places-index page. Do you feel confident to do that? The good news is that feature tests are [more straightforward to write](https://github.com/Growstuff/growstuff/blob/dev/spec/features/members_list_spec.rb) - that's why we deprecated view and controller tests!

GabrielSandoval:
Thanks @pozorvlak! I'll start working on this as soon as I'm confident about testing. I'm at bit new to rails and I'm excited to try this one.

GabrielSandoval,
AELOGICA Intern

GabrielSandoval:
Hi @pozorvlak,

I added the feature test for the places search form. Please see if it's all good.

Thanks!

config/initializers/geocoder.rb
```diff
@@ -15,3 +15,16 @@
   Geocoder.configure(:lookup => :nominatim)
 end
 
+Geocoder::Lookup::Test.add_stub(
```
pozorvlak: Could you please move this to `config/environments/test.rb` along with the other Geocoder stubs?

GabrielSandoval: Hi, I moved the stubs to config/environments/test.rb. Just tell me if there's anything more needed to be done.

pozorvlak:
Excellent - thanks!

GabrielSandoval:
Thanks @pozorvlak!

Issue #546, opened 2014-12-12
Show last login date on member profile


Skud:
When I look at someone's profile page, I should see the last time they logged in/visited the site.

Note: Miles has already done some work on this and you can find it in his github repo.  He writes:

The old code (with history rewritten to apply on top of the non-broken version, rather than the other way around) can still be seen at https://github.com/pozorvlak/growstuff/compare/story_profile, if whoever does the "show last login date" story wants to build on it.

pozorvlak:
Fixed in #753.

PR #755, opened 2015-06-23
Show last login on member profile

jacemonje:
added last login time in response to https://github.com/Growstuff/growstuff/issues/546

### DIFF
--- a/app/views/members/_account.html.haml
+++ b/app/views/members/_account.html.haml
@@ -10,5 +10,16 @@
   account
 
 %p
+<<<<<<< HEAD
   %strong Administrator:
   = member.has_role? :admin
+=======
+  %strong Member Role:
+  - if member.has_role? :admin
+    Administrator
+  - else
+    Crop Wrangler
+%p
+  %strong Last login:
+  #{member.last_sign_in_at}
+>>>>>>> 5a94502... added last sign in
--- a/app/views/members/_account.html.haml
+++ b/app/views/members/_account.html.haml
@@ -9,3 +9,6 @@
   = member.account_type
   account
 
+%p
+  %strong Administrator:
+  = member.has_role? :admin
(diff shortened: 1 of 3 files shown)


jacemonje:
@pozorvlak thoughts?

app/views/members/_account.html.haml
```diff
@@ -9,3 +9,17 @@
   = member.account_type
   account
 
+%p
+<<<<<<< HEAD
```
pozorvlak: You've committed a git conflict marker (the `<<< === >>>` stuff). When you merge or rebase and git tells you that it couldn't resolve a conflict, you need to resolve these conflicts manually before continuing with the merge/rebase. Git doesn't care what you put in your files, but conflict markers are not syntactically correct HAML or Ruby code :-) I've written a [slightly more detailed explanation of conflict-resolution in Git here](http://wiki.growstuff.org/index.php/Git#Merging_in_Git) - could it be made clearer?

spec/features/following_spec.rb
```diff
@@ -80,7 +80,7 @@
         visit member_follows_path(member)
         expect(page).not_to have_content "#{other_member.login_name}"
         visit member_followers_path(other_member)
-        expect(page).not_to have_content "#{member.login_name}"
+        # expect(page).not_to have_content "#{member.login_name}"
```
pozorvlak: [Committing commented-out code is generally a bad idea](http://programmers.stackexchange.com/questions/45378/is-commented-out-code-really-always-bad) - if it doesn't need to be there then just delete it, because if we need it later we can get it back from the Git history. In this case, the code shouldn't be removed, because your code changes shouldn't have broken this test (but it appears they have, because the Git conflict markers mean your HAML code is no longer syntactically correct).

pozorvlak:
Closing this because it's no longer needed - sorry about that, and I hope you found my feedback useful!

jacemonje:
Hi! Thanks for the feedback! Just new to open-source. I apologize for my newbie mistakes. Thank you, though! Makes sense :)

pozorvlak:
No need to apologize - we were all beginners once :-)

Issue #678, opened 2015-01-24
Calculate "days until maturity" for a crop


Skud:
We should calculate how many days from planting (from seed) until harvest for a crop.

We can base this on various data:

- seed.days_until_maturity (mean/median/???) (added in #493)
- actual time from planting to harvest based on GS tracking (requires #501 and might be less accurate)

The former is definitely easier, and should probably be what we use for a first cut of this.

Realistically we're unlikely to be able to do this for at least, say, a few months after #493 is pushed to production.

GabrielSandoval:
Hi @Skud,
I'm an intern at AELOGICA and I'd like to work on this one. I'll be submitting a PR soon.
Thanks.

GabrielSandoval:
/places page:
![screen shot 2015-06-24 at 1 50 09 pm](https://cloud.githubusercontent.com/assets/6015897/8323513/0e406514-1a78-11e5-9584-ecc4e59c9d73.png)

Hi @Skud,
I added this feature and redesigned the plantings index page (see photo above). I added a days_before_planting column on the plantings table. I based the calculations on the previous plantings' start and end date.

Also, I added the days_before_maturity attribute on the show page.
![screen shot 2015-06-24 at 2 01 52 pm](https://cloud.githubusercontent.com/assets/6015897/8323659/05bd2394-1a7a-11e5-84f0-75e39ceb5d5a.png)

Gabriel Sandoval,
AELOGICA Intern

PR #758, opened 2015-06-25
Gab/days before maturity (#678)

GabrielSandoval:
Old /plantings page:
plantings_old

New /plantings page:
plantings_new

Hi
I added this feature and redesigned the plantings' index page (see photo above). I added a days_before_planting column on the plantings table. I based the calculations on the previous plantings' start and end date.

Also, I added the days_before_maturity attribute on the show page.
screen shot 2015-06-25 at 12 33 32 pm

Gabriel Sandoval,
AELOGICA Intern


maco:
When you said "I added a days_before_planting column" did you mean days
before maturity? I'm not seeing anything about days until planting

GabrielSandoval:
Oh, I'm so sorry. Yes, I meant 'days_before_maturity'.

GabrielSandoval:
Hi @maco,

I'm wondering if you're interested to this feature. I'd like to know if I should push on to refactoring and making feature tests about this one.

Thanks.

Gabriel Sandoval,
AELOGICA Intern

pozorvlak:
Wow, looks great! "Yes please" to the refactoring and testing, and please also add yourself to CONTRIBUTORS.md.

GabrielSandoval:
Hi @pozorvlak,

I'm done with adding tests for this and I also fixed some minor bugs. Just tell me if there's anything more needed to be done.

Thanks!

Gabriel Sandoval,
AELOGICA Intern

app/views/plantings/_planting_progress.html.haml
```diff
@@ -0,0 +1,18 @@
+- if DateTime.now.to_date < planting.planted_at
+	= "Progress: 0% - not planted yet"
+	.progress
```
pozorvlak: I don't like the repetition here - could you factor out the progress bar into another partial? [Here's an example of how to pass variables when rendering partials](https://github.com/Growstuff/growstuff/blob/dev/app/views/alternate_names/show.html.haml#L3).

app/views/plantings/index.html.haml
```diff
@@ -22,40 +22,71 @@
...
-            - if ! planting.finished
-              = link_to "Mark as finished", planting_path(planting, :planting => {:finished => 1}),  :method => :put, :class => 'btn btn-default btn-xs append-date'
-          - if can? :destroy, planting
-            =link_to 'Delete', planting, method: :delete, data: { confirm: 'Are you sure?' }, :class => 'btn btn-default btn-xs'
+        %td
+          %ul{:style => "list-style-type:none"}
+            %li
+              Crop name:
+              =link_to planting.crop.name, planting.crop
+            - if planting.finished?
```
pozorvlak: There's a lot of moderately complex logic shared between here and your changes in  app/views/plantings/show.html.haml: it would be better if you could factor out the common logic into a helper function, or perhaps a method on the Planting model.

app/views/plantings/index.html.haml
```diff
@@ -22,40 +22,71 @@
...
+    - @plantings.each.with_index do |planting, index|
       %tr
-        - unless @owner
-          %td= link_to planting.owner.login_name, planting.owner
-        %td= link_to planting.crop.name, planting.crop
-        %td= link_to planting.garden.name, planting.garden
-        %td= planting.quantity
-        %td= planting.planted_at
+        %td= "##{index+1}"
+        %td= link_to image_tag((planting.default_photo ?  planting.default_photo.thumbnail_url : 'placeholder_150.png'), :alt => planting.crop_id, :class => 'img'), planting
```
pozorvlak: I think the code in the body of this loop should be moved into plantings/_thumbnail.html.haml, where it can replace the rather dull planting-summary that we're currently using in Garden detail views. Then replace the loop body here with a call to the thumbnail partial :-)

pozorvlak:
Great stuff - we're definitely getting there! Just a couple of refactorings as described above, and I think this will be good to merge.

GabrielSandoval:
Hi @pozorvlak,

I finished the refactoring on the index, show, thumbnails and progress views. I created a helper file for plantings. Also, as requested, I replaced the old dull plantings/_thumbnail view with the new one.

I had some issues regarding the test about the plantings/_thumbnail view but since those tests are deprecated, I decided to comment them out to get rid of errors.

pozorvlak:
When I said "the body of the loop", I meant the block *inside* the `each` - all our other thumbnail partials only show one object, not a list of objects. Also, I belatedly see you've used HTML tables rather than Bootstrap CSS classes - I don't know if that's likely to cause problems. @cesy @maco @Skud? It would be better if you could do it in a way consistent with the rest of the site's code, but I don't know how hard that would be.

Committing commented-out code is a [really bad idea](http://programmers.stackexchange.com/questions/45378/is-commented-out-code-really-always-bad) - if we need old code back again, we have version control :-) In this case it's unsurprising the tests failed, since you were changing the behaviour of that partial quite dramatically. Since your new code is covered by feature tests, I think it's fine to just delete the thumbnail view spec file.

> I'm looking forward to this getting merged.

Me too - I'm sorry this is taking so long! I think it's been a valuable process, though.

I'll be away for the next few days in a region without phone or Internet access, so won't be able to look at this again until Monday. Hopefully one of the other committers will be around in the meantime :-)

maco:
I'll keep an eye out while you're gone.

The table display is going to be bad for screen readers, especially because
you have a ul in a td several times. The entire ul will be read top to
bottom instead of matching up label and value. I would use a definition
list (dl) and some css to display the planting date etc.

GabrielSandoval:
Thanks for the tip @maco and @pozorvlak !  I'll be working on it.

GabrielSandoval:
Hi @maco , @pozorvlak ,

I've deleted the thumbnail view spec file, changed the ul into dl, made the thumbnail partial pass only a single object, and converted the html tables into bootstrap css classes. All tests passed.

I'll be waiting for your response. Thanks!

GabrielSandoval:
Thanks @maco !
I'll be looking for more issues to work on. :)

pozorvlak:
Hurrah - good work!

GabrielSandoval:
Thanks too @pozorvlak ! I've learned so much from you guys!

Issue #503, opened 2014-12-12
Refactor: "let" in rspec tests


Skud:
I just learned about "let", which looks like being a lot more efficient than some of our current before(:each) blocks.  This is a chore to refactor to use it where appropriate.

Some more info: http://stackoverflow.com/questions/5359558/when-to-use-rspec-let

pozorvlak:
Current status:

Directory | Contains `before(:each)` | Deprecated? | Still to do?
--------|------------|-----------|------------
controllers | Yes | Yes | No
factories | No | No | No
features | No | No | No
helpers | Yes | No | Yes
lib | No | No | No
mailers | No | No | No
models | Yes | No | Yes
requests | Yes | Yes | No
routing | No | No | No
support | Yes | No | Yes
views | Yes | Yes | No

`spec/requests` looks to be entirely autogenerated, does only the most superficial tests, and should probably be deleted.

pozorvlak:
Looking through `spec/models`, it looks like we could further `let`ify the following files:

 - comment_spec.rb
 - crop_spec.rb
 - follow_spec.rb
 - order_spec.rb

There are a few other `before` blocks, but they can't be replaced by `let`. Huge thanks to @CjayBillones, who did the vast majority of the refactoring in #762.

PR #760, opened 2015-07-06
Refactor spec tests

CjayBillones:
Refactor spec by changing before(:each) into let as requested in #503 

@pozorvlak


pozorvlak:
Thanks!

PR #761, opened 2015-07-07
Refactor spec/support

CjayBillones:
Changed before(:each) into before as requested in #503

### DIFF
--- a/spec/support/controller_macros.rb
+++ b/spec/support/controller_macros.rb
@@ -1,9 +1,11 @@
 # Taken unashamedly from https://github.com/plataformatec/devise/wiki/How-To%3a-Controllers-and-Views-tests-with-Rails-3-%28and-rspec%29
 module ControllerMacros
   def login_member(member_factory=:member)
+    
+  	let(:member) { member = FactoryGirl.create(member_factory || :member) }
+
     before(:each) do
       @request.env["devise.mapping"] = Devise.mappings[:member]
-      member = FactoryGirl.create(member_factory || :member)
       sign_in member
     end
   end
--- a/spec/support/elasticsearch_helpers.rb
+++ b/spec/support/elasticsearch_helpers.rb
@@ -10,7 +10,7 @@ end
 RSpec.configure do |config|
   config.include ElasticsearchHelpers
 
-  config.before(:each) do
+  config.before(:all) do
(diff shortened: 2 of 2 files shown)


CjayBillones:
@pozorvlak Is this fine? The let is not working and I'm not sure if doing it this way makes any difference.

pozorvlak:
I don't think this improves things - the efficiency gain comes from using `let` (which is cached) rather than `before`, and I can't find the meaning of `before` with no `:each`/`:all`/`:suite` documented. So we'd be making the code harder to understand and not gaining any performance.

spec/support/elasticsearch_helpers.rb
```diff
@@ -10,7 +10,7 @@ def sync_elasticsearch(crops)
 RSpec.configure do |config|
   config.include ElasticsearchHelpers
 
-  config.before(:each) do
+  config.before do
```
pozorvlak: I don't think there's any benefit to replacing `before(:each)` with `before` here - though maybe we could make it a `before(:all)` block instead of a `before(:each)` block?

spec/support/controller_macros.rb
```diff
@@ -1,7 +1,7 @@
 # Taken unashamedly from https://github.com/plataformatec/devise/wiki/How-To%3a-Controllers-and-Views-tests-with-Rails-3-%28and-rspec%29
 module ControllerMacros
   def login_member(member_factory=:member)
-    before(:each) do
+    before do
       @request.env["devise.mapping"] = Devise.mappings[:member]
       member = FactoryGirl.create(member_factory || :member)
```
pozorvlak: I think this line can be replaced with a `let`, but I think the `@request.env` line probably has to stay as it is.

CjayBillones:
@pozorvlak fixed! :)

cesy:
Hi @CjayBillones, looking at the current changes I think we're not getting any benefit here, based on @pozorvlak's comments? Would you mind taking another look to see if we can change it to let, please? Given your comment didn't have any commits with it, I suspect some work you've done hasn't pushed to Github properly.

CjayBillones:
@cesy oops my bad. Seems like I forgot to push the modifications in my machine that I use at work. Will try to fix it with my machine at home. Sorry!

CjayBillones:
@cesy @pozorvlak I think it's all good now. :)

PR #762, opened 2015-07-07
Refactor spec/models

CjayBillones:
Changed before(:each) into let as requested in #503 .

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -2,36 +2,37 @@ require 'rails_helper'
 require 'cancan/matchers'
 
 describe Ability do
-  before(:each) do
-    @member = FactoryGirl.create(:member)
-    @ability = Ability.new(@member)
-  end
+
+  let(:member) { FactoryGirl.create(:member) }
+  let(:ability) { Ability.new(member) }
 
   context "notifications" do
+
+    let(:notification) { FactoryGirl.create(:notification, :recipient => member) }
+
     it 'member can view their own notifications' do
-      @notification = FactoryGirl.create(:notification, :recipient => @member)
-      @ability.should be_able_to(:read, @notification)
+      ability.should be_able_to(:read, notification)
     end
 
     it "member can't view someone else's notifications" do
-      @notification = FactoryGirl.create(:notification,
+      notification = FactoryGirl.create(:notification,
         :recipient => FactoryGirl.create(:member)
       )
-      @ability.should_not be_able_to(:read, @notification)
+      ability.should_not be_able_to(:read, notification)
     end
     it "member can't send messages to themself" do
-      @ability.should_not be_able_to(:create,
+      ability.should_not be_able_to(:create,
         FactoryGirl.create(:notification,
-          :recipient => @member,
-          :sender => @member
+          :recipient => member,
+          :sender => member
         )
       )
     end
     it "member can send messages to someone else" do
-      @ability.should be_able_to(:create,
+      ability.should be_able_to(:create,
         FactoryGirl.create(:notification,
           :recipient => FactoryGirl.create(:member),
-          :sender => @member
+          :sender => member
         )
       )
     end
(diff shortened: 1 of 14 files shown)


CjayBillones:
@pozorvlak I also worked on refactoring the spec/models. Check it out if it's good to go. :)

spec/models/ability_spec.rb
```diff
@@ -39,245 +38,248 @@
...
-        @member.roles << @role
-        @cw_ability = Ability.new(@member)
+
+      let(:role) { FactoryGirl.create(:crop_wrangler) }
+
+      before do
+        member.roles << role
       end
 
+      let(:cw_ability) { Ability.new(member) }
```
pozorvlak: Now that `ability` is a local variable and populated lazily, I think you can use it instead of `cw_ability`.

spec/models/ability_spec.rb
```diff
@@ -2,36 +2,35 @@
...
-  end
+
+  let(:member) { FactoryGirl.create(:member) }
+  let(:ability) { Ability.new(member) }
 
   context "notifications" do
     it 'member can view their own notifications' do
-      @notification = FactoryGirl.create(:notification, :recipient => @member)
-      @ability.should be_able_to(:read, @notification)
+      @notification = FactoryGirl.create(:notification, :recipient => member)
```
pozorvlak: I think this would be a good opportunity to make `@notification` a local variable - IIRC when we wrote these tests we didn't understand Ruby's scoping rules very well :-)

spec/models/ability_spec.rb
```diff
@@ -39,245 +38,248 @@
...
-        @member.roles << @role
-        @admin_ability = Ability.new(@member)
+
+      let(:role) { FactoryGirl.create(:admin) }
+
+      before do
+        member.roles << role
       end
 
+      let(:admin_ability) { Ability.new(member) }
```
pozorvlak: See above comment regarding `cw_ability`.

spec/models/ability_spec.rb
```diff
@@ -39,245 +38,248 @@
...
 
     context "crop wrangler" do
-      before(:each) do
-        @role = FactoryGirl.create(:crop_wrangler)
-        @member.roles << @role
-        @cw_ability = Ability.new(@member)
+
+      let(:role) { FactoryGirl.create(:crop_wrangler) }
+
+      before do
```
pozorvlak: As I said on #761, I think `before(:each)` is more explicit and hence preferable to a bare `before`. The point of #503 wasn't to eliminate `before(:each)` for reasons of linguistic purity, it was to use the more efficient `let` construct wherever possible.

spec/models/ability_spec.rb
CjayBillones: @pozorvlak just to confirm, when you say "make @notification a local variable", is it the same as using let for the notification?

spec/models/ability_spec.rb
CjayBillones: On this one, do you mean that I should just simply change "let(:cw_ability) {Ability.new(member) }" into "let(:ability) {Ability.new(member) }" or do you mean something else?

CjayBillones:
@pozorvlak will you check this out again? I'm not sure if I understood your instructions correctly (I hope I did). :)

CjayBillones:
@cesy would you like to also check this one if I missed or misunderstood anything?

spec/models/ability_spec.rb
pozorvlak: No, I think you can remove the `let(:cw_ability)` line entirely and replace all references to `cw_ability` with `ability`. I could be wrong about that, though.

spec/models/ability_spec.rb
pozorvlak: Not really: I'm talking about [variable scope](http://www.techotopia.com/index.php/Ruby_Variable_Scope). Variables whose names start with "@" are *instance variables* - they live on an object, and can be accessed from any method of that object. But since `@notification` is only accessed from the block in which it's defined, we can replace it with a local variable `notification`, which is only visible from within that block. It's not a huge issue, but it would give our tests slightly better isolation from each other. Concretely, I mean "replace all uses of `@notification` with `notification` and don't introduce any new `let` constructs".

CjayBillones:
@pozorvlak I think it's already good to go. :)

Issue #549, opened 2014-12-12
On crop page, show if I have seeds


Skud:
If I'm looking at a crop page (eg. arugula) and I already have seeds for that crop, it should say "You have seeds for this crop" and link to the seed pages for them.

PR #765, opened 2015-07-08
Added feature and test for request #549

GabrielSandoval:
Hi,

screen shot 2015-07-08 at 8 07 20 am

To answer issue #549, I added a simple feature wherein the crop page shows whether a user has seeds for that crop or none.

### DIFF
--- a/app/helpers/crops_helper.rb
+++ b/app/helpers/crops_helper.rb
@@ -1,0 +1,17 @@
+module CropsHelper
+  def display_seed_availability(member, crop)
+    total_quantity = 0
+    member.seeds.each do |seed|
+      if seed.crop.name == crop.name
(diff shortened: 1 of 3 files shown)


maco:
@GabrielSandoval for future reference, put the word "fixes" in front of the issue number to make the issue automatically close on merge. I *think* it works whether you put that in your git commit message or in the PR description when you first submit it.

maco:
Oh, also, that git commit message "added feature and test for bug number" isn't something that's going to be easy to make sense of a year from now when we're trying to figure out why a certain line of code exists (at least, not without digging up the bug number). Please be more detailed about the actual functionality you're adding when writing your commit messages.

GabrielSandoval:
Noted! Thanks @maco .

PR #766, opened 2015-07-08
Added sunniness icons on plantings/_thumbnail and show page.

GabrielSandoval:
Hi,

screen shot 2015-07-08 at 10 05 43 am

screen shot 2015-07-08 at 10 11 32 am

I added these icons to reduce the heavy text display on plantings/_thumbnail. I think this would make it more presentable.

Any comments about this?

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -8,10 +8,13 @@
 development:
   secret_key_base: 'b1b67abb399261478f4721e704eb3851466daf60d9cd2b53a1839b056d641c4c1c2a476bcaf7addc6d6548926cfd32fa5a00a8de258880257ebb5a6fd86cb08f'
   # run 'rake secret' to generate your own
- 
+
 test:
   secret_key_base: 'be557aa019b181f201c9906663dbf8f22efb1b70b11f78035bfeda86aa7dcfd1efb184e2ee894a0ae0dc37fe67d311f38e7731fa16d8d595f2e1ef5447bae020'
   # run 'rake secret' to generate your own
- 
+
+staging:
+  secret_key_base: <%= ENV["RAILS_SECRET_TOKEN"] %>
+
 production:
   secret_key_base: <%= ENV["RAILS_SECRET_TOKEN"] %>


app/views/plantings/show.html.haml
```diff
@@ -32,7 +32,7 @@
       %p
         %b 
           = "Sun or shade?: "
-        = "#{display_sunniness(@planting)}"
+        = image_tag("#{@planting.sunniness}.png", :size => "25x25", :title => "#{@planting.sunniness}")
```
cesy: What's the reason for skipping the "not specified" here?

app/views/plantings/_thumbnail.html.haml
```diff
@@ -16,7 +16,7 @@
       %dt Finished on:
       %dd= "#{display_finished(planting)}"
       %dt Sun/shade?:
-      %dd= "#{display_sunniness(planting)}"
+      %dd= planting.sunniness.blank? ? "not specified" : image_tag("#{planting.sunniness}.png", :size => "25x25", :title => "#{planting.sunniness}")
```
cesy: Can we include alt text for screenreaders, please, not just title text?

spec/views/plantings/show.html.haml_spec.rb
```diff
@@ -39,12 +39,6 @@ def create_planting_for(member)
       )
     end
 
-    it "shows the sunniness" do
-      render
-      rendered.should have_content 'Sun or shade?'
-      rendered.should have_content 'sun'
-    end
-
```
cesy: Can we replace this with a check for the alt text and/or for the image, please?

cesy:
I like the idea of more pictures. I've added a few comments for specific details. It may also be worth having the icon alongside the text rather than instead of the text, at least to ease people into the change and get them used to the icons. The icon for shade is very clever, but leaves it inconsistent with the semi-shade one, so I'd appreciate a second opinion on that.

maco:
Perhaps an icon that's a sun with a ? inside would convey "not specified"
rather than switching to text?

cesy:
Yes, that would work.

spec/views/plantings/show.html.haml_spec.rb
GabrielSandoval: I can do that, but it says in the file that the test is deprecated. Will it be okay if add some feature tests instead?

GabrielSandoval:
Hi,

A s requested by maco, I created this "not specified" icon.
screen shot 2015-07-09 at 9 13 36 am

I still don't know what to choose between these two for the "semi_shade" icon. What do you guys think?

screen shot 2015-07-09 at 9 13 31 am

maco:
Yes, a feature test is the correct thing to do.

PR #770, opened 2015-07-08
Set staging secret key from environment

pozorvlak:
Currently the staging server is down because it can't find the secret key needed to sign cookies. This is related to the changes in #710.


pozorvlak:
...aaaaand it's back up :-)

PR #771, opened 2015-07-09
Remove one-off tasks that were run for Release 8.

pozorvlak:
The ElasticSearch index task was run, but failed because we didn't have an ElasticSearch instance to run it against.

There's also a potential speed improvement on the `growstuff:import_crops` rake task. Currently we load the cropbot user once per row, but we could load them once at the beginning. I don't know how much of a speedup this would give. Right now it doesn't matter too much, but we'll no doubt want to add crops in the future (and the same code is run by `rake db:seed`).


cesy:
Looks good, thanks.

Issue #568, opened 2014-12-12
After signout, only redirect to homepage if necessary


Skud:
When I sign out, I'm sent back to the homepage all the time.  I should only be sent there if the page I'm on isn't visible when logged out (eg. my Inbox).

PR #774, opened 2015-07-10
[#568] Signout redirects to last page visited

jacarandang:
Sign out redirects to last page visited however, it doesn't redirect to root_path if the page last visited needs authentication (e.g. edit account), instead it is redirected to the login page

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,6 +22,10 @@ class ApplicationController < ActionController::Base
     stored_location_for(:member) || root_path
   end
 
+  def after_sign_out_path_for(resource_or_scope)
+    request.referrer
+  end
+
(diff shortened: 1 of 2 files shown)


jacarandang:
#568

pozorvlak:
Looks good, but can you please add a feature test for this? See the `spec/features` directory for some examples.

jacarandang:
Added some test

pozorvlak:
Thanks!

CloCkWeRX:
This didn't seem to work on staging.

When signing out, you get redirected to the last page you were on, if it's something public, rather than always to the homepage
* Login
* Edit your profile
* Signout

Expected: redirect to /
Actual: prompted to login to access the member profile edit UI

PR #779, opened 2015-07-22
#718 optional and required fields

sha1sum:
Given that users are torn between marking required fields and marking optional fields, I decided to go with both and mark required fields with a red asterisk and denote optional fields with an "optional" placeholder or or used "include_blank: 'optional'" on select fields that previously had "include_blank: true". issue #718

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ custom_plan.rb
 zeus.json
 .bundle
 config/application.yml
+.idea/**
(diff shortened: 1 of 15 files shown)


pozorvlak:
Looks good, and works fine without JavaScript, but can you add some tests to ensure the optional fields are marked as such?

pozorvlak:
This PR just marks fields as optional/required; unlike #749, it doesn't attempt to hide optional fields. It may be possible to take that aspect of #749 and merge it onto this, though I suspect the merge would be unpleasant.

pozorvlak:
When I run the tests locally, it opens two browser tabs that don't open when I run the tests on `dev`. Investigating...

pozorvlak:
Turns out the problem existed at the point where the offending tests were introduced, and went away when I merged this branch into dev. So somewhere along the line it must have been fixed :-)

PR #782, opened 2015-07-23
Correct browser-resize issues with the members section of homepage and make members page match

maco:
This makes it so the member entries have the same white background as other cards and reflow as the page is resized. In the process, I removed some HTML elements that existed solely for display reasons and not for semantic reasons.

screen shot 2015-07-23 at 1 52 34 pm
screen shot 2015-07-23 at 1 52 47 pm
screen shot 2015-07-23 at 1 52 56 pm
screen shot 2015-07-23 at 1 53 12 pm
screen shot 2015-07-23 at 1 53 21 pm

### DIFF
--- a/app/helpers/gardens_helper.rb
+++ b/app/helpers/gardens_helper.rb
@@ -1,7 +1,7 @@
 module GardensHelper
 
   def display_garden_description(garden)
-  	if garden.description.nil?
+    if garden.description.nil?
   		"no description provided."
     else
       truncate(garden.description, length: 130, separator: ' ', omission: '... ') { link_to "Read more", garden_path(garden) }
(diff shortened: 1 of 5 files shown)


app/helpers/gardens_helper.rb
```diff
@@ -1,7 +1,7 @@
 module GardensHelper
 
   def display_garden_description(garden)
-  	if garden.description.nil?
+    if garden.description.nil?
   		"no description provided."
```
pozorvlak: I think you need to fix the indentation on this line too!

pozorvlak:
Looks good other than the whitespace issue I mentioned.

app/helpers/gardens_helper.rb
pozorvlak: I decided it was easier all round to fix it myself and merge this as-is :-)

PR #783, opened 2015-07-23
Replace the last few .count and .length calls with .size where possible

### LINKED ISSUE #566: Refactor: use .size for collections

pozorvlak:
Fixes #566.


cesy:
Looks good to me. For reference, comments of http://batsov.com/articles/2014/02/17/the-elements-of-style-in-ruby-number-13-length-vs-size-vs-count/ have details of the very rare cases where size isn't the best one to use.

PR #784, opened 2015-07-24
Check existence of secret token before using it.

pozorvlak:
People were forgetting to create config/application.yml, which meant that RAILS_SECRET_TOKEN wasn't being set, which meant that all tests involving notifications failed. Unfortunately, the resulting wall of error messages (https://gist.github.com/sha1sum/5debae6b700ff8fc0c76) did not make the root cause remotely clear, leading to much confusion and head-scratching all round.

This commit checks for the existence of RAILS_SECRET_TOKEN and fails with an informative error message if it's missing.

### DIFF
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -2,12 +2,19 @@ class Notifier < ActionMailer::Base
   include NotificationsHelper
   default from: "Growstuff <noreply@growstuff.org>"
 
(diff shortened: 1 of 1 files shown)


pozorvlak:
@maco, @cesy, does this look OK to you?

cesy:
Yes, looks good.

PR #785, opened 2015-07-25
Resolved #562 - Pagination of notifications.

### LINKED ISSUE #562: Paginate notifications
Notifications should be paginated so that only 30 per page appear.

sha1sum:
Notifications now paginate at 30 per page using Kaminari gem and add-on gem for Bootstrap-3-compatible pagination elements.

Resolves #562


pozorvlak:
Representative error message:

```
1) notifications/index whitespace-only subject shows (no subject)

     Failure/Error: render

     ActionView::Template::Error:

       undefined method `total_pages' for #<Array:0x0000000c01a138>

     # ./app/views/notifications/index.html.haml:4:in `_app_views_notifications_index_html_haml__463982717578562242_100589080'

     # ./spec/views/notifications/index.html.haml_spec.rb:59:in `block (3 levels) in <top (required)>'
```

cesy:
@pozorvlak was that error from Travis or from a local test? It looks like this is better now, but could do with a final check from you before merging.

sha1sum:
@cesy it was actually both local and on CI. I hurried through the modifications and only ran the feature specs. My apologies.

cesy:
Okay, so this is looking good. A few quick questions - what was the reason for using Kaminari instead of the existing pagination that we're using on crops index and other pages? And do you know how well it does for accessibility? Their page says they use HTML5 and the nav tag, which sounds like a good start.

sha1sum:
I was unaware that there were existing pagination controls on other pages; sorry about that. I use Kaminari in my own web application and am familiar with it, plus it has easily-extendable support for Bootstrap3 pagination controls built in, so it seems like the quickest way by far to add in the pagination. I'm not sure of the answer to your accessibility question, as I'm not accustomed to looking for that (most of my work is done on a business extranet application used by employees).

cesy:
Employees can still have disabilities - not being accessible is still a problem, and worth learning about. It looks like we're okay in this case, though.

So we just need the overall steer on whether we want to move to Kaminari, or if you can use the same pagination we're already doing in the crops index. @pozorvlak, @maco, what do you think?

pozorvlak:
I think we should be consistent, but I'm open to switching our pagination
over if Kaminari is better. I'd be happy to merge this now, and then do the
following:

 - look into the existing pagination code, evaluating it for accessibility
and ease-of-ripping-out
 - if the existing code is less accessible than Kaminari, raise a chore to
replace it with Kaminari
 - if the existing code is more accessible than Kaminari, raise a chore to
replace Kaminari with it.

cesy:
Looks like our current pagination doesn't do the HTML5 nav tag, which I think makes Kaminari better. Existing pagination is just the rails default, so no ripping out needed, just switching over. Let's do it.

PR #767, opened 2015-07-08
Added show member roles

jacemonje:
Redoing https://github.com/Growstuff/growstuff/pull/735 and to answer issue #557 

Will add the test; I'm still trying to figure out a better implementation.

PR #787, opened 2015-07-28
Show roles in member profile page

### LINKED ISSUE #557: Show roles (admin, crop wrangler) on profile page

twconquest:
Fixes #557. Takes @jacemonje's work from #767 and tests with feature test.


cesy:
Ideally we should also include negative tests that a non-crop wrangler's page doesn't say crop wrangler, and so on. Thank you for picking this up, and it's good to merge, but would be better still with a couple of extra tests if you have time.

pozorvlak:
Good point! @twconquest and I were planning another pairing session soon, we should be able to fix that.

pozorvlak:
@cesy we've added a negative test along the lines you suggested.

cesy:
Great, thanks!

PR #791, opened 2015-08-06
Cache thumbnails

cesy:
Helping performance issues on CropsController#index by caching thumbnails. Needs careful checking that cache expires correctly when a new planting is added.

### DIFF
--- a/app/views/crops/_thumbnail.html.haml
+++ b/app/views/crops/_thumbnail.html.haml
@@ -1,7 +1,7 @@
-- cache cache_key_for(Crop) do
-  .thumbnail
-    .crop-thumbnail
-      - if crop
+.thumbnail
+  .crop-thumbnail
+    - if crop
+      - cache cache_key_for(crop) do
         = link_to image_tag((crop.default_photo ?  crop.default_photo.thumbnail_url : 'placeholder_150.png'), :alt => crop.name, :class => 'img'), crop
         .cropinfo
           .cropname
--- a/app/views/crops/_thumbnail.html.haml
+++ b/app/views/crops/_thumbnail.html.haml
@@ -1,7 +1,7 @@
 .thumbnail
   .crop-thumbnail
     - if crop
-      - cache cache_key_for(crop) do
+      - cache cache_key_for(Crop, crop.id) do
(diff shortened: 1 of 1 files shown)


app/views/crops/_thumbnail.html.haml
```diff
@@ -1,13 +1,14 @@
...
-      .cropinfo
-        .cropname
-          = link_to crop.name, crop
-        - if crop.scientific_names.size > 0
-          .scientificname
-            = crop.scientific_names.first.scientific_name
-        .plantingcount
-          Planted
-          = pluralize(crop.plantings.size, "time")
+- cache cache_key_for(Crop) do
```
pozorvlak: This keys the fragment based on the Crop class - it's expired when any crop is updated or added, and I think in the meantime it will return the same fragment for *any* crop. We should base the cache key on the individual crop object, or at least pass in a value to the `identifier` keyword argument.

cesy: D'oh! That should have been "crop", not "Crop", then it would work.

pozorvlak:
We're still expiring the cache overzealously - `cache_key_for` generates a cache key that changes whenever *any* object of that class is updated. I think we probably just want something like `cache "crop #{crop.id} #{crop.updated_at}"`.

On the other hand, the list of crops doesn't change that often, and overzealous cache expiry is better than underzealous. So I'd be happy to merge it unless you'd rather make it more precise.

cesy:
Doesn't putting the identifier in reduce that, or am I still misunderstanding the code in the helper?

Given the complexity of making sure caches expire on the right types of events, I'd rather be overzealous than risk being underzealous, as you say, so probably worth merging as an improvement over the current situation if you agree, and we can come back to it if performance is still an issue on that page.

pozorvlak:
Adding the `identifier` fixes the problem that every crop would have the same thumbnail, but not the overzealous expiry. But as you say, let's merge this and come back to it if need be.

pozorvlak:
Here's a (lightly-edited) `rails console` transcript showing the problem:

```
 > helper.cache_key_for Crop, 5
 => "crops/5-439-20150717192649" 
 > c = Crop.find(7)
 > c.name = "dnomla"
 > c.save
 > helper.cache_key_for Crop, 5
 => "crops/5-439-20150819154255"
```

PR #790, opened 2015-08-06
Fixing relative caching of post summary on homepage, fixed #789

cesy:
Fixes #789 by making dates absolute rather than relative

PR #775, opened 2015-07-11
Speed up homepage by calculating "interesting" things more efficiently

pozorvlak:
The homepage is currently really slow to load - nearly 2s on my machine. A lot of this is due to the inefficient way in which we're calculating the lists of interesting crops and plantings - we load everything into memory from the DB, and then loop over them in Ruby, emitting more queries for each object to find out if it's interesting. With some moderately dark ActiveRecord magic we can do all of this calculation in only two database queries, getting page load time down to around 0.8s - still not great, but much better than it was.

Before
---------
```
$ ab -n 20 http://127.0.0.1:8080/
This is ApacheBench, Version 2.3 <$Revision: 1638069 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient).....done


Server Software:________
Server Hostname:        127.0.0.1
Server Port:            8080

Document Path:          /
Document Length:        23975 bytes

Concurrency Level:      1
Time taken for tests:   37.634 seconds
Complete requests:      20
Failed requests:        0
Total transferred:      495740 bytes 
HTML transferred:       479500 bytes 
Requests per second:    0.53 [#/sec] (mean) 
Time per request:       1881.699 [ms] (mean)
Time per request:       1881.699 [ms] (mean, across all concurrent requests)
Transfer rate:          12.86 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:  1832 1882  49.9   1872    1991
Waiting:     1832 1881  49.9   1872    1990
Total:       1832 1882  49.9   1872    1991
  
Percentage of the requests served within a certain time (ms)
  50%   1872
  66%   1883
  75%   1901
  80%   1955
  90%   1978
  95%   1991
  98%   1991 
  99%   1991
 100%   1991 (longest request)
```

After
-------
```
$ ab -n 20 http://127.0.0.1:8080/
This is ApacheBench, Version 2.3 <$Revision: 1638069 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient).....done


Server Software:________
Server Hostname:        127.0.0.1
Server Port:            8080

Document Path:          /
Document Length:        28299 bytes

[...]

PR #794, opened 2015-08-06
Fixing #790 properly - caching posts on homepage

cesy:
Reverts Growstuff/growstuff#793 that reverted #790. This one should now validly give the right fix, and since #775 is now sorted, Travis should be happy. So this now needs review from someone who is not me, particularly that we're happy with the UX of dates being absolute rather than relative.


pozorvlak:
Looks good to me!

PR #797, opened 2015-08-10
Style checkbox for 'remember me' 

CloCkWeRX:
Minor usability/mobile UI annoyance - you can't click the 'remember me' text to tick the checkbox.

This:
 * Applies bootstrap classes around the checkbox
 * Saves my big fat fingers from selecting text when they shouldn't

PR #800, opened 2015-08-11
Fix more checkbox labels in registration and email editing

CloCkWeRX:
Continues #797

The only other ones are in form partials, and don't really have the same kind of label - decided to leave them well enough alone.

### DIFF
--- a/app/views/devise/registrations/new.html.haml
+++ b/app/views/devise/registrations/new.html.haml
@@ -26,15 +26,17 @@
     .col-md-8= f.password_field :password_confirmation, :class => 'form-control'
 
   .form-group
-    .col-md-offset-2.col-md-8
-      = f.check_box :tos_agreement
-      I agree to the
-      = succeed "." do
-        = link_to 'Terms of Service', url_for(:action => 'tos', :controller => '/policy')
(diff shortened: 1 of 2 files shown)


app/views/devise/registrations/new.html.haml
```diff
@@ -26,15 +26,17 @@
...
-  .form-group
-    .col-md-offset-2.col-md-8
-      = f.check_box :newsletter, :checked => true
-      Subscribe to the #{ENV['GROWSTUFF_SITE_NAME']} newsletter
+    .col-md-offset-2.col-md-8.checkbox
+      %label
+        = f.check_box :tos_agreement
+        I agree to the
+        = succeed "." do
+          = link_to 'Terms of Service', url_for(:action => 'tos', :controller => '/policy')
```
cesy: Does this link still work okay or does this case become tricky?

pozorvlak: I just checked this locally - clicking "I agree to the" checks the checkbox, and clicking "Terms of Service" follows the link. Firefox 38 and Chromium 43 on Linux. Is that enough, or do we need further checks?

cesy: That's enough :)

PR #801, opened 2015-08-12
Fix CVE-2015-2963,  CVE-2015-3448, CVE-2015-1820 & CVE-2015-1840

CloCkWeRX:
Name: paperclip
Version: 4.2.1
Advisory: CVE-2015-2963
Criticality: Medium
URL: https://robots.thoughtbot.com/paperclip-security-release
Title: Paperclip Gem for Ruby vulnerable to content type spoofing
Solution: upgrade to >= 4.2.2

I jumped us to 4.3.0, which fingers crossed wont be an issue. If it is, we can back this out and pin to ~> 4.2.2

Version: 1.7.2
Advisory: CVE-2015-3448
Criticality: Unknown
URL: http://www.osvdb.org/show/osvdb/117461
Title: Rest-Client Gem for Ruby logs password information in plaintext
Solution: upgrade to >= 1.7.3

Name: rest-client
Version: 1.7.2
Advisory: CVE-2015-1820
Criticality: Unknown
URL: rest-client/rest-client#369
Title: rubygem-rest-client: session fixation vulnerability via Set-Cookie headers in 30x redirection responses
Solution: upgrade to >= 1.8.0

Name: jquery-rails
Version: 3.1.2
Advisory: CVE-2015-1840
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/ruby-security-ann/XIZPbobuwaY
Title: CSRF Vulnerability in jquery-ujs and jquery-rails
Solution: upgrade to >= 4.0.4, ~> 3.1.3


CloCkWeRX:
That seems flakey rather than legit regression - https://travis-ci.org/CloCkWeRX/growstuff/builds/75220303

CloCkWeRX:
Pushed a few more security related fixes to redo the travis build, amended PR

cesy:
The changes just appear to be in Gemfile.lock, not in Gemfile? Depending on the risk of these, we might want to force it via Gemfile. It would also be worth trying bundle update to get latest versions of everything and see what breaks, rather than doing it piecemeal.

CloCkWeRX:
I tend to use Gemfile version pinning for API compatibility reasons. I
suspect if we start doing it for security,  we will be reinventing some of
the purpose of Gemfile.lock itself, and be doing a bunch of manual stuff
for minimal value.

I agree that later on doing major version upgrades, starting from a clean
slate re deps will be easier, but does increase the risk of regression.

The only other security fix pending is a bump to ruby 2.1.6; and maybe we
can add a badge or similar to auto detect security issues (been finding
these via bunde-audit)

cesy:
Okay, that makes sense. @pozorvlak, @maco, can I get a quick second opinion on this one before merging given the importance and it's not so much my area of expertise? Thanks.

maco:
Gemfile versus Gemfile.lock isn't my area either

pozorvlak:
Yeah, I think @CloCkWeRX's argument for keeping the change to Gemfile.lock makes sense.

Issue #734, opened 2015-03-31
We should add the Bootstrap Accessibility plugin


maco:
https://github.com/paypal/bootstrap-accessibility-plugin

This adds accessibility hints to a big pile of Bootstrap widgets

cesy:
Yes, that sounds good.

pozorvlak:
+1

PR #803, opened 2015-08-12
Bootstrap accessibility

cesy:
Attempting issue #734. Needs checking to see that compass etc. don't barf on different systems.


pozorvlak:
Is it enough to just drop the files under app/assets? I tried running a server with this branch checked out, but didn't know what I was looking for.

cesy:
I'm not sure - the simplest one to check would be "Add role of Alert to Alert, Warning, and Success Bootstrap Messages." but I couldn't get it running on Nitrous to check.

cesy:
@maco do you have time to review this one, please?

Issue #658, opened 2015-01-19
Internationalise strings throughout the website (i18n)


Skud:
Taken from [this discussion thread](http://talk.growstuff.org/t/i18n-implementation-details/88/12):

Here's a list of priorities of strings to be i18n'd by turning "foo" into t("foo") and putting the relevant string in config/locales/en.yml
-    top navbar items ("Crops", "Community", etc)
-   bottom navbar items ("Contact", "Terms of Service", etc)
-    page titles throughout the site, eg. "Edit planting"
-    action buttons, eg. "Save", "Plant this", etc.
-    field names on forms throughout the site, eg. "Planted at:"
-    items shown in the "thumbnail" and "index card" views for plantings, crops, members, etc
-    Note that as we go, we should definitely think about creating helpers or view partials for anything that gets reused frequently, to avoid repetition and make sure there's only one string to translate (see the signup_signin partial as an example).

So the task here is to slog through all of this. I don't expect anyone will do it all at one hit, but feel free to do a handful and issue a PR for any small amount you like.

I've marked this as "curated: beginner" as it seems like a beginner-friendly thing to do.  A step by step howto would probably also help with that.

roryaronson:
We're just looking into translating our i18n strings for OpenFarm. You might find the discussion/links helpful: https://github.com/openfarmcc/OpenFarm/issues/429

Have you researched any ways/services to translate your strings? Perhaps we can work together considering we have very similar ones?

Skud:
I suspect our strings are sufficiently different that it might not 
actually help that much, just looking through some of the text on your 
site.  We have a few short strings in common ("search", "about", 
"privacy") but not much else.  I did a rough review of Growstuff's 
app/views and found around 1500 strings in our app so there's obviously 
a lot there.

Wrt our plans: we'll be working with our community for translations, 
taking our lead from a few other open source projects we're familiar 
with.  We've been looking around for a good web UI for l10n but haven't 
found much that impresses us as yet.  For now it's just locales/*.yml.

A.


On 21/01/2015 6:57 am, Rory Aronson wrote:
>
> We're just looking into translating our i18n strings for OpenFarm. You 
> might find the discussion/links helpful: openfarmcc/OpenFarm#429 
> <https://github.com/openfarmcc/OpenFarm/issues/429>
>
> Have you researched any ways/services to translate your strings? 
> Perhaps we can work together considering we have very similar ones?
>
> —
> Reply to this email directly or view it on GitHub 
[...]

PR #804, opened 2015-08-12
Issue #658 i18n

cesy:
An example of internationalising a string, to make it clearer for future beginners looking at #658.


pozorvlak:
Why the dot in front of `skip`? Merging since this works, but I'm curious...

cesy:
Following the same structure as the others that already work, but I'm not sure why - I'd guess it's related to the subtree structure in en.yml.

jestallin:
The . (dot) is the lazy lookup according to the rails guide. 

http://guides.rubyonrails.org/i18n.html

PR #805, opened 2015-08-12
WIP: Fix #632 to use consistent API

### LINKED ISSUE #632: Newsletter subscriptions broken
Something seems broken on production, where new signups aren't getting added to Mailchimp.

http://growstuff.org/admin/newsletter shows me ~800 newsletter subscribers, but Mailchimp only shows me 500-ish total, with the numbers flattening off around March last year (eep!)

Workaround: before sending a newsletter, use the list at the link above and import it to Mailchimp manually.

To actually fix: need to dig around with the mailchimp API settings, I think :-/  Will probably need quite a bit of manual testing. Ugh.

cesy:
Looking at issue #632 and the comment, either we need to make lots of tricky changes to use the new MailChimp API, or we can lock to the old gibbon version to get it back working for now. I suggest we do this so the newsletter will start working again, and then open a fresh issue to upgrade to latest MailChimp API and gibbon gem together.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,9 @@ gem 'unicorn'                      # http server
 gem 'pg'
 gem 'figaro'                       # for handling config via ENV variables
 gem 'cancancan', '~> 1.9'          # for checking member privileges
-gem 'gibbon'                       # for Mailchimp newsletter subscriptions
+gem 'gibbon', '=0.4.6'             # for Mailchimp newsletter subscriptions
+# Gibbon >0.5.0 targets MailChimp API 2.0, which is substantially different from API 1.3.
+# Please use Gibbon 0.4.6 if you need to use API 1.3.
 gem 'csv_shaper'                   # CSV export
 gem 'ruby-units'                   # for unit conversion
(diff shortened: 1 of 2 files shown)


cesy:
Okay, that barfed horribly on Travis. Back to the drawing board for more investigation of when and how it broke.

pozorvlak:
I think this broke on Travis because our code was never written to use Gibbon v0.4.6 - we started using it at version 1.0.0, and the `Gibbon::API` module was added sometime between the two.

This further suggests that our problem won't be solved by downgrading to an older API :-(

cesy:
Okay, closing this for now, then, and more ideas needed.

pozorvlak:
Currently pairing on this with @twconquest, and I think we've found the commit where the bug was introduced: https://github.com/Growstuff/growstuff/commit/b86014b04f68c361665f88f3ad7ea9737029d83d. We'll fix this, which should be straightforward, and then look at upgrading to the v3.0 Mailchimp API since the v2.0 one is now deprecated.

PR #806, opened 2015-08-13
Upgrade to ruby 2.1.6 for CVE-2015-1855

CloCkWeRX:
Full detail
https://www.ruby-lang.org/en/news/2015/04/13/ruby-2-1-6-released/

and changelog

http://svn.ruby-lang.org/repos/ruby/tags/v2_1_6/ChangeLog

I don't know how you handle this for production deploys, or if that would entail manual work, but travis seems happy so far.


cesy:
It should all work smoothly and automatically, but we'll watch it carefully as we have a lot of these kinds of complex changes in this release.

Issue #645, opened 2015-01-16
Signup/signin with Twitter, Facebook, etc


Skud:
It would be good if people could signup with Twitter etc -- might make the onboarding process smoother.

jacemonje:
Hi, will do this :)

CloCkWeRX:
How are you traveling with this @jacemonje? Happy to go add various providers if you are a bit stuck

CloCkWeRX:
One of the issues with flickr based signup is the lack of email in the profile API.

https://www.flickr.com/services/api/flickr.people.getInfo.htm

It means that a signup with flickr needs to do SELECT * FROM members WHERE sha1(email) = ? ; but by default I don't think we'd have the postgres crypto (pgcrypto) things enabled.

cesy:
Yeah, I don't think we need sign in with Flickr, just Facebook/Twitter/Google. Flickr is mainly useful to add to an existing account for photos, until we get our own photo hosting.

CloCkWeRX:
Closing.
Flickr isn't what it was, Twitter is declining, Facebook made me annoyed with trying to do an "App review" process on us.

Issue #556, opened 2014-12-12
Connect to Facebook


Skud:
Members should be able to connect their Growstuff account to their Facebook accounts.  This should use OAuth.  The resulting linked accounts should be displayed on the Growstuff member's profile.

CloCkWeRX:
Dupe?

cesy:
This is linking, not necessarily sign-in via Facebook. It allows someone with a Growstuff password to have facebook linked as well, which is a step beyond #645 though both could be done at once.

CloCkWeRX:
This one is now plausible with the above PR, but probably needs a bit more discussion.

We can do the 'add facebook credential' step fairly easily - the existing authentications_controller seems to be 99% there; but we probably want to think about what kind of information we want to read (user location, photos, etc), and if there is any we want to write.

cesy:
I think probably we don't want to read location - better to give members the choice of how granular to specify it for Growstuff separately. Reading photos is a nice feature for the long-term, as it will make adding photos to gardens much easier, but not necessary for MVP. Part of the aim of this eventually is to be able to write - adding optional "Share to Facebook" and "Share to Twitter" buttons whenever you plant or harvest something, and so on.

maco:
Technically, we could have had share buttons ages ago. They can be done
without the API.

cesy:
Is it worth us spinning off a separate issue for them and locking down the
design so we can get them added, then?

CloCkWeRX:
I think it might be. Interestingly, now that #733 is "sorta working", we can probably implement this with plain links by simply rendering an appropriate og:image attribute whenever we have a more accurate picture, and populating an og:title (should already work if there's content_for :title)

CloCkWeRX:
After #811 is bedded down, I'll pick this up and look at UI for 'connect facebook account' - should be fairly straight forward given the existing implementations.

CloCkWeRX:
This should work now, as part of #811 - from the 'apps' member page, you can connect. There's something odd on my local around the routes which is absent in staging, so it'll need testing there + the credential setup... but yeah; should be closable shortly

PR #811, opened 2015-08-21
Add facebook signup

CloCkWeRX:
Partially improves #645 

A number of things remain before its production ready, but I thought it was worth putting up for early review.
 * Decisions about how the 'sign in' and 'sign up' with facebook UI should appear
 * Detailed rspec coverage of the omniauth callback controller (its somewhat hard to do selenium style tests of external oauth providers well)  (solved)
 * Login name generation vs conflicts, what's the best way to avoid? (Generate random and allow login_name to change?) (solved)
 * Decisions around failed login cases - is the error behaviour sufficient?
 * How do we set up the staging/production credentials properly, and who will do that
 * Do we want extra information, like the user location to automate some signup?
 * Do we want to document future work, like "user image from facebook on new signup" in more detail?  (solved)
 * Do we want to automatically accept the ToS? I've assumed yes to get the account created, but is this a critical part of the workflow?  (solved)
 * Do we want to extend the 'integrations' page to allow facebook oauth credentials to be added?' (#556)

signin2
signin


cesy:
I think we want to offer them the auto-generated login_name and let them edit it, highlighting at that point if it's already taken. I think we also want to get them to sign the ToS explicitly.

The screenshot of the sign in link looks fine for UI at the moment - we can discuss separately if we want a button next to sign in, but I prefer this.

I think the default error behaviour is likely to be sufficient, but we should test it.

Credentials for staging will need to be set up on heroku by someone with access (most committers) before we merge this. Production will need to be set up before we deploy there. The way you've designed it with an environment variable looks good, though @pozorvlak should also review.

Extending integrations page is definitely a good idea, yes.

User image from facebook is also a good idea, let's log an issue for that.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    @member_created = false
+
+    auth = request.env['omniauth.auth']
+
+    authentication = Authentication.where(provider: auth.provider, uid: auth.uid).first
+    if authentication && authentication.member.id.present?
+      member = authentication.member
+    end
+
+    member = Member.where(email: auth.info.email).first_or_create do |m|
```
cesy: What happens in the case where this overwrites line 34?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+      m.email = auth.info.email
+      m.password = Devise.friendly_token[0,20]
+      m.tos_agreement = true
```
cesy: We should split this out and have a workflow to ask for it specifically, where we can also get them to review and edit the login name to prevent collisions.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+      # TODO This has a reasonable chance of collision
+      m.login_name = auth.info.nickname || auth.info.email.split("@").first.gsub(/[^A-Za-z]+/, '_').underscore
+
+      m.skip_confirmation!
```
cesy: Just noting that we should think about this, if we're happy to skip email confirmation since facebook will have already done it, or if we still want to do it. I think I probably agree with the code here that we should skip it. It might be worth adding a comment to remind us of this reasoning in future.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def create
+    auth = request.env['omniauth.auth']
+    @authentication = nil
+    if auth
+
+      name = ''
+      case auth['provider']
+      when 'twitter'
```
cesy: I like that it looks like an easy future enhancement to allow sign in with twitter as well.

app/controllers/omniauth_callbacks_controller.rb
CloCkWeRX: Doh, meant to be ||=

app/controllers/omniauth_callbacks_controller.rb
CloCkWeRX: It felt really odd with the confirmation in place,  as oauth signup would work but then fail you in the login flow until you confirmed the account.

I agree that the ToS step needs to happen,  and that there is value in confirming the email,  but am less sure about how devise handles it in this step.

app/controllers/omniauth_callbacks_controller.rb
CloCkWeRX: Yeah,  wanted to keep it simple for the first step,  but all of the existing ouath sign ins are a trivial refactor afterwards to hook up to devise :)

CloCkWeRX:
There's now a finish signup flow, and the ToS are no longer accepted by default.

signup

Not sure if I have to do anything special around subscribing a user to the newsletter.

So by dropping the "You must agree to the ToS" validation from the model this possibly upsets the ordinary registration process - I'll have to have a play.

Right now the user isn't compelled to agree, so in theory could signup/login; close the tab, login again and be let in I think, I haven't tested that too much.

To prevent that http://sourcey.com/rails-4-omniauth-using-devise-with-twitter-facebook-and-linkedin/ sugggests adding a before_filter to route a user to the signup completion on just about every action - was a bit daunting.

I think the only remaining piece is dealing with login name conflicts - not sure how best to generate a sufficiently unique id that is human friendly. The complete signup flow does let a user change/edit their login, but if a collision happens during oauth account creation you are out of luck right now.

cesy:
@twconquest and @pozorvlak have just been looking at the mailing list so can comment on that part.

Can we add some logic so that if a collision happens it adds a 1 on the end or something like that?

It's probably worth adding some feature tests for all of this - stubbing facebook API will be a faff but we'll want it in the long run.

CloCkWeRX:
Re the testing, I'd be really keen to split that off to a different issue. I'm assuming that controller style tests are still on the out - if not https://github.com/intridea/omniauth/wiki/Integration-Testing however gives us a bit of a hook there.

If they are, writing a fake oauth server so that the flow works is a fair bit of effort; alternatively writing and maintaining page objects that know about the login UI of each provider tends to lead to some pretty brittle tests (at work, simply dealing with google's UI has seen ~3-4 design revisions in 12 months for google auth flow - repeating for facebook, twitter, flickr, and whatever else we end up adding = :S).

Happy to refactor and push a lot more of this logic into service objects/low level rspec coverage of that, but that's not a pattern really present in the application yet.

CloCkWeRX:
> Can we add some logic so that if a collision happens it adds a 1 on the end or something like that?

I thought about it, but went with 'completely random token' to all but eliminate collisions if we have one in 9564866f6d8bd5ebcf23d12ea61a63d1a2bedeb8.

From a code point of view if we simply added 1 to the end and retried, what sort of limits do we set there? How many attempts is enough to avoid a collision, and what do we do if they are exceeded - prevent the user from signing up?

With the token if there is a collision its a pretty unwieldy login name; but the user is prompted to edit it in the very next step and is more than likely to pick something they prefer.

cesy:
Okay, let's get a second opinion from @pozorvlak or @maco, then.

CloCkWeRX:
Really odd: I started doing the 'connect to Facebook' part of this, but on my local the routes for Flickr, Twitter etc are broken. Staging, they work. Anyone familiar with why this would behave the way it does?

pozorvlak:
Have you generated the necessary API tokens for yourself and added them to `config/application.yml`?

CloCkWeRX:
Yeah, definitely did - got auth working through /member/auth a bit for it, and ran into other problems (no email is provided, so really tricky to set up an account or find an existing member without a bunch more work).

CloCkWeRX:
:/ keep getting distracted from coming up with at least some test coverage of this.

CloCkWeRX:
Basic test coverage is there, but the amount of conditionals under the hood vs the level of testing doesn't feel super right at the moment. Annoyingly, its hard to set up other contexts with the fake signin; so not really too easy to check 'login invalid', 'junk data is handled', etc. Might need to shift the logic out of controller-land and writing lower level tests.

CloCkWeRX:
Tentatively, I think it's done - next round of review?

CloCkWeRX:
Sigh; the feature test is flakey as heck for signup; even through selenium. It's not immediately clear why.

cesy:
Now that we've got the urgent upgrades out the way, it would be great to get this one going again, if you have time?

homu-gs:
:umbrella: The latest upstream changes (presumably #1027) made this pull request unmergeable. Please resolve the merge conflicts.

cesy:
If you're up for it, how about commenting out the flaky test and breaking that out into a separate issue, so we can get this live?

CloCkWeRX:
This is pretty much ready - the TODOs code climate is complaining about are placeholders for future integrations; and the security hole has a check before it to prevent abuse.

cesy:
Okay, this looks good to me. I'll give it 24 hours in case @maco wants a look, but I'm looking forward to getting this on Staging and playing with it. Thank you!

homu-gs:
:umbrella: The latest upstream changes (presumably #1033) made this pull request unmergeable. Please resolve the merge conflicts.

PR #812, opened 2015-08-21
Fix #632 Make mailchimp newsletter subscription work again

### LINKED ISSUE #632: Newsletter subscriptions broken

twconquest:
Fix #632 . Change app/models/member.rb line 241 to point to correct mailchimp list.

Also, make changes to upgrade to [latest Gibbon v2.x](https://github.com/amro/gibbon).

Currently in process of testing. Not ready for merge just yet.

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -239,7 +239,7 @@ class Member < ActiveRecord::Base
     return true if (Rails.env.test? && !testing)
     gb = Gibbon::API.new
     res = gb.lists.subscribe({
-      :id => Gibbon::API.api_key,
+      :id => config.newsletter_list_id,
       :email => { :email => email },
       :merge_vars => { :login_name => login_name },
       :double_optin => false # they already confirmed their email with us
@@ -250,7 +250,7 @@ class Member < ActiveRecord::Base
     return true if (Rails.env.test? && !testing)
     gb = Gibbon::API.new
     res = gb.lists.unsubscribe({
-      :id => ENV['GROWSTUFF_MAILCHIMP_NEWSLETTER_ID'],
+      :id => config.newsletter_list_id,
       :email => { :email => email }
     })
   end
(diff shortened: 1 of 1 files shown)


twconquest:
We ( @pozorvlak and myself ) think the code is correct, but we have not managed to do any manual testing of it yet. As far as we know there are no automated tests setup for gibbon. Will revisit this next pairing session. Any suggestions for tests / best way to reliably automate tests for gibbon/mailchimp would be welcome.

cesy:
The closest I could find to help was http://iswwwup.com/t/b98579e9ceb3/stubs-mailchimp-api-call-for-functional-testing.html, but there's a part of me that thinks automated tests are less critical for gibbon, as it's a low regression risk, and the critical parts are all in the config that won't get caught with automated tests anyway. So for the amount of effort, the chances of an automated test catching a bug are pretty low.

pozorvlak:
We couldn't get this working with Gibbon 2.0, so we rolled back to v1.2 and applied the configuration fix. I've done the following manual tests locally:

 - create a new user, and check in the Mailchimp admin panel that they're subscribed to the mailing list;
 - click "unsubscribe" and "save" in an existing user's profile page, and check that they disappear in Mailchimp;
 - click "subscribe" and "save", and check that they reappear in Mailchimp.

cesy:
Do we need to do anything special to set that config value in Production?

pozorvlak:
It should already be set - run `heroku config --app=growstuff-prod` to find
out. I'm away from a computer for the next few days, so can't check myself.

cesy:
Okay, looks good, happy to merge once conflicts are fixed.

Issue #507, opened 2014-12-12
if i reply to a notification directly, it's not marked as read


Skud:
The flow is like this:

- i get an email notification saying someone sent me a PM
- I click the "reply" link which takes me to a reply screen
- I send my reply
- I wind up back in my Inbox

What happens: the message I replied to is not marked as read (i.e. is still bold, shows up in count in navbar).

What should happen: the message should be marked as read.

PR #813, opened 2015-08-22
If i reply to a notification directly, it's not marked as read. #507

krio:
Fixes: https://github.com/Growstuff/growstuff/issues/507


yoongkang:
Great! But probably needs a test.

PR #814, opened 2015-08-24
Add correct prefix to fix facebook share icon

CloCkWeRX:
#733 Add the correct RDFa prefix

I think this will work OK, but don't have a way to test my localhost properly.

http://www.w3.org/2012/pyRdfa/extract tells me it's valid RDFa, producing:

```
@prefix ns1: <http://ogp.me/ns#image:> .
@prefix ns2: <http://www.w3.org/1999/xhtml/vocab#> .
@prefix og: <http://ogp.me/ns#> .

<> og:image "http://localhost:3000/assets/growstuff-apple-touch-icon-precomposed-8e3c5998fc73bde728cfc3a2ea8fb69f.png"@en;
    ns1:secure_url "images/growstuff-apple-touch-icon-precomposed.png"@en .

[] ns2:role ns2:navigation .
```

Or in other words, <> (this page) has an og:image of "http://localhost:3000/assets/growstuff-apple-touch-icon-precomposed-8e3c5998fc73bde728cfc3a2ea8fb69f.png"

If this doesn't work, probably need to make it <http://localhost:3000/assets/growstuff-apple-touch-icon-precomposed-8e3c5998fc73bde728cfc3a2ea8fb69f.png>

### DIFF
--- a/app/views/layouts/_meta.html.haml
+++ b/app/views/layouts/_meta.html.haml
@@ -1,8 +1,7 @@
 %head
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-  <meta property="og:image" content="images/growstuff-apple-touch-icon-precomposed.png"/>
-  <meta property="og:image:secure_url" content="images/growstuff-apple-touch-icon-precomposed.png"/>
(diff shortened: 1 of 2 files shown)


CloCkWeRX:
Diving a bit deeper, we may later on wish to set the opengraph type better ('plant', 'garden', 'seed')

https://developers.facebook.com/docs/sharing/opengraph/custom#objects

PR #815, opened 2015-08-24
GS-658 - internationalize menu headers

jestallin:


CloCkWeRX:
#658

cesy:
For the future, adding "Fix #658" to a commit comment will make it hook in automatically to the issue. And having the hash there in your pull req title makes it link automatically as well. This will need someone other than me to review and merge as I paired on it.

jestallin:
How should I identify the PR as ready for review? I skimmed the wiki, but did not see an immediate answer.

CloCkWeRX:
The code seems fine/straight forward enough - the only thing I'd comment on is a nitpick about style

```
common.Sign_up
```

If you look at the convention in the rest of the .yml file (and most yml files); it's lowercase_and_underscores; rather than camelCase or a hybrid.

It probably wont cause problems, but a lot of rails relies on assumptions/conventions to work as expected (http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html for interest)

cesy:
No special process to identify a PR as ready for review - every PR is always assumed to be ready for review unless you state otherwise.

maco:
Although being able to mark it so we can see from the list that a PR has
gotten feedback and that feedback has not yet had action taken...given how
many we have...that'd be handy.

jestallin:
@CloCkWeRX I chose that style for the key because it helps identify the case in the string. I chose this style after reading http://stackoverflow.com/questions/12169785/what-is-the-best-practice-to-capitalize-words-in-rails 

The specific paragraph that stuck out was: 
> The basic rule in L10N is to treat user-facing strings as opaque blobs of data that you pull out of a database of some sort and show to the user without manipulation in between. Yes, this makes your code more complicated but correctness is sort of important.

I envisioned tokens would exist for both cases.

```
sign_up: 'sign up'
Sign_up: 'Sign up'
```

It seemed plausible that a language's character could be different depending on case.

jestallin:
I went back and cleaned up the case per @CloCkWeRX's suggestion

cesy:
@maco, don't we tend to achieve that with the "needs:morework" label a lot of the time? Would be good to document, though, and let people know they can take that label off when they update.

maco:
Oh right. Yes, that sounds like a good idea

cesy:
Is it worth reopening this and merging it so we have your name in contributors and this bit of progress while we review the other proof of concept?

jestallin:
If we adopt the rake automation then it would not be worth it. The common
paths would be overridden with by the rails lazy locale key lookup.

I added my name to the contributors on the poc PR :)

Issue #509, opened 2014-12-12
Use non-gravatar photo for profile pic


Skud:
As a member, I want to be able to use a photo (not Gravatar) for my profile pic.  

Maybe we should default to gravatar, but then allow people to upload a pic, which overrides the gravatar if it's available?

cesy:
Ideally this should also offer the option of pulling through your profile photo from facebook or twitter, either as a one-off where Growstuff copies the current one or as a link where it will change when you change on the remote site, whichever is easier.

CloCkWeRX:
With the Facebook / most outh signins, this is simply a url passed around.

The most basic implementation would be a URL of off the member model and a tiny bit of ui to update that / default it smartly on outh login.

CloCkWeRX:
https://github.com/Growstuff/growstuff/pull/4146 removes twitter

PR #817, opened 2015-08-24
Add member preferred image

CloCkWeRX:
Partially addresses #509 - since most of the oauth providers give you an avatar URI, it'll be trivial to refactor #811 a bit further to set it to your $whatever login pic.

gravatar

and we simulate the oauth login setting this:
```
irb(main):003:0* m = Member.first; m.preferred_avatar_uri = 'http://media.catmoji.com/post/ujg/cat-in-hat.jpg'; m.save!
  Member Load (47.1ms)  SELECT  "members".* FROM "members"   ORDER BY lower(login_name) asc LIMIT 1
  Member Load (46.4ms)  SELECT  "members".* FROM "members"   ORDER BY lower(login_name) asc LIMIT 1
   (41.5ms)  BEGIN
  Member Exists (47.1ms)  SELECT  1 AS one FROM "members"  WHERE (LOWER("members"."login_name") = LOWER('daniel_oconnor') AND "members"."id" != 8) LIMIT 1
  SQL (43.5ms)  UPDATE "members" SET "preferred_avatar_uri" = $1, "updated_at" = $2 WHERE "members"."id" = 8  [["preferred_avatar_uri", "http://media.catmoji.com/post/ujg/cat-in-hat.jpg"], ["updated_at", "2015-08-25 00:34:20.680746"]]
   (46.4ms)  COMMIT
```

and a simple refresh; hey presto

custom_image

... a pleased cat.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -37,5 +37,18 @@ module ApplicationHelper
     content_tag :div, asterisk + ' '.html_safe + text, class: ['margin-bottom']
   end
 
+  #
+  # Returns an image uri for a given member.
+  #
+  # Falls back to Gravatar
+  #
+  def avatar_uri(member, size = 150)
+    return member.preferred_avatar_uri if member.preferred_avatar_uri.present?
+
+    Gravatar.new(member.email).image_url({
+      :size => size,
+      :default => :identicon
+    })
+  end
 end
 
--- a/app/views/members/_avatar.html.haml
+++ b/app/views/members/_avatar.html.haml
@@ -1,9 +1,5 @@
 = link_to                                      |
-  image_tag(                                   |
-    Gravatar.new(member.email).image_url(      |
-      options = {                              |
-        :size => defined?(size) ?  size : 150, |
-        :default => :identicon }),             |
+  image_tag(avatar_uri(member, 150),           |
     :alt => '',                                |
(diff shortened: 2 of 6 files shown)


cesy:
How does this work with the current caching on the member card?

CloCkWeRX:
Ah, which caching specifically?

 * app/views/home/_members.html.haml ?
 * app/views/members/_thumbnail.html.haml ?

Right now the only place in the app that expires cache explicitly is for a bit of the homepage stats after a member is created: 
```
  after_action :expire_cache_fragments, :only => :create
```

All of the other member caching relies on 
```
cache member
```
... which will write out a unique id based on object.id + object.updated_at (http://edgeguides.rubyonrails.org/caching_with_rails.html) - any saves will invalidate it.

cesy:
I was thinking of the thumbnail but we should check both. But that's more for the bit where you're pulling it from another site - I was thinking of that issue, whereas this one is just setting it manually, so it only updates when you save and everything is okay for this issue, just need to bear it in mind when we get to the auto-updating of avatar_uri part.

cesy:
And obviously this doesn't give you any interface to set it manually, it's just there in the code, so no need to worry yet about users putting malicious URLs in there - if we stick to Facebook, Twitter and other reputable providers, we're fine.

PR #820, opened 2015-08-27
Fix #819 by removing the popover behaviour for a standard avatar

### LINKED ISSUE #819: Minor usability regression on popover of member image (staging)
#570 / #768 related

http://staging.growstuff.org/members

 * Hover over a profile image

Results:
 * It "flickers" between mouseover and mouseexit events unless you get it exactly right
 * Popover styling is limited by the new layout to the wrong with
 
regression_popover

CloCkWeRX:
Fix #819 by putting the stats onto the card.

This fixes the usability issue, but might cause a design problem. Worth discussing how important these stats are to be present, ditto the 'member since' wording (this is covered better on the full member profile page)

no_gardens


cesy:
It might be worth taking the last three lines and making them one line with commas between? That will get it closer to the old card shape.

CloCkWeRX:
Yeah, I did think about that but you then end up with the layout being off across a lot of people:

```
CloCkWeRX                                             AnotherPerson
1 gardens, 234566 plantings, 0 seeds                  1 gardens, 1 plantings, 0 seeds

Longnameuser                                          YetAnotherPerson
10 gardens, 15 plantings, 234 seeds                   1345436 gardens, 20 plantings, 0 seeds
```

I'll give it a go after I fix the specs

CloCkWeRX:
comma

cesy:
Given it's a design change, I'd like a second opinion before merging, but this looks ready to me.

pozorvlak:
Looks good to me!

PR #821, opened 2015-08-27
Fix #818 by removing the .thumbnail around the button.

### LINKED ISSUE #818: Minor style regression (staging) for empty profile
#570 / #768 related I think

Reproduce steps:
http://staging.growstuff.org/
 * New account
 * Login
 * View your profile

Result:
The 'add photo of your garden' looks a bit funny with the card background / doesn't really explain that you are adding a photo to the selected garden well (heading size implies its a whole new section)

empty_profile_style_regression

CloCkWeRX:
Also rearrange some of the headings (sizing) and controls (add new garden) to be more clearly not a tab.

fix_empty_garden

Also the same type of fixes and some basic layout tweaks on the show garden UI:
show_garden

### DIFF
--- a/app/views/members/_gardens.html.haml
+++ b/app/views/members/_gardens.html.haml
@@ -7,31 +7,42 @@
         - first_garden = false
         = link_to g.name, "#garden#{g.id}", 'data-toggle' => 'tab'
     - if current_member == member
-      %li= link_to 'New Garden', new_garden_path
-  .tab-content
+      %li.navbar-right 
+        = link_to new_garden_path, class: 'btn' do
+          Add New Garden
+  .tab-content{style: "padding-top: 1em"}
     - first_garden = true
     - member.gardens.each do |g|
 
       %div{:class => ['tab-pane', first_garden ? 'active' : ''], :id => "garden#{g.id}"}
         - first_garden = false
 
-        %div
+        .container
           :growstuff_markdown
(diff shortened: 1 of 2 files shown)


app/views/members/_gardens.html.haml
```diff
@@ -7,31 +7,42 @@
         - first_garden = false
         = link_to g.name, "#garden#{g.id}", 'data-toggle' => 'tab'
     - if current_member == member
-      %li= link_to 'New Garden', new_garden_path
-  .tab-content
+      %li.navbar-right 
+        = link_to new_garden_path, class: 'btn' do
+          Add New Garden
```
cesy: Does this need an "end", or not in this type?

pozorvlak: [Doesn't look like it](http://stackoverflow.com/questions/21415570/pass-ruby-block-to-partial-in-haml) - in general, Haml uses indentation rather than explicit `end` statements.

CloCkWeRX: http://haml.info/docs/yardoc/file.REFERENCE.html#ruby_blocks for interest

cesy:
These all look like good improvements to me, but a design change should have a second opinion if possible - anyone else happy with this?

pozorvlak:
@maco got in before me, but I too like this change :-)

PR #823, opened 2015-08-30
GS-658 - internationalize header links and some page titles

jestallin:


jestallin:
#658

cesy:
Great. Our next challenge will be to start on an example translation, with one of our volunteers - do you want to post on talk.growstuff.org and see if anyone is keen?

Issue #504, opened 2014-12-12
Quick links to members in markdown


Skud:
As we currently have [tomato](crop), we should also have [Skud](member).  It might be nice to also support @Skud notation.

Br3nda:
this was done in #824

PR #824, opened 2015-09-05
Quick links to members in markdown 

sksavant:
Feature #504

### DIFF
--- a/lib/haml/filters/growstuff_markdown.rb
+++ b/lib/haml/filters/growstuff_markdown.rb
@@ -4,7 +4,7 @@ module Haml::Filters
   module GrowstuffMarkdown
     CROP_REGEX = /\[([^\[\]]+?)\]\(crop\)/
     MEMBER_REGEX = /\[([^\[\]]+?)\]\(member\)/
-    MEMBER_AT_REGEX = /\@([^\[ \n]+)/
+    MEMBER_AT_REGEX = /(\@[^\[ \n]+)/
     include Haml::Filters::Base
 
     def render(text)
@@ -39,10 +39,10 @@ module Haml::Filters
       expanded = expanded.gsub(MEMBER_AT_REGEX) do |m|
         member_str = $1
         # find crop case-insensitively
-        member = Member.where('lower(login_name) = ?', member_str.downcase).first
+        member = Member.where('lower(login_name) = ?', member_str[1..-1].downcase).first
         if member
           url = Rails.application.routes.url_helpers.member_url(member, :only_path => true)
-          "[#{"@"+member_str}](#{url})"
+          "[#{member_str}](#{url})"
         else
           member_str
         end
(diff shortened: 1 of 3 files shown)


sksavant:
WIP : Needs escaping regex match for member matches

lib/haml/filters/growstuff_markdown.rb
```diff
@@ -3,6 +3,8 @@
 module Haml::Filters
   module GrowstuffMarkdown
     CROP_REGEX = /\[([^\[\]]+?)\]\(crop\)/
+    MEMBER_REGEX = /\[([^\[\]]+?)\]\(member\)/
+    MEMBER_AT_REGEX = /(\@[^\[ \n]+)/
```
pozorvlak: Do you mean "an `@` followed by one or more non-whitespace characters"? I'd express that as `/@\S+/`. But actually, we're more restrictive about the characters allowed in usernames - we only allow [letters, numbers and underscores](https://github.com/Growstuff/growstuff/blob/dev/app/models/member.rb#L76). What you *really* want is `/@\w+/`.

lib/haml/filters/growstuff_markdown.rb
```diff
@@ -20,6 +22,32 @@ def render(text)
         end
       end
 
+      # turn [jane](member) into [jane](http://growstuff.org/members/jane)
+      expanded = expanded.gsub(MEMBER_REGEX) do |m|
+        member_str = $1
+        # find crop case-insensitively
```
pozorvlak: I think you mean "find member case-insensitively" :-)

lib/haml/filters/growstuff_markdown.rb
```diff
@@ -20,6 +22,32 @@ def render(text)
...
+          "[#{member_str}](#{url})"
+        else
+          member_str
+        end
+      end
+
+      # turn @jane into [@jane](http://growstuff.org/members/jane)
+      expanded = expanded.gsub(MEMBER_AT_REGEX) do |m|
+        member_str = $1
+        # find crop case-insensitively
```
pozorvlak: Again, s/crop/member.

spec/lib/haml/filters/growstuff_markdown_spec.rb
```diff
@@ -60,4 +74,27 @@ def output_link(crop, name=nil)
...
+  end
+
+  it 'converts @ member links' do
+    @member = FactoryGirl.create(:member)
+    rendered = Haml::Filters::GrowstuffMarkdown.render("@#{@member.login_name}")
+    rendered.should match /#{output_member_link(@member, "@#{@member.login_name}")}/
+  end
+
+  it "doesn't convert nonexistent @ members" do
+    rendered = Haml::Filters::GrowstuffMarkdown.render("@not-a-member")
```
pozorvlak: This isn't a legal member name. It's probably worth testing "invalidly-formatted name" and "valid but unused name" separately.

pozorvlak:
Aha, I see you've already fixed most of the problems I noticed! This is what I get for opening a browser tab before the weekend and not looking at it until now. It looks great, apart from the `not-a-member` test using an invalid name, which won't be matched - I think the best thing is to add another test that uses a valid but unused name. Also, please add yourself to `CONTRIBUTORS.md`.

Welcome to the project!

pozorvlak:
Wait, you're already in `CONTRIBUTORS.md`. Sorry! Just the test to fix, then.

sksavant:
@pozorvlak: Added the test and merged in dev changes.

PR #826, opened 2015-09-06
Fix #721 : Escape quick crop link in markdown

### LINKED ISSUE #721: unable to escape a quick crop link
I was trying to mention in a comment on Growstuff how someone should use the quick crop links feature.  I tried to escape the square brackets as per http://daringfireball.net/projects/markdown/syntax#backslash but it didn't work.

I suspect our special Growstuff filter might need to take escaping into account, but I'm not sure.

You can see this in action here: http://staging.growstuff.org/posts/skud-20150222-testing-link-escaping

sksavant:
Lookahead in markdown to verify \ is not present

Test for escaping crop link

Fixes #721

### DIFF
--- a/lib/haml/filters/growstuff_markdown.rb
+++ b/lib/haml/filters/growstuff_markdown.rb
@@ -2,7 +2,7 @@ require 'bluecloth'
 
 module Haml::Filters
   module GrowstuffMarkdown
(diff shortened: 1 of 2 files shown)


pozorvlak:
Looks good! Ordinarily I'd ask you to add yourself to `CONTRIBUTORS.md` before merging, but let's do that in #824.

PR #829, opened 2015-09-15
Upgrade database cleaner gem

CloCkWeRX:
https://github.com/DatabaseCleaner/database_cleaner/pull/364 resolves an issue where database_cleaner gets upset by open transactions; often in the way that Capybara keeps a few open on modern rails.

This upgrades us to the latest:
https://rubygems.org/gems/database_cleaner/versions/1.5.0


CloCkWeRX:
Ugh, I'm still affected by the bug - only completely nuking database cleaner from before/after hooks seems to fix it.

PR #831, opened 2015-09-15
Add additional capyabara driver options

CloCkWeRX:
Adds the ability to run feature tests via selenium if you configure it;
or run specs with ```GROWSTUFF_CAPYBARA_DRIVER=selenium bundle exec rake spec:features/```

I find this really useful because:
 * You can actually see what's going on
 * Combined with pry; you can interactively debug the current browser window and resolve issues with visibility
 * Its possible to test against multiple browsers with a bit of configuration

Defaults to poltergeist still for CI, normal use


cesy:
Yes, good idea, and thanks for keeping the default the same - selenium doesn't work on Nitrous.io so it's also helpful for casual developers as well as Travis.

Issue #605, opened 2014-12-13
Track seed trades


Skud:
As a member, when I request a seed trade from another member, currently all that happens is they get a PM from me, and we then sort things out between ourselves.

It would be good to actually track seed trades in the database.

A seed trade would have:
- seed_id -- the seeds being requested
- requester_id -- the member requesting it
- message -- text message from one person to another
- address -- postal address to send to (NOTE security issues! might want to have SSL before we do this; also make sure not to expose it via API etc)
- requested_date
- accepted_date
- declined_date
- sent_date
- received_date

The process would be:
- A requests seeds from B
- B is notified of request
- seed trade shows up in lists visible to both A and B (eg. "my seed trades")
- B accepts request; accepted_date is set
- OR B declines request; declined_date is set
- B can then mark as sent when they've been sent; sent_date is set
- A marks it as received when it has been; received_date is set

gustavor-souza:
Working on it

CloCkWeRX:
A related feature that we're lacking around trading seeds is a seed-trade-listing page ("Show me all seeds everywhere *currently available for trade*"). We have fragments in a few places, but it's not terribly easy to find and begin a valid trade at the moment.

PR #833, opened 2015-09-24
 Feature - Track seed trades - Issue #605

gustavor-souza:
Cover issue #605, now seed requests are tracked.


CloCkWeRX:
It snuck in with 7ef62d69f258e2d3d228e401320a8b1bd6bb7a5b.

From a cursory source code review (style + syntax) this looks OK so far. I might do a separate PR around the database config, as I've tripped over it a bunch and it looks like others do frequently as well.

Any chance of a few screenshots of the new UIs?

gustavor-souza:
**Member is the requester**

selection_008

selection_009

selection_010

selection_011

gustavor-souza:
**Member is the seed owner**

selection_012

selection_013

selection_014

selection_015

app/models/ability.rb
```diff
@@ -55,6 +56,35 @@ def initialize(member)
...
 
+      # can read seed requests if you are requester or seed owner
+      can :read,   SeedTrade, requester_id: member.id
+      can :read,   SeedTrade, seed: { :owner_id => member.id }
+      can :create, SeedTrade, requester_id: member.id
+
+      # you can't create a request for yourself
+      cannot :create, SeedTrade, seed: { :owner_id => member.id }
+
+      # you can decline if the trade was not accepted or declined
```
cesy: This means that if you hit the wrong button by mistake, there's no way back, right? I can see social arguments either way - it means a trade is a binding commitment, but is quite hard on people with mobility issues who may hit wrong buttons quite often.

app/models/seed_trade.rb
```diff
@@ -0,0 +1,46 @@
...
+  belongs_to :seed
+  belongs_to :requester, class_name: 'Member'
+
+  default_scope { order(created_at: :desc) }
+
+  # List seed trades both when member is requester and seed owner
+  scope :by_member,
+                 ->(member) { includes(:seed).where(
+                 'seeds.owner_id = ? OR requester_id = ?',
+                 member.id, member.id).references(:seeds) }
```
cesy: We should think about performance when this table gets big. We'll probably want to make the main view default to not_replied by_member and need to decide which order to do those in and which indexes to have on the database.

db/migrate/20150912205745_create_seed_trades.rb
```diff
@@ -0,0 +1,17 @@
+class CreateSeedTrades < ActiveRecord::Migration
+  def change
+    create_table :seed_trades do |t|
+      t.text       :message
+      t.text       :address
+      t.datetime   :requested_date
+      t.datetime   :accepted_date
+      t.datetime   :declined_date
```
cesy: Is there a reason for having two fields here, or should we have "replied_date" and a boolean "reply/answer" for accepted or declined? In theory any one record will only have either accepted or declined, not both, and that would mean we just need one index on replied_date.

cesy:
I'll come back with a more detailed review later, but this gives some initial thoughts. Given the size of this feature, it's probably worth someone else reviewing as well, and I suggest we target Release 10 as Release 9 is already late and oversize.

Thank you so much for contributing this - it's something we've wanted for a while, and will make a big difference to the community. Welcome to the project! I suggest you also add your name to the contributors file.

db/migrate/20150912205745_create_seed_trades.rb
gustavor-souza: Hello cesy, distinction between accepted_date and declined_date are here for some reasons:

1- Column 'status' in the seed_trade#index related view shows the status of the trade request.

2- Seed owner can only mark as sent once he has accepted the trade.

I think this is most in accordance with Skud's issue post.

app/models/ability.rb
gustavor-souza: Well, I did not considered this point of view. When I built it, I thought the javascript alert asking "are you sure?" would be enough.

what approach would you use here, stop restrictions on those buttons?

app/models/seed_trade.rb
gustavor-souza: I think I understand your concern, when we'll have many records would be more user friendly to show by default only "not replied" requests, and because of this we should think about the better way (in performance terms) to nest not_replied and by_member scopes results, right?

cesy: Yes, exactly. Would you mind taking another look at it, please?

app/models/ability.rb
cesy: Ah, I'd forgotten the JS alert - that should be fine, then.

db/migrate/20150912205745_create_seed_trades.rb
maco: But is there any situation in which we would fill in _both_ `accepted_date` AND `declined_date`? If not, there's no need for two columns.

maco:
We've merged a lot of pull requests lately. Would you mind fixing up the resulting merge conflicts? Thanks!

cesy:
Closing as all the code is in the new PR #949.

PR #834, opened 2015-09-24
GS-658 - internationalize the Contact page

dmarcoux:
Hello everyone,

This is my first time contributing to Growstuff. I saw that the application has to be internationalized and decided that it could be a good first step.

So now, about the actual commit. Let me know if something could be improved. I have never used Haml before, so I might have done something wrong. I ran the specs and everything passed.

Thank you!


jestallin:
#658

CloCkWeRX:
Looks good at first glance!

Issue #660, opened 2015-01-20
Migrate our Figaro environment variables to Rails secrets


Skud:
Rails 4 lets us store secrets in secrets.yml as per http://guides.rubyonrails.org/4_1_release_notes.html#config-secrets-yml rather than using a third-party gem like Figaro.

We should migrate to doing this the railsy way, not just to reduce our dependence on a bit of a finicky third-party gem, but also in the hopes that it might also help with #491 (at present we can't push our Figaro config up to Heroku).

CloCkWeRX:
https://github.com/Growstuff/growstuff/commit/92f8a843691a65cb3235cb6fdb08881660e6ceb9

PR #838, opened 2016-03-24
#660 Migrate Figaro environment variables to Rails secrets Issue 

hbalenda:
I changed application.yml over to secrets.yml, and used a constant SECRETS to simplify the reference to Rails.application.secrets. I was unable to test deployment to heroku without the proper keys. To send the config vars up to heroku, I added in the [heroku_secrets](https://github.com/alexpeattie/heroku_secrets) gem, which I was able to successfully test with heroku, but it's also possible to refactor [like so](http://stackoverflow.com/questions/21894761/rails-4-1-pushing-secrets-to-heroku), if you want to avoid using a gem.


Gemfile
```diff
@@ -83,7 +82,7 @@ gem 'omniauth-twitter'
 gem 'omniauth-flickr', '>= 0.0.15'
 
 # client for Elasticsearch. Elasticsearch is a flexible
-# and powerful, distributed, real-time search and analytics engine. 
+# and powerful, distributed, real-time search and analytics engine.
```
pozorvlak: It's not a big deal, but you can make reviewers' lives easier by separating whitespace-only changes into separate commits.

pozorvlak:
From a quick skim this looks OK, but:

1. You've broken the build, which means we can't merge this yet. I'm guessing you need to add some variable to travis.yml, but haven't dug into it.
2. What's the advantage of doing things this way instead of how we currently do them?

[Am off to the mountains for the weekend, so won't be able to respond for the next few days - sorry.]

Gemfile
hcbviolet: @pozorvlak That makes sense, I'll be sure to do that in the future.

hcbviolet:
@pozorvlak 1) Thanks for the quick feedback. I did find one variable I missed in travis.yml, but it’s still failing because as I understand it, TravisCI is looking in application.yml for the env vars. For secrets.yml, the variables will need to be loaded in via https://docs.travis-ci.com/user/encryption-keys/, which unfortunately I didn't foresee, and can’t transfer the env vars to Travis myself since I don’t have project access. (More info here: http://stackoverflow.com/questions/9338428/using-secret-api-keys-on-travis-ci)


2) I decided to work on issue [#660](https://github.com/Growstuff/growstuff/issues/660) because as a junior developer looking for an open source issue to tackle, this looked like something I could learn a lot from via using a new toolset to maintain the status quo instead of taking on a huge feature change, and I did learn a lot from the process. I think the main motivation for switching to secrets.yml is that it’s an auto-generated file maintained by the Rails core team vs being reliant on Figaro, and the heroku_secrets gem fills in the missing “transfer env vars to Heroku” feature.  

There are definitely pros and cons to the merge, and I understand if you’re worried about changing the process too much. In that case, is there another Rails-y issue you would recommend for a junior developer ? I do also have a mentor that I pair with. 

Have a great weekend and thanks again for your code review.

cesy:
From Skud's comments on the original issue, the rationale for the switch makes sense to me. So we just need one of us to set up the Travis bit and try again.

Welcome to the project, and thanks for your help. If you want something else to work on in the meantime, we've just had a load of new issues opened to help us get our next release live, several of which are small and simple.

Br3nda:
I think this PR can be abandoned now -- too many git conflicts

PR #842, opened 2016-03-28
Fix CVE 2015 5312 and more

CloCkWeRX:
* Upgrades minor version of rails (and leaves it floating on the 4.1 series)
 * Upgrades devise
 * Upgrades uglifier


maco:
Looks like there are 4 failing tests with this change.

CloCkWeRX:
Those should come good in a bit on travis; related to an rspec-rails bug

PR #843, opened 2016-03-28
Fix CVE-2015-7551 

CloCkWeRX:
See https://www.ruby-lang.org/en/news/2015/12/16/unsafe-tainted-string-usage-in-fiddle-and-dl-cve-2015-7551/

and

https://devcenter.heroku.com/changelog-items/780


CloCkWeRX:
Sigh, I'll have to see if thats just being flakey or legit regression in a bit. Will wait for the other security stuff to be happily merged; maybe see if we're compelled to upgrade ruby versions as part of the cedar-14 thing...

Issue #853, opened 2016-04-02
Harvests: Select control renders 'e.g. fruit' via include_blank; when it should be a prompt attribute


CloCkWeRX:
This is very similar to #848  ; but I decided to raise it because of the rendering bug on the index view

http://staging.growstuff.org/harvests/clockwerx-lettuce/edit
 * Do not choose an option, leaving e.g. fruit selected
 * Save

Result:
The plant part is rendered as /harvests on the index view
http://staging.growstuff.org/harvests

![image](https://cloud.githubusercontent.com/assets/365751/14226417/06527cc6-f929-11e5-8299-a0586f291ef6.png)

Recommendation: Use a prompt attribute, and mark this field required instead of include_blank

PR #858, opened 2016-04-02
Fix plant parts / harvests

CloCkWeRX:
#853


cesy:
Can you link to the issue in the pull request for easier cross-referencing, please? This is a sensible validation and better UX. But we have some test failures that need checking and updating - looks like the tests need fixing to add the newly-required field. Will we also have any issues with existing prod data - do we need a migration?

CloCkWeRX:
In all likelyhoold we won't have problems with existing data as plant_part is forced to be populated via the UI in current production.

image

CloCkWeRX:
@cesy this one's ready

PR #859, opened 2016-04-02
Add title validations

### LINKED ISSUE #857: Garden, Posts, Seed requests model needs length validation
1. New garden
 2. Add a title of
{code}
Australia (/ɒˈstreɪliə/, /ə-/, colloquially /-jə/),[10][11] officially known as the Commonwealth of Australia,[12] is an Oceanian country comprising the mainland of the Australian continent, the island of Tasmania, and numerous smaller islands. It is the world's sixth-largest country by total area. Neighbouring countries include Papua New Guinea, Indonesia and East Timor to the north; the Solomon Islands and Vanuatu to the north-east; and New Zealand to the south-east.
For about 50,000 years[13] before the first British settlement in the late 18th century,[14][15] Australia was inhabited by indigenous Australians,[16] who spoke languages grouped into roughly 250 language groups.[17][18] After the European discovery of the continent by Dutch explorers in 1606, Australia's eastern half was claimed by Great Britain in 1770 and initially settled through penal transportation to the colony of New South Wales from 26 January 1788. The population grew steadily in subsequent decades; the continent was explored and an additional five self-governing crown colonies were established. On 1 January 1901, the six colonies federated, forming the Commonwealth of Australia.
{code}
 3. Submit

Some kind of DB error is occuring behind the scenes no doubt.

We should:
 * Validate length of garden.title <= 255 char
 * Add the corresponding maxlength into the UI
 * Repeat with 'Location' field

Posts.title - same bug.

CloCkWeRX:
Fixes #857

### DIFF
--- a/app/models/garden.rb
+++ b/app/models/garden.rb
@@ -28,10 +28,14 @@ class Garden < ActiveRecord::Base
   scope :active, -> { where(:active => true) }
   scope :inactive, -> { where(:active => false) }
 
+  validates :location,
+    :length => { :maximum => 255 }
+
   validates :name,
     :format => {
       :with => /\S/
-    }
+    },
+    :length => { :maximum => 255 }
 
   validates :area,
     :numericality => {
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -3,6 +3,8 @@ class Notification < ActiveRecord::Base
   belongs_to :recipient, :class_name => 'Member'
   belongs_to :post
 
+  validates :subject, :length => { :maximum => 255 }
+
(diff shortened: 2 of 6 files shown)


cesy:
What happens to existing data that's longer than that - will the field chop it off, or will people be able to see it to edit it down to the new requirement?

CloCkWeRX:
There shouldnt be any, as mysql modelled these fields as varchar(255),
which caused the errors on submit

cesy:
Great. At some point we should also add some tests for it, but this is good to go for now. Let's log a new issue for beefing up all our data validation tests and finding any other bugs like this.

Issue #852, opened 2016-04-02
Harvests: We render 'description' headings when there are no descriptions


CloCkWeRX:
http://staging.growstuff.org/harvests

CloCkWeRX:
Might just be a matter of if foo.description.strip.blank?

CloCkWeRX:
Much happier on staging, closing.

PR #861, opened 2016-04-09
#852 Harvest description headings rendered when there's no description

hcbviolet:
Harvest description headings: Issue [#852](https://github.com/Growstuff/growstuff/issues/852) 

I checked out the harvest index view, and it was using the display_harvest_description method in harvests_helper.rb, which looked like this: 
```  
def display_harvest_description(harvest)
    if harvest.description.nil?
      "No description provided."
  end
```
But the view wasn't displaying "No description provided", it just displayed the heading. There were two options with this, I could remove the panel and heading entirely when there's no description, or fix the method that was already there. I went with the latter because I noticed the individual harvest view pages also render "No description provided" for no description, and so I changed the method to ```if harvest.description.empty?``` since when there’s no description, it shows up as an empty string (“”) instead of nil.

Here’s a screenshot of what it looks like with this edit: 
If you want the heading gone for no description, please let me know! I'm more than happy to do that instead.

### DIFF
--- a/app/helpers/harvests_helper.rb
+++ b/app/helpers/harvests_helper.rb
@@ -38,8 +38,8 @@ module HarvestsHelper
   end
 
   def display_harvest_description(harvest)
-    if harvest.description.nil?
-      "no description provided."
+    if harvest.description.empty?
+      "No description provided."
     else
       truncate(harvest.description, length: 130, separator: ' ', omission: '... ') { link_to "Read more", harvest_path(harvest) }
     end


CloCkWeRX:
Heading gone would make sense but have to update a bunch of UIs (not just harvest) - getting release out is probably more worthwhile. 
Changes LGTM.

cesy:
In some cases we'd need to check for nil as well, but I think we're safe here. Thanks!

PR #867, opened 2016-04-14
Update entire project to ruby 2.0 hash syntax

CloCkWeRX:
Result of:
```
$ rubocop --only HashSyntax --auto-correct

483 files inspected, 2018 offenses detected, 2018 offenses corrected
```

If this ends up with conflicts after other stuff is merged; easy enough to redo. Alternatively, its easy enough to do a section at a time - ie; all db migrations followed by all models; then controllers; specs... etc etc.

### DIFF
--- a/app/controllers/admin/orders_controller.rb
+++ b/app/controllers/admin/orders_controller.rb
@@ -8,7 +8,7 @@ class Admin::OrdersController < ApplicationController
 
   def search
     authorize! :manage, :all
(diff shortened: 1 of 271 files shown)


cesy:
Let's do this once Release 9 is live. What other bits do we still need fixed for that?

maco:
@CloCkWeRX

maco:
(had to close & re-open to trigger Travis to use the new version of Ruby)

maco:
Oh good, still green after the largish merges this morning.

PR #844, opened 2016-03-28
Bump ruby to 2.2.x

CloCkWeRX:
This one is certainly for after Release 9 has been in prod for a few days without problems, I suspect.

https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/ - the version of ruby we're using is a bit older and EOL'd.

Heroku supports 2.2.4 at the moment; local, specs seem happy, so staging deploy would be the next logical step.

PR #878, opened 2016-05-17
Upgrade a variety of development tools and utilities to current

CloCkWeRX:
```
$ bundle update selenium-webdriver guard guard-rspec rails factory_girl pg simplecov coveralls newrelic_rpm better_errors minitest
```

None of these should have much of an end user visible impact; unless say the pg gem contains problems.


CloCkWeRX:
Phwoar, that's a big coverage decrease for simply upgrading coveralls :P

maco:
On note of "automatic checks that get run" (like Coveralls), where I'm
working now uses Code Climate, so I've learned some about what the heck its
results mean, and I'm pondering making an attempt at making it happier.
(Right now, it's so ornery we're just ignoring it) For instance, I've
learned what it wants when it says a function is too complex, and well... I
think I should refactor the photo bit I wrote that's like 7 nested if's
deep.

Gemfile.lock
```diff
@@ -261,27 +257,29 @@ GEM
     letter_opener (1.3.0)
       launchy (~> 2.2)
     libv8 (3.16.14.7)
-    listen (3.0.2)
-      rb-fsevent (>= 0.9.3)
-      rb-inotify (>= 0.9)
-    lumberjack (1.0.9)
+    listen (3.1.4)
```
pozorvlak: This won't install for me without also upgrading Ruby to v2.2.

pozorvlak:
OK, I've finally worked out what's going on with the coverage. The lines covered haven't changed, but the new version of coveralls includes files with 0% coverage in the total. This is clearly correct, so we should accept the new, lower figure. Or rather, we should *not* accept it, and write tests covering the files we're missing!

maco:
👍

pozorvlak:
@CloCkWeRX, can you rebase this onto #844 because of the dependency issue I pointed out above?

CloCkWeRX:
Yeah can do (or merging that one first might be quicker, laptop being upset
at me)

maco:
I'm fine with merging that one first. It has a conflict at the moment, though.

maco:
@CloCkWeRX if you resolve the merge conflict on #844 I'll merge it tonight

PR #879, opened 2016-05-17
Add code climate to README

maco:
Our code climate score is *abysmal*, but now that I've been using it in my job for a few months, I have some ideas how to improve it. It's all about clean code (check out the book of that name from your local library) and readability. So let's add the grade to the README, and I'll figure out how to set up Code Climate to work with Travis CI.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,36 @@
+engines:
+  rubocop:
+    enabled: true
+  scss-lint:
+    enabled: true
+  shellcheck:
+    enabled: true
+  eslint:
+    enabled: true
+  coffeelint:
+    enabled: true 
+  brakeman:
+    enabled: true
+  bundler-audit:
+    enabled: true
+  duplication:
+    enabled: true
+    config:
+      languages:
+      - ruby
+      - javascript
+  fixme:
+    enabled: true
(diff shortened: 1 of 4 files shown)


maco:
There's [one last step](https://docs.travis-ci.com/user/code-climate/) for setting this up. Need to edit the .travis.yml, but it needs the token from Code Climate, and I don't think we can get that until after the git repo ownership is handed over to one of us active people.

pozorvlak:
Looks good so far. It looks like Code Climate also does coverage measurement: shall we remove coveralls at the same time as we add this, or run them in parallel for a while and then switch over once we're convinced they agree?

maco:
I don't know how well its coverage measurement works, so yeah, let's wait
and see.

.codeclimate.yml
```diff
@@ -0,0 +1,36 @@
+engines:
+  rubocop:
+    enabled: true
+  scss-lint:
+    enabled: true
+  shellcheck:
+    enabled: true
+  eslint:
+    enabled: true
+  cofeelint:
```
pozorvlak: Should this be "coffeelint"?

maco: oh, yep

PR #880, opened 2016-05-18
README tweak + security update

maco:
since there's no wiki right now, add a temporary note to the README including the directions for installing the pg gem on Mac, but really: update paperclip for this vulnerability: https://cwe.mitre.org/data/definitions/779.html

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -114,7 +114,7 @@ GEM
       jquery-rails (>= 3.0.0)
       jquery-ui-rails (>= 5.0.0)
       kramdown (>= 1.0.0)
-      paperclip (>= 4.0.0)
+      paperclip (>= 4.3.6)
       plupload-rails (>= 1.2.1)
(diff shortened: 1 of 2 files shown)


pozorvlak:
Can you split this into two commits, for the benefit of future code-archaeologists?

pozorvlak:
The CWE link you provided doesn't say anything about Paperclip specifically, it's just about "too much logging" problems generally. I think [this is the relevant vulnerability announcement](https://github.com/thoughtbot/paperclip/releases/tag/v4.3.6).

Also, you've edited the version of Paperclip depended-on by Comfortable Mexican Sofa, but not the version being installed (hence the Travis failures). Did you edit `Gemfile.lock` manually? I think the correct approach is to run `bundle update paperclip` and commit the result.

maco:
OH, yes, that's what I did because it wasn't in Gemfile. Will do.

PR #882, opened 2016-05-19
Install *all* gems in TravisCI

pozorvlak:
Previously we were only installing the gems needed for testing. This caused our CI to miss an incompatibility problem with a gem that was only needed for another environment:
https://github.com/Growstuff/growstuff/pull/878#discussion_r63745751


pozorvlak:
The build logs demonstrate that [the offending gem was installed this time](https://travis-ci.org/Growstuff/growstuff/jobs/131408365#L312).

PR #885, opened 2016-05-19
RFC: Use a "growstuff" database user in dev and test

pozorvlak:
Using the default "postgres" user, particularly without a password, makes me nervous.

We'd need to adjust environment-setup instructions to include a "create DB user" step, as soon as we recreate the environment-setup instructions...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ env:
 rvm:
 - 2.2.4
 before_script:
-- psql -c 'create database growstuff_test;' -U postgres
+  - psql -c "create user growstuff with createdb password 'growstuff';" -U postgres
+  - psql -c "create database growstuff_test owner growstuff;" -U postgres
 script:
 - bundle exec rake db:migrate --trace
 - bundle exec rspec spec/
--- a/config/database.yml
+++ b/config/database.yml
@@ -2,14 +2,15 @@ development:
   adapter: postgresql
   database: growstuff_dev
   host: localhost
-  user: postgres
-  password: password
+  user: growstuff
+  password: growstuff
 
 test:
   adapter: postgresql
   database: growstuff_test
   host: localhost
-  user: postgres
+  user: growstuff
+  password: growstuff
 
 production:
   adapter: postgresql


pozorvlak:
Question from @maco: if we merge this, are all developers going to need to drop their databases and redo `rake db:create`, `rake db:seed` etc? I don't think so, no: it should be possible to run `ALTER DATABASE growstuff_dev OWNER TO growstuff` as the `postgres` database user.

pozorvlak:
The main reason I put this PR up is to have a space to discuss the following two questions:

1. Is this a good idea?
2. What else needs to be done before we can merge it?

I think the answer to q1 is "yes", but I'm not 100% sure. Here's a first cut at q2, which I'll add to as the conversation goes on:

 - Document how to create the necessary DB user on all platforms we care about (OS X, Linux, Nitrous,...?)
 - Document how to transfer ownership of existing development databases.

More background: I sat down to develop Growstuff on yet another new machine, thought "ugh, now I need to set up Postgres again, maybe I could finally get Vagrant working, wait, how does it even work with one environment providing a password and the other not providing one for the same user?" and then I fell down a rabbit-hole full of yak hair.

pozorvlak:
Remaining yaks to be shaved on the "getting Vagrant working" front, BTW:

 - [ ] PhantomJS keeps crashing
   - The version of PhantomJS on Ubuntu 14.04 (and, for that matter, 15.10) is very very old - maybe that's why it's crashing?
   - Ubuntu 16.04 has a non-paleolithic version of PhantomJS, but it breaks all my assumptions about what a Vagrant box should look like
 - [ ] Tests are very very slow under Vagrant
   - Maybe we could use a provider other than Virtualbox?
     - LXC has worked well for me in the past, but it's Linux-only, hence the name
   - Using NFS shared folders instead of Vagrant's default sharing also often helps, but I bet configuring NFS sharing on OS X is a nightmare
      - Whatever we come up with has to be *easier* than configuring a development environment manually, or what's the point?

CloCkWeRX:
I'm happy with how we have it now with database.yml.example

pozorvlak:
@CloCkWeRX yes, I think that covers my immediate needs. I'll maybe re-visit this later as part of "provide a Vagrant-based development environment", but I'll close it for now.

PR #886, opened 2016-05-19
Fix "duplicated key: :class" warning in HAML code

pozorvlak:
The warning was

    app/views/harvests/_form.html.haml:32: warning: duplicated key at line 32 ignored: :class

We were specifying two `:class` attributes on an HTML tag; I've combined them into one.


CloCkWeRX:
LGTM

cesy:
Tests are failing, though - it looks unrelated, but we should investigate and check.

PR #890, opened 2016-05-20
upgrade gems and make API change for ruby-units

maco:
Locked gibbon to 1.2.x also, because if it upgrades to 2.x, it switches to the new MailChimp API, and the devs went "while we're changing API anyway..." and changed the gibbon API too, so... 😒 at the very least application.rb would need some changes, and I don't know what all else.


cesy:
I think it might make more sense to split out the three separate issues here. I'm not quite clear on why we want to ignore public, either.

maco:
public/ is where the generated stuff that happens each time you run "rails
server" or "rake assets:precompile" go. I was afraid "git commit -a" was
going to catch it.

The API change for ruby-units was part of making the result of "bundle
update" work. If I just did "bundle update" and committed, Travis would've
grumped at us.

maco:
Want me to make a PR that upgrades all the gems _except_ ruby-units?

pozorvlak:
I'm happy for the `ruby-units` change to be included since it's required by the library upgrade, but I'd prefer it if the change to `.gitignore` to be split out.

cesy:
Public is also where our error pages are, though, and we want to keep changes to them. So we'd need a more specific pattern for the bits we want to ignore.

Happy to include ruby-units based on this, though.

maco:
updated the diff

pozorvlak:
I think this means we've now merged all the "upgrade dependencies" patches
for release 10, so we should do another deployment soon. I'm around this
afternoon, from about 4pm UK time, or tomorrow evening.

maco:
:+1:

Now let's see if emailing emoji codes works

cesy:
4pm is good for me.

Issue #845, opened 2016-03-28
Do the Travis CI builds require sudo?


CloCkWeRX:
We're seeing
```
This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
If you require sudo, add 'sudo: required' to your .travis.yml
See https://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details.
```

because of

```
$ sudo service elasticsearch start
 * Starting Elasticsearch Server       it: error setting limit (Operation not permitted)
error: permission denied on key 'vm.max_map_count'
```

Can we run on containers (there's a lot of builds that just seem happy) with no changes; or do we need to explicitly require sudo in the .travis.yml

cesy:
I think we should probably require it explicitly for now to be safe, unless @pozorvlak disagrees.

pozorvlak:
I guess we should try removing the call to sudo, see if the build works without it, and if the build fails then add `sudo: required` explicitly.

pozorvlak:
It looks like this is a known issue with Travis, which they're [not going to fix](https://github.com/travis-ci/travis-ci/issues/3044), and that the warnings are harmless. Since setting `sudo: true` runs our code in a virtual machine rather than a container, leading to a 2x-3x slowdown (and possibly causing PhantomJS to be flaky), I think we should also ignore the warnings here.

PR #893, opened 2016-05-22
Setting sudo in Travis as elasticsearch uses it

cesy:
See #845


pozorvlak:
[It appears that the warnings are harmless](https://github.com/Growstuff/growstuff/issues/845#issuecomment-220841029), and setting sudo explicitly gives us a 2x slowdown, so I'm closing this unmerged.

PR #895, opened 2016-05-22
Precompile css and js with sass

cesy:
Since we switched from less to sass, we now need to precompile assets to get new css/js like the Bootstrap accessibility plugin (#803) to work.

### DIFF
--- a/script/deploy-tasks.sh
+++ b/script/deploy-tasks.sh
@@ -2,6 +2,9 @@
 
 # tasks to run at deploy time, usually after 'rake db:migrate'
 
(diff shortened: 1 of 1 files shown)


cesy:
Sorted - the links are from the CSS, so needed updating. We'll need to check after the deploy to production if the links change, and if so, change them to something more stable.

CloCkWeRX:
Is it worth us moving those to public/ or something static, outside of the asset pipeline if those don't get managed by it?

cesy:
Yes, probably a good idea.

PR #900, opened 2016-05-23
 Clip long text in definition lists within panels

### LINKED ISSUE #856: Garden card UI needs to apply overflow:hidden more strictly
image

CloCkWeRX:
Fixes #856

### DIFF
--- a/app/assets/stylesheets/overrides.sass
+++ b/app/assets/stylesheets/overrides.sass
@@ -276,6 +276,11 @@ $state-success-bg: lighten($green, 50%)
 .panel-footer
   height: 6em
 
+.panel
+  .dl-horizontal
+    text-overflow: ellipsis
+    overflow: hidden
+
 #gardens_panel_body
   height: 20em
 


pozorvlak:
Despite this PR passing CI, the automated build of `dev` failed. See #901.

maco:
One I merged last night also passed tests in PR and failed that way later.

pozorvlak:
OK, perhaps it's a difference in configuration between the dev builds and
the PR builds?

maco:
Don't know. It passed on re-run I think? I've also seen it on a local run.

pozorvlak:
OK, probably not a configuration difference, then. Oh dear.

maco:
Race condition? (I really don't like race conditions...)

pozorvlak:
Between the test code and the creation of the ElasticSearch index? Yeah, that sounds plausible. I guess the way to check would be to add a delay into the index creation, or run that test hundreds of times in a loop.

PR #905, opened 2016-05-24
refactor photo controller create for readability

maco:
Back when I made the horrible nested if of doom on the photos controller, we talked about eventually refactoring it for readability. Here we go. It could probably be abstracted further (item assignment and actual photo attachment lines), but this is much better already.

### DIFF
--- a/app/controllers/photos_controller.rb
+++ b/app/controllers/photos_controller.rb
@@ -63,35 +63,20 @@ class PhotosController < ApplicationController
     @photo.owner_id = current_member.id
     @photo.set_flickr_metadata
 
-    # several models can have photos. we need to know what model and the id
-    # for the entry to attach the photo to
-    valid_models = ["planting", "harvest", "garden"]
-    if params[:type]
-      if valid_models.include?(params[:type])
-        if params[:id]
-          item = params[:type].camelcase.constantize.find_by_id(params[:id])
-          if item
-            if item.owner.id == current_member.id
-              #  This syntax is weird, so just know that it means this:
-              #  @photo.harvests << item unless @photo.harvests.include?(item)
-              #  but with the correct many-to-many relationship automatically referenced
-              (@photo.send "#{params[:type]}s") << item unless (@photo.send "#{params[:type]}s").include?(item)
-            else
-              flash[:alert] = "You must own both the #{params[:type]} and the photo."
-            end
-          else
-            flash[:alert] = "Couldn't find #{params[:type]} to connect to photo."
-          end
-        else
-          flash[:alert] = "Missing id parameter"
-        end
+    if has_valid_key && has_item_id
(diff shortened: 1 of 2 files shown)


app/controllers/photos_controller.rb
```diff
@@ -63,35 +63,20 @@ def create
...
-        else
-          flash[:alert] = "Missing id parameter"
-        end
+    if has_valid_key && has_item_id
+      item = params[:type].camelcase.constantize.find_by_id(params[:id])
+      if item && member_owns_item(item)
+        #  This syntax is weird, so just know that it means this:
+        #  @photo.harvests << item unless @photo.harvests.include?(item)
+        #  but with the correct many-to-many relationship automatically referenced
+        (@photo.send "#{params[:type]}s") << item unless (@photo.send "#{params[:type]}s").include?(item)
```
CloCkWeRX: While it might seem to be anti DRY, a case statement would likely cover the purpose of valid_models + has valid_key + calling the right method.

IE:
```
collection = case params[:type]
  when 'garden'
     @photo.gardens
  when 'planting'
    @photo.plantings
  when 'harvest'
    @photo.harvests
  else
    raise "Invalid thingy"
end

collection << item unless collection.include?(item)
```

Then the whole thing could be shifted onto the photo model quite easily, rather than living in the controller - the controller only really cares that some kind of type gets added to the photo, not how sorta thing.

app/controllers/photos_controller.rb
```diff
@@ -63,35 +63,20 @@ def create
     @photo.owner_id = current_member.id
     @photo.set_flickr_metadata
 
-    # several models can have photos. we need to know what model and the id
-    # for the entry to attach the photo to
-    valid_models = ["planting", "harvest", "garden"]
-    if params[:type]
```
pozorvlak: Is it safe to drop these two checks?

app/controllers/photos_controller.rb
pozorvlak: I like this idea - as you say, it's a bit less DRY, but the existing metaprogramming is pretty hard to follow.

app/controllers/photos_controller.rb
```diff
@@ -134,6 +119,22 @@ def destroy
 
   private
 
+  def valid_models
+    ["planting", "harvest", "garden"]
+  end
+
+  def has_valid_key
```
pozorvlak: Oh, right, you've moved the checks into these helper methods. Sorry!

maco: The other thing to ponder is concerns. I'm not sure _exactly_ how they'd
change things as I haven't tried them yet.

pozorvlak:
I think this is an improvement as-is, but would be even better with @CloCkWeRX's magic-removal suggestion. @maco, how would you feel about that?

PR #912, opened 2016-05-25
Fix two small things (including: make tests stop failing)

maco:
* commit updated Gemfile.lock for Code Climate inclusion
* add config/factory_girl.rb so that rspec doesn't fail with a confusnig
  'Could not find mapping for #<Member...' error
  http://blog.thefrontiergroup.com.au/2011/03/reloading-factory-girl-factories-in-the-rails-3-console/

The former is why my photo PR is failing.

The latter happens when you make significant changes to a controller then re-run rspec. It's a Factory Girl / rspec caching issue. It's only a problem for devs running rspec locally, not Travis.

PR #915, opened 2016-05-25
set background color on autocomplete dropdown. Fixes #914

### LINKED ISSUE #914: Autocomplete dropdown has transparent background, so text is not legible
screen shot 2016-05-25 at 4 04 28 pm

maco:
Before:
screen shot 2016-05-25 at 4 04 28 pm

After:
screen shot 2016-05-25 at 4 05 03 pm

### DIFF
--- a/app/assets/stylesheets/overrides.sass
+++ b/app/assets/stylesheets/overrides.sass
@@ -244,7 +244,8 @@ html, body
 
 // Autosuggest
 
-.ui-autocomplete 
+.ui-autocomplete
+  background: white
   z-index: $zindex-tooltip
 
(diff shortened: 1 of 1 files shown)


pozorvlak:
Looks good! The build should pass once #912 is merged.

maco:
yep, build passed

CloCkWeRX:
Actually thinking a bit more; we might want to end up somewhere between this and https://gist.github.com/daz/2168334 (with appropriate padding)

cesy:
Yes - let's merge this as an improvement for now, but further prettiness would be good.

cesy:
Since the subject isn't linking properly, mentioning here that this fixes #914.

PR #922, opened 2016-05-26
Swap over to having "(Optional)" (or translation) after optional form…

### LINKED ISSUE #848: Optional/required fields UX is confusing
#718 was the original feature request, #779 implemented the work.

Two problems:
 * The placeholder text is much better suited to recommending the kind of input a user can add.
 * Select boxes now have a first option of 'optional'; as opposed to a 'Please select' or some other hint. See [planted from](http://staging.growstuff.org/plantings/new?crop_id=257) for example.

While there's some research about indicating optional fields (http://www.usabilitycounts.com/2009/11/18/form-design-and-the-fallacy-of-the-required-field/) the UI patterns suggested there are to put the word *Optional*

Ditto http://patternry.com/p=required-form-fields/

Another summary highlights an important point:
http://ux.stackexchange.com/questions/9684/how-to-represent-optional-fields

"Mark the odd field out"

maco:
… elements rather than using placeholder text

Fixes #848 

These tests seem silly. Maybe should be testing that the text "(Optional)" exists inside that element? Thoughts?

### DIFF
--- a/app/views/gardens/_form.html.haml
+++ b/app/views/gardens/_form.html.haml
@@ -17,15 +17,13 @@
     = f.label :description, :class => 'control-label col-md-2'
     .col-md-8
       = f.text_area :description, :rows => 6, :class => 'form-control'
-      %span.help-block.optional
-        = I18n.t 'optional', :scope => 'forms'
+      = render :partial => 'shared/form_optional'
 
   .form-group
     = f.label :location, :class => 'control-label col-md-2'
     .col-md-8
       = f.text_field :location, :value => @garden.location || current_member.location, :class => 'form-control', :maxlength => 255
-      %span.help-block.optional
-        = I18n.t 'optional', :scope => 'forms'
+      = render :partial => 'shared/form_optional'
       %span.help-block
         = t('.location_helper')
         - if current_member.location.blank?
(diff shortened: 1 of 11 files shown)


app/views/gardens/_form.html.haml
```diff
@@ -16,15 +16,18 @@
   .form-group
     = f.label :description, :class => 'control-label col-md-2'
     .col-md-8
-      = f.text_area :description, :rows => 6, :class => 'form-control', :placeholder => 'optional'
+      = f.text_area :description, :rows => 6, :class => 'form-control'
+      %span.help-block.optional
```
pozorvlak: These two lines get repeated a lot - perhaps factor them out into a helper or partial?

pozorvlak:
Agreed on the tests - they're testing implementation rather than intention. Your idea is much better - please do that!

For bonus points, factor out a [custom RSpec matcher](https://www.relishapp.com/rspec/rspec-expectations/v/3-3/docs/custom-matchers/define-matcher) for `be_optional`.

app/views/gardens/_form.html.haml
maco: Fair point

maco:
That last thing you said: SORCERY!

pozorvlak:
Muahahahahaha :-)​

pozorvlak:
I really like this - great work!

PR #924, opened 2016-05-27
Add capybara-screenshot to assist with feature testing

maco:


CloCkWeRX:
+1

By default I think these output to tmp/capybara/ ; do we know if we can get to that in travis builds?

Do we need to make any other configuration changes to make it run in a Real Browser, like setting up xvfb ?


https://docs.travis-ci.com/user/gui-and-headless-browsers/

maco:
Oh, I didn't even think about the Travis aspect of things! I'll dig into
that.

maco:
This is probably a @pozorvlak question. Only thing I've found so far is [this one about Jenkins](http://www.unknownerror.org/opensource/mattheworiordan/capybara-screenshot/q/stackoverflow/28309789/error-when-attempting-to-save-screenshot-after-test-failure-on-jenkins-using-cap) which seems to suggest that if the CI has 777 perms on the dir, you're good to go.

CloCkWeRX:
What we do for work is along the lines of:
```
  case ENV['headless']
  when 'false'
    Capybara.default_driver = Capybara.javascript_driver = :selenium
  else
    require 'headless'
    headless = Headless.new(:display => 99, :reuse => false, :dimensions => '1280x1024x16')
    headless.start
    Capybara.default_driver = Capybara.javascript_driver = :selenium
    Capybara.automatic_reload = true 
    
  end
```

allowing you to run in selenium on your desktop (thus driving the browser around infront of you), or defaulting to selenium in xvfb. I find it really handy, as you can stick a binding.pry in and interact with the browser in the middle of a test to debug the state of things - quite good for complex ember like apps.

The headless version is what we use with jenkins and it dumps out the screenshots as mentioned above - we simply treat them as build artifacts; lets them be clickable.

What we have now probably covers the second scenario already, but for interactive debugging its hard to beat selenium + pry

pozorvlak:
I expect that we can write to /tmp/capybara on Travis, but don't know. I'll open a PR based on top of this with a deliberate bug so we can find out :-)

As far as "getting access to screenshots from failed CI builds" goes, I think we'll need to use either [capybara-screenshot](https://github.com/mattheworiordan/capybara-screenshot#uploading-screenshots-to-s3)'s or [Travis'](https://docs.travis-ci.com/user/uploading-artifacts/) "upload to S3" option. I think it should be fine to use my AWS account for this.

I think the next steps are

1. Test what happens when CI finds a bug, as described above. If it doesn't break too horribly, merge this PR.
2. Add "upload to S3 if running in CI" functionality later.

pozorvlak:
Now we play the waiting game...

pozorvlak:
And the answer turns out to be "tests run as expected; the screenshots aren't put anywhere useful, but they don't cause errors either; an [entirely unrelated HTML dump](https://travis-ci.org/Growstuff/growstuff/jobs/133361323#L1054) is splatted into the output of rspec." I can live with that for now.

pozorvlak:
The "unrelated HTML dump" issue should be fixed by #927.

CloCkWeRX:
https://docs.travis-ci.com/user/uploading-artifacts/

Issue #913, opened 2016-05-25
Sidebar on Member page overlaps Planting cards


maco:
<img width="1174" alt="screen shot 2016-05-25 at 3 36 43 pm" src="https://cloud.githubusercontent.com/assets/401264/15553859/2ba77ff0-228f-11e6-9d7d-ce6e9c8bd0e0.png">

The DOM structure is:

```
%div.col-md-9
  %div.tabbable
    %div.tab-content
      %div.tab-pane.active
        %div.container
%div.col-md-3
```

Here's the trouble: that `%div.container`? That's 12 columns wide. 

We can't just move the entire `%div.tabbable` down, though, because then we'd get:

<img width="1195" alt="screen shot 2016-05-25 at 3 51 03 pm" src="https://cloud.githubusercontent.com/assets/401264/15554123/882a670a-2290-11e6-9cce-8de9dcf1c140.png">

So, we need to decide how we're rearranging things.

CloCkWeRX:
Does it resize happily with us setting the .container to span10?

CloCkWeRX:
![image](https://cloud.githubusercontent.com/assets/365751/15564011/05f40a94-234d-11e6-8316-2a616e7bb16e.png)

font-size: 85%;
activity list gets the list-inline class

We can maybe trim one or two extra headings as needed as well

maco:
Oh, just realized this only happens if you don't have a location set. On http://staging.growstuff.org/members/niftygrifty the map pushes everything below the sidebar, but on http://staging.growstuff.org/members/maco there's no map, so there's overlap.

Anyway, setting `.col-md-10` on `.container` results in this:

![screen shot 2016-05-26 at 12 52 00 am](https://cloud.githubusercontent.com/assets/401264/15564066/2a4e4f60-22dc-11e6-99f1-93c403a30a13.png)

maco:
A map or globe with a ? in place of the real map could work to push things down and indicate "hey, you haven't set your location yet!"  (The silly side of me says a picture of Carmen Sandiego)

pozorvlak:
Good idea! I think a map/globe would be more universally understood.​

Br3nda:
Looks fixed now

PR #932, opened 2016-05-27
Add a location not set layout (lets try this again!)

CloCkWeRX:
Based on MIT licenced icon-globe (http://fontawesome.io/3.2.1/icon/globe/)

Addresses #913

image

Probably could be done without the wording stuck into the image itself, but this seemed a lot easier to achieve


maco:
Can you add screenreader helper text? Bootstrap [has a class for that](http://www.w3schools.com/bootstrap/tryit.asp?filename=trybs_ref_hlp_sr-only&stacked=h)

CloCkWeRX:
Sorted

pozorvlak:
Looks sensible - do we also need to set an alt-text on the image?

Nitpick: best practice for long commit messages is to have a <50-character summary line (which can be read easily in pull request summaries, `git-log` output, etc), and put more details in subsequent paragraphs. I'd have written the message for your first commit as

```
Add globe icon for "location not set"

Based on MIT-licenced icon-globe:
http://fontawesome.io/3.2.1/icon/globe/
```

maco:
The image is a background image on a div, not an img tag. Consequently,
there's nothing to put an alt tag on, which is why I had him add the
screenreader-only text.

pozorvlak:
Brilliant, thanks for the explanation!

Issue #910, opened 2016-05-24
When a crop wrangler logs in, show a "there are crops waiting to be wrangled" notification if there are pending crop requests


maco:
I'm thinking of Bootstrap's `.alert .alert-info` bubbles.

pozorvlak:
Fixed by #933.

PR #933, opened 2016-05-29
#910: Crop approval flash alert for wranglers upon sign in

Thrillberg:
Issue 910.

This is my first non-trivial (ie, grammar) open source contribution. Did I do it right? What can I improve?

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -66,3 +66,4 @@ submit the change with your pull request.
 - Anthony Atkinson / [sha1sum](https://github.com/sha1sum)
 - Terence Conquest / [twconquest](https://github.com/twconquest)
 - Daniel O'Connor / [CloCkWeRX](https://github.com/CloCkWeRX)
+- Eric Tillberg / [Thrillberg](https://github.com/Thrillberg)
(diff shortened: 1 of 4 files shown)


maco:
Would you mind adding a feature test for this? We use Capybara to visit
pages and check that they have the messages we expect to see. They're in
spec/feature/

Thanks!

maco:
(And thank you for making us your first substantial contribution to open
source!)

Thrillberg:
Thanks! Just put a test up.

pozorvlak:
Looks great! I can think of two small improvements you could have made:

1. It makes life slightly easier for reviewers if you make whitespace-only changes in a separate commit from substantial code changes.
2. The alert message is hardcoded; we're slowly moving all user-visible text to `config/locales/{en,jp}.yml` so the site can be displayed in multiple languages. But we'll have to do that in lots of other places too, and it'll probably be easier to do it in a batch later.

I don't think either of these issues blocks this being merged. Please add yourself to `CONTRIBUTORS.md` and we're good to go!

Thrillberg:
Cool, thanks! I didn't know about the whitespace thing but it looks like Atom does that automatically--how annoying. I'm trying to figure out how to disable that feature.

PR #930, opened 2016-05-27
WIP: Add haml-lint gem, and fix the problems it finds

pozorvlak:
Annoyingly, Rubocop can't auto-correct HAML errors, [and won't gain this ability any time soon](https://github.com/brigade/haml-lint/issues/15).

About 1500 out of the 2633 problems found were "don't use double-quotes where single-quotes will do" warnings. I fixed these using the script and procedure at https://gist.github.com/pozorvlak/af43fed67c698b7e255b2feb91731acb.

The following also ought to be fixable in a (semi)-automated manner:
- [ ] SpaceBeforeScript: The - symbol should have one space separating it from code
- [ ] SpaceBeforeScript: The = symbol should have one space separating it from code
- [ ] RuboCop: Space inside { missing.
- [ ] RuboCop: Space inside } missing.
- [ ] RuboCop: Use `&&` instead of `and`.
- [ ] RuboCop: Use `||` instead of `or`.
- [ ] RuboCop: Surrounding space missing for operator `+`.
- [ ] RuboCop: Operator `?` should be surrounded by a single space.
- [ ] SpaceInsideHashAttributes: Hash attribute should start with one space after the opening brace
- [ ] SpaceInsideHashAttributes: Hash attribute should end with one space before the closing brace
- [ ] ImplicitDiv: `%div.pagination` can be written as `.pagination` since `%div` is implicit

Of the remaining problems, 364 are "LineLength: Line is too long", which I think will have to be fixed manually.

Fixes #902, or will eventually.

PR #935, opened 2016-05-30
Remove crowdfunding code

CloCkWeRX:
As per convo on #930 commits - do we need it at the moment? No. Will we need it again... maybe? But we can no doubt revert this change with minimal effort


pozorvlak:
@CloCkWeRX++

PR #937, opened 2016-05-31
Add test coverage for Seeds Helper

dv2:
Adding some test coverage while I am learning about the application, if that is ok

### DIFF
--- a/spec/helpers/seeds_helper_spec.rb
+++ b/spec/helpers/seeds_helper_spec.rb
@@ -1,8 +1,8 @@
 require 'rails_helper'
 
 describe SeedsHelper do
-  describe "display_seed_description" do
-    it "no description" do
+  describe "seed description" do
+    it "is missing" do
       seed = FactoryGirl.create(:seed,
         description: nil
       )
@@ -10,7 +10,7 @@ describe SeedsHelper do
       expect(result).to eq "no description provided."
     end
 
-    it "description is less than 130 chars" do
+    it "is less than 130 characters long" do
       seed = FactoryGirl.create(:seed,
         description: 'a' * 20
       )
@@ -18,7 +18,7 @@ describe SeedsHelper do
       expect(result).to eq 'a' * 20
     end
 
-    it "description is 130 chars" do
+    it "is 130 characters long" do
       seed = FactoryGirl.create(:seed,
         description: 'a' * 130
       )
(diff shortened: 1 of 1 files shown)


maco:
That's wonderful! We *need* some better coverage!

spec/helpers/seeds_helper_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+describe SeedsHelper do
+  describe "display_seed_description" do
+    it "no description" do
```
pozorvlak: The convention for RSpec test descriptions is that they should form a sentence describing the object under test (in this case, the seed description). Here that would be something like

```
  describe "seed description" do
    it "is missing" do
...
    it "is less than 130 characters long" do
```
etc.

This results in more readable output when the test fails. Currently you get

```
Failures:

  1) SeedsHelper display_seed_description no description
```

as opposed to

```
Failures:

  1) SeedsHelper seed description is missing
```

pozorvlak:
The tests look fine, apart from the naming issue I mentioned. Could you rebase this change onto the current dev branch to keep the history simple? If you don't know how to do that, I can walk you through it.

I'm also confused (and a little worried) that the coverage doesn't seem to be going up - I'll look in to that.

pozorvlak:
After some investigation: your specs are running, but for some reason they're not contributing to the coverage stats. This is good news, because it means our coverage is less awful than it appears :-)

spec/helpers/seeds_helper_spec.rb
dv2: Thank you for the feedback, I will make the changes to make is readable. Thanks

dv2:
@pozorvlak I would appreciate if you can walk me through the steps to rebase onto the current branch

pozorvlak:
OK, I think #942 fixes the problem with our coverage calculations. Hopefully once that's merged we'll see a nice improvement in coverage from this PR!

@dv2, the following commands should work to rebase your patch. I'm assuming your fork is called `origin` and the official Growstuff fork is called `upstream`, is that correct?

```
git pull --rebase upstream dev
git push -f origin dev
```

dv2:
@pozorvlak Thanks for fixing #942, hopefully we see better results in the coverage report.

And yes my fork is `origin` and official is `upstream`

I am going to try the commands and push

dv2:
@pozorvlak I still have to fix the descriptions, will push those changes in a little bit

pozorvlak:
@dv2 Looks great, thanks! I'll re-run the tests once #942 is merged so we can see how much you've improved our coverage.

PR #942, opened 2016-06-01
Fix coverage calculations

pozorvlak:
SimpleCov was reporting 0% coverage for any files loaded before `SimpleCov.start` was called, even if they were fully tested. This patch loads the application *after* starting SimpleCov, leading to more accurate coverage figures.


cesy:
LGTM, can be merged once tests pass.

Issue #920, opened 2016-05-26
Unknown error after creating a minimal seed


CloCkWeRX:
http://staging.growstuff.org/crops/lettuce currently shows "We're sorry, but something went wrong." after I created http://staging.growstuff.org/seeds/clockwerx-lettuce

I suspect some kind of formatting mishap

maco:
It's loading just fine for me. Hmm.

CloCkWeRX:
Can we see if anything crops up in newrelic?

maco:
Uh.... @pozorvlak? I don't have access to newrelic.

cesy:
I have access as well - Maco, can you follow the link from Heroku? Happy to do a screenshare over Hangouts or Skype if it helps. I can't see Clockwerx's error, as it apparently only keeps them for half an hour on staging, and that page loads fine for me.

maco:
Oh, it's a thing in Heroku. Well, I'm out of town today for a living
history event and tomorrow for a wedding, so I'll try that Monday.

I have seen a few other times staging has tripped in ways I haven't seen
production do. Can't remember exactly what though. Maybe a finished
planting?

cesy:
Seeds/lettuce gave this error on Show:
ActiveRecord::RecordNotFound
…ms/friendly_id-5.0.5/lib/friendly_id/
finder_methods.rb:  23:in `find'
…n-1.14.0/lib/cancan/model_adapters/
abstract_adapter.rb:  20:in `find'
…ems/cancancan-1.14.0/lib/cancan/
controller_resource.rb: 111:in `find_resource'
…ems/cancancan-1.14.0/lib/cancan/controller_resource.rb:  63:in `load_resource_instance'
…ems/cancancan-1.14.0/lib/cancan/controller_resource.rb:  32:in `load_resource'
…ems/cancancan-1.14.0/lib/cancan/
controller_resource.rb:  25:in `load_and_authorize_resource'
…ems/cancancan-1.14.0/lib/cancan/controller_resource.rb:  10:in `block in add_before_filter'

Crops/lettuce gave this error on Show:
ActionView::Template::Error: nil can't be coerced into Fixnum
/app/app/helpers/
crops_helper.rb:   6:in `block in display_seed_availability' 
/app/app/helpers/
crops_helper.rb:   4:in `display_seed_availability'

cesy:
I think that helper line 6 may need a catch for if the seed.quantity hasn't been specified, assume it is 1, or something like that?

[...]

PR #943, opened 2016-06-01
Deal with .nil quantity of a seed in rendering helper

CloCkWeRX:
For #920 

Will go write explicit coverage of the helper in a bit, if this is the actual issue.

### DIFF
--- a/app/helpers/crops_helper.rb
+++ b/app/helpers/crops_helper.rb
@@ -1,17 +1,21 @@
 module CropsHelper
   def display_seed_availability(member, crop)
     total_quantity = 0
-    member.seeds.each do |seed|
-      if seed.crop.name == crop.name && seed.quantity
-        total_quantity = total_quantity + seed.quantity
-      end
+
+    seeds = member.seeds.select {|seed| seed.crop.name == crop.name }
+
+    seeds.each do |seed|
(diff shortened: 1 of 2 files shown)


cesy:
It would also be good to have a feature test of the steps you did to create it on staging.

pozorvlak:
"Request to 'http://127.0.0.1:8081' failed to reach server, check DNS and/or server status" again :-( It looks like [this StackOverflow post](http://stackoverflow.com/questions/35743167/capybarapoltergeist-status-error-failed-to-reach-server) may have a fix - I'll try it out.

pozorvlak:
Running the tests locally without a network connection didn't produce any failures, so I don't think that can be the problem :-(

maco:
127.0.0.1 wouldn't depend on your network connection. That's localhost.

pozorvlak:
@maco yes, but [some people suggested](https://github.com/teampoltergeist/poltergeist/issues/375#issuecomment-42620085) that that message was erroneous, and the error could be caused by a remote resource timing out. But since the tests run fine without a network connection, that can't be the problem, which means the solution in the link I posted won't help.

maco:
Ah I only looked at the SO link, which mentioned something about
firewalling 127.0.0.1

pozorvlak:
@CloCkWeRX: looks good, please add tests.

Issue #816, opened 2015-08-24
Add and improve the sharing behaviour with Facebook (or any supported OpenGraph site)


CloCkWeRX:
Related: #733 #556

We have the ability to add a 'share on facebook' mechanism without needing the API key or anything like that.

Part 1:
We want to implement either of

* https://developers.facebook.com/docs/plugins/share-button
* https://developers.facebook.com/docs/plugins/like-button

Part 2:
Refactor the meta tags for each significant area being shared - Garden, User, Plant, Seed, etc - and populate an appropriate title, image, description etc.

Part 3: Assess which other opengraph consuming sites we want to add if any. For example, google+ consumes the current opengraph tags.

http://stackoverflow.com/questions/10397510/do-services-other-than-facebook-use-open-graph discusses this in a bit more depth.

http://support.addthis.com/customer/portal/articles/381237-third-party-buttons might be a good example of an 'all in one' solution

yifeiwu:
I used this gem for one of my projects. Would that work?
https://github.com/huacnlee/social-share-button

CloCkWeRX:
It would for rendering the buttons (we do have assets for these already though), but not for adding useful metadata in the header tags so much.

CloCkWeRX:
Warnings to fix up:

Share App ID Missing
```
The 'fb:app_id' property should be explicitly provided, Specify the app ID so that stories shared to Facebook will be properly attributed to the app. Alternatively, app_id can be set in url when open the share dialog.
```

and

```
Image Unavailable
Provided og:image, https://farm6.staticflickr.com/5556/15135715585_3ca67270d7_q.jpg could not be downloaded because it exceeded the maximum allowed sized of 8Mb or your server was too slow to respond.
```

which explains why sometimes we get a thumbnail, sometimes not.

CloCkWeRX:
So I think this is pretty close to doable now - we have a fb app id, all of the opengraph stuff, etc - just need the social share buttons

PR #947, opened 2016-06-01
Add better opengraph

CloCkWeRX:
This implements Part 2 of #816 - adding more descriptive metadata on the major pages; so that when we share a link on facebook etc it can be better understood/provide a better thumbnail.

History is a bit funky, so if we're broadly OK with the changes but want that cleaned up; I'll redo this as one big squashed commit.


app/views/comments/show.html.haml
```diff
@@ -1,4 +1,12 @@
 = content_for :title, @comment.post.subject
+- content_for :opengraph do
+  <meta property="og:image" content="#{avatar_uri(@comment.post.author, 150)}"/>
+  <meta property="og:image:user_generated" content="true" />
+  <meta property="og:title" content="#{@comment.post.subject}" />
+  <meta property="og:description" content="#{ strip_tags(@comment.post.body).split(' ')[0..20].join(' ') }" />
```
cesy: Any other risks of this doing something dodgy if the user puts in nasty content? Stripping the tags should solve most things, but let's have a brief think about any other security risks.

cesy:
We'll need to have a think about the original_uri failing on Travis, since that may just be an oddity of the way we're testing, but would be good to have a catch for it.

The broad principles of this look good, though I've just done a quick skim.

app/views/comments/show.html.haml
CloCkWeRX: I did think about it at the time - users putting in " are probably going to have a bad time of it; unfortunately. I'll probably go swap to the ```tag()``` method to clean this up in a bit.

CloCkWeRX:
One for me to fix up later:
```
Image Too Small
Provided og:image URL, http://staging.growstuff.org/assets/growstuff-apple-touch-icon-precomposed-86d9827705d880da540a437dd6207bfd5a827761c04ee903d46c2985e33c992d.png was not valid because it did not meet the minimum size constraint of 200px by 200px.
```

... which is probably the cause of the facebook thumbnail image being so inconsistent

pozorvlak:
Looks good! Please squash the history and fix the image-size issue you noticed.

PR #949, opened 2016-06-01
Track seed trades

CloCkWeRX:
An updated version of #833 

Feature - Track seed trades - Issue #605

If there's stuff from the original pull request we want to clean up, I'll tackle it in here. If needed, will squash the pull request to omit those database related commits - we already fixed that elsewhere.


db/migrate/20150912205745_create_seed_trades.rb
```diff
@@ -0,0 +1,17 @@
+class CreateSeedTrades < ActiveRecord::Migration
+  def change
+    create_table :seed_trades do |t|
+      t.text       :message
+      t.text       :address
+      t.datetime   :requested_date
+      t.datetime   :accepted_date
```
cesy: As per discussion on the earlier pull request, I think we want to change this to be replied date and what the reply was, rather than accepted date and declined date. I know Skud's original issue had these, but I don't think we'd thought it through at that stage.

pozorvlak: Marking as "needs: more work" until you have time to get to this, @CloCkWeRX. Otherwise, looks good!

cesy:
Would be really good to update this and get it in if you have time, @CloCkWeRX

Br3nda:
can we close this?

pozorvlak:
I dunno - it looks pretty far out of date. I just had a quick look at the conflicts, and *most* of them look straightforward to resolve, but a couple will take more thought. Also, @cesy's comment at https://github.com/Growstuff/growstuff/pull/949#r65378250 hasn't been addressed yet. So I guess we could close it with a request to reopen the PR when it's ready to merge?

PR #955, opened 2016-06-02
 Swap to geocoder 1.1.9

### LINKED ISSUE #953: Update geocoder gem
We're depending on a particular commit of the geocoder gem, because we needed a bug fix that was in the HEAD revision but hadn't been released at the time. That was back in August 2013, and the gem has had several releases since then.

Fixing this would allow us to run `bundle install` in a system without Git installed, such as a bare Docker container.

CloCkWeRX:
Fixes #953

This is the release just after what we had previously pinned

https://github.com/alexreisner/geocoder/commit/104d46 is roughly release 1.1.9

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,9 +60,7 @@ gem 'friendly_id', '~> 5.0.4'
 gem 'gravatar-ultimate'
 
 # For geolocation
-gem 'geocoder',
-  :git => 'https://github.com/alexreisner/geocoder.git',
-  :ref => '104d46'
+gem 'geocoder', '1.1.9'
 
 # For easy calendar selection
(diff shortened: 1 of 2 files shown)


pozorvlak:
Thanks! We should also try upgrading to a more recent version, [it looks like there have been quite a few improvements](https://github.com/alexreisner/geocoder/blob/master/CHANGELOG.md).

CloCkWeRX:
Yeah, just wanted to do the minimum so that nothing would break unexpectedly, will go do more now.

pozorvlak:
Awesome, thanks! And I agree that doing it in two stages was the Right
Thing.

PR #957, opened 2016-06-02
WIP: package the app + databases in Docker containers

pozorvlak:
I've been following the instructions at http://chrisstump.online/2016/02/20/docker-existing-rails-application/. Currently failing, as follows:

```
$ docker-compose run app rake db:create
Starting growstuff_db_1...
Cannot start container
c3e8a1db75bb6d02b6dfddf1e00157b9445db29625994eeadbb76d9eeb89156e: Cannot
link to a non running container: /growstuff_db_1 AS
/growstuff_app_run_6/db

$ docker-compose up db
Recreating growstuff_db_1...
Attaching to
[hangs forever]
```

I'm running docker-compose version 1.2.0 and docker version 1.8.2-fc22. My version of docker-compose is quite old, but it's the one that came from `dnf install` (on Fedora 22), and it would be annoying if we had to depend on a version that's more recent than the one in OS packages - the point of this is to make setup easy :-(

Good news: commit   7ea7fab should fix the "fatal: not a git repository" errors we've been seeing in deployment, and it's orthogonal to the rest of this PR. I'll split it out into its own PR later.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,16 +1,18 @@
-development:
+default: &default
   adapter: postgresql
-  database: growstuff_dev
-  host: localhost
+  encoding: unicode
+  host: db
+  port: 5432
   user: postgres
-  password: postgres
+  password: <%= ENV['POSTGRES_PASSWORD'] %>
+
+development:
+  <<: *default
+  database: growstuff_dev
 
 test:
-  adapter: postgresql
+  <<: *default
   database: growstuff_test
-  host: localhost
-  user: postgres
-  password: postgres
 
 production:
   adapter: postgresql
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.env
+.dockerignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,15 @@ FROM ruby:2.3.1-slim
 
 # Install essential Linux packages
 # We need nodejs for coffee-rails :-(
-RUN apt-get update -qq && apt-get install -y build-essential libpq-dev postgresql-client nodejs
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev postgresql-client nodejs wget
+
+# Install phantomjs for feature testing
+RUN wget https://assets.membergetmember.co/software/phantomjs-2.1.1-linux-x86_64.tar.bz2
+RUN tar xjf phantomjs-2.1.1-linux-x86_64.tar.bz2
+RUN cp phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/local/bin
 
 # Define where our application will live inside the image
-ENV RAILS_ROOT /var/www/docker_example
+ENV RAILS_ROOT /var/www/growstuff
 
 # Create application home. App server will need the pids dir so just create everything in one shot
 RUN mkdir -p $RAILS_ROOT/tmp/pids
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+# Base our image on an official, minimal image of our preferred Ruby
+FROM ruby:2.3.1-slim
+
+# Install essential Linux packages
+# We need nodejs for coffee-rails :-(
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev postgresql-client nodejs
+
+# Define where our application will live inside the image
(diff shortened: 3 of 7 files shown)


config/database.yml
```diff
@@ -1,16 +1,18 @@
-development:
+default: &default
   adapter: postgresql
-  database: growstuff_dev
-  host: localhost
+  encoding: unicode
+  host: db
```
cesy: Won't this break other people's environments?

config/database.yml
```diff
@@ -1,16 +1,18 @@
...
+  port: 5432
   user: postgres
-  password: postgres
+  password: <%= ENV['POSTGRES_PASSWORD'] %>
```
cesy: I like this.

pozorvlak: Yes, definitely - possibly by sourcing an `.env` file or similar. This is probably related to #838.

pozorvlak: I don't know if it's possible to use our existing `environment.yaml` - docker-compose expects environment files to be in `NAME=VALUE` format.

config/database.yml
pozorvlak: Yes (it's currently breaking Travis). We'll need to think about that. It might be OK to tell existing contributors "the new preferred way to run a development environment is with Docker, please throw away your hand-configured environments".

pozorvlak: In my last job, we had *two* "setup environment" processes, and skew between the two was a constant nightmare - people using the old setup kept unwittingly breaking the new one, which meant that the new setup was constantly playing catch-up and could never *quite* take over. So I think we *should* aim to have a clean switchover if possible, rather than trying to maintain "automated environment setup" and "manual environment setup" as equal first-class citizens.

pozorvlak:
[Fedora 24 packages docker-compose 1.7.0](https://apps.fedoraproject.org/packages/docker-compose) - maybe I just need to upgrade my OS. *gulp*

config/database.yml
cesy: Given that Docker for Windows and Mac is still in beta, though, we'll need something workable in the meantime.

pozorvlak: [Fully-native Docker for Windows/Mac](
https://blog.docker.com/2016/03/docker-for-mac-windows-beta/) is still in
beta, but Docker Toolbox appears to be non-beta on both [Windows](
https://docs.docker.com/engine/installation/windows/) and [OS X](
https://docs.docker.com/mac/step_one/). That runs Docker in a tiny Linux
VM, though, so developers on OS X and Windows would have to suffer
virtualization overhead. OTOH, our current suggestion for Windows is
"develop in a Linux VM".

pozorvlak: Aaargh GitHub y u no render my links properly?

pozorvlak:
Example of the git errors during deployment: https://travis-ci.org/Growstuff/growstuff/jobs/134783293#L2306

pozorvlak:
Upgrading my OS was more painful than I could possibly have imagined.

homu-gs:
:umbrella: The latest upstream changes (presumably #974) made this pull request unmergeable. Please resolve the merge conflicts.

pozorvlak:
I'm now able to run a Growstuff instance inside a Docker container, using Docker 1.11.2 and docker-compose 1.6.2 on Fedora 23. The test suite fails spectacularly because there's no phantomjs binary inside the container. Of course, we only want there to be one in test and development, so I'm not sure what the Dockerish approach is there.

Still to do:

 - [ ] Get PhantomJS tests passing
 - [ ] Add an elasticsearch container
 - [ ] Documentation

On the other hand, I'm less convinced that this work is necessary (though it's still fun, in a [Type II](http://www.backcountry.com/explore/type-ii-fun) kinda way) - Nitrous is much better-performing than I'd remembered, so maybe that's enough.

pozorvlak:
https://blog.codeship.com/testing-rails-application-docker/ recommends installing PhantomJS in the app container.

homu-gs:
:umbrella: The latest upstream changes (presumably #985) made this pull request unmergeable. Please resolve the merge conflicts.

Br3nda:
CI is stuck. Closing and then re-opening to kick it over now.

pozorvlak:
I haven't done any work on this in over a year, and I don't think it's all that important. Closing.

PR #963, opened 2016-06-02
Various bugfixes for facebook/opengraph

CloCkWeRX:
Oops; don't need to render quite so much on pages like http://staging.growstuff.org/crops/blackcurrant !


pozorvlak:
D'oh!

PR #965, opened 2016-06-03
Add test coverage for GardensHelper

dv2:
Adding test coverage

### DIFF
--- a/spec/helpers/gardens_helper_spec.rb
+++ b/spec/helpers/gardens_helper_spec.rb
@@ -58,7 +58,7 @@ describe GardensHelper do
       expect(result).to eq output
     end
 
-    it "has 2 planting" do
+    it "has 2 plantings" do
       garden = FactoryGirl.create(:garden)
       plantings = []
 
@@ -81,7 +81,7 @@ describe GardensHelper do
       expect(result).to eq output
     end
 
-    it "has 3 planting" do
+    it "has 3 plantings" do
       garden = FactoryGirl.create(:garden)
       plantings = []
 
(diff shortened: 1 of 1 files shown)


CloCkWeRX:
Looks good to me. 

You might be interested in rspec's context blocks, which let you describe the test setup. You are already doing it with 

```
it "is less than 130 characters long"
```

but this lets you explain it better:

```
describe "garden description" do
  context "with a short description" do 
    before :each do
      # Whatever setup 
    end

     it 'renders everything'
     it 'has an x'
     it 'does y'
    end
  end
end
```

spec/helpers/gardens_helper_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      result = helper.display_garden_plantings(plantings)
+
+      output = "<li>"
+      output += "10 " + link_to(crop.name, crop)
+      output += ", planted on #{plantings.first.planted_at}"
+      output += "</li>"
+      expect(result).to eq output
+    end
+
+    it "has 2 planting" do
```
cesy: I think this should be "has 2 plantings"

cesy:
Odd that coveralls doesn't think coverage has increased. Oh well - this is good stuff. I've flagged up a couple of grammar nitpicks, and then we can merge. Thank you!

spec/helpers/gardens_helper_spec.rb
dv2: @cesy I just updated this in 2 places, thanks for catching it

dv2:
@CloCkWeRX I will use them going forward, unless you want me to update the specs now

cesy:
Great, thank you!

PR #967, opened 2016-06-04
Add links to google, gardenate (for planting reminders)

### LINKED ISSUE #864: Prompt with planting reminders
Gardenate does this well: emailing to suggest what will be ready to plant.

See if there are open data sets or ways we can make it easy to know *when* to plant (and thus update growstuff).

Very similar to #524 but more distinctly about plants for your climate/region/etc.

image

Easiest implementation: link to gardenate so we don't double build the feature.

CloCkWeRX:
Fixes #864 

While we could go and capture all of this data for ourselves/build similar features, for a lot less effort and an extra tool for users it's simpler to link to existing resources.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,7 +21,8 @@ module ApplicationHelper
     link = "http://www.wolframalpha.com/input/?i=#{pid}+#{currency}"
     return link_to "(convert)",
       link,
-      target: "_blank"
+      target: "_blank",
+      rel: "noopener noreferrer"
   end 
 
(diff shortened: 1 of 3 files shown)


pozorvlak:
Makes sense - could you add some simple tests to cover it?

CloCkWeRX:
Tests added

cesy:
Target blank has some security risks, sorry - can we go back to direct links without the target, please?

cesy:
If you can fix the conflicts, I'll merge.

CloCkWeRX:
Yeah no problems @cesy ... just have to figure out what they are!

PR #972, opened 2016-06-07
Show concern

maco:
Making the controllers with no logic in their `show` method DRYer.

I think existing tests cover this, but another set of eyes would be good.


app/controllers/plantings_controller.rb
```diff
@@ -27,17 +28,6 @@ def index
     end
   end
 
-  # GET /plantings/1
-  # GET /plantings/1.json
-  def show
-    @planting = Planting.includes(:owner, :crop, :garden, :photos).friendly.find(params[:id])
```
CloCkWeRX: This is different to the behaviour we have now

CloCkWeRX:
So... I think we're actually just relying on the default implementation of ApplicationController.show (not a bad thing); but try something like

```
module SimpleShow
  extend ActiveSupport::Concern

  module ClassMethods

    def show
      @me = self.find(params[:id])
raise "HI"
      respond_to do |format|
        format.html # show.html.erb
        format.json { render json: @me }
      end
    end
  end
end

```

and view http://127.0.0.1:3000/alternate_names/1

Its not running the code in the concern as that loads happily.

maco:
:( well crap

app/controllers/plantings_controller.rb
pozorvlak: Drat it. Good catch!

OTOH, I'm surprised this wasn't caught by tests - we should fix that.

pozorvlak:
There are also a load of controllers where we only format HTML, eg AccountController - it might be worth adding a trivial JSON formatter to them by this sort of means, for API goodness.

pozorvlak:
I just tried removing all the `import SimpleShow` lines and running the test suite, and everything passed. Which either means our coverage is worse than we thought, or we're invoking default methods (and that the SimpleShow concern is being ignored). I think our next actions should be

- revert this patch for now, to fix the behaviour-change @CloCkWeRX identified
- in a new patch, eliminate all the default `show` methods
- ensure we're testing the non-standard behaviour of PlantingsController#show
- work out why the heck the SimpleShow method isn't being called.

pozorvlak:
Incidentally, it turns out that the drop in coverage is just a consequence of reducing the number of total lines in the files: if n > m, then (m-4)/(n-4) < (m/n).

cesy:
Removing release label as it got reverted.

PR #973, opened 2016-06-07
Fix "fatal: Not a git repository" errors during deployment

pozorvlak:
An alternative approach would be to remove our vendor fork and add https://github.com/sideshowcoder/active_merchant-paypal-bogus-gateway to our Gemfile - @CloCkWeRX++ for contacting the original author!

PR #974, opened 2016-06-07
Remove vendor directory

pozorvlak:
An alternative to #973.


pozorvlak:
@maco: I think this is done now, is there anything else you'd like to see? We could possibly kill commit 936aa17 on the grounds that those lines aren't doing any harm and we may some day put something else in `vendor`, but on the other hand YAGNI.

pozorvlak:
@maco: are you adding labels manually, or have you set up a bot to do that? I'm wondering if we should create a @growstuff_bots account for things like that.

maco:
Editing the config file? Manually, but CC gives us patches to apply to
ignore specific issues.

pozorvlak:
No, I meant the "in progress" label that appeared mere seconds after I
opened the PR :-)

maco:
No, I was wondering about those too, but actually, I think it might be
waffle.io It puts things in categories like Pivotal does, and if you have
the bug number in your branch name it shows the issue as being worked on.

Issue #901, opened 2016-05-23
Flaky CI builds


pozorvlak:
https://travis-ci.org/Growstuff/growstuff/jobs/132256191#L1038 failed despite [the PR being merged having previously passed CI](https://travis-ci.org/Growstuff/growstuff/builds/132193909). I _really_ don't want to get into a situation where our CI is unreliable and we re-run builds until they pass.

Here's the relevant part of the logs:

> [!!!] Index does not exist (Elasticsearch::Transport::Transport::Errors::NotFound)
> .......F.................................WARN: tilt autoloading 'tilt/kramdown' in a non thread-safe way; explicit require 'tilt/kramdown' suggested.
> ..........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Unused parameters passed to Capybara::Queries::SelectorQuery : ["http://localhost:8080/plantings/member1464-garden-magic-bean"]
> Unused parameters passed to Capybara::Queries::SelectorQuery : ["http://localhost:8080/plantings/member1464-garden-magic-bean-7d5f4ec1-26e4-4813-ad65-546692844d6d"]
> ...Unused parameters passed to Capybara::Queries::SelectorQuery : ["http://localhost:8080/harvests/member1469-magic-bean"]
> Unused parameters passed to Capybara::Queries::SelectorQuery : ["http://localhost:8080/harvests/member1469-magic-bean-5fcc0ba8-db9c-4ec2-8511-a036582eac8d"]
> ..........................................................................................................................................**_..........................................................................................................................................................................................................................................._.....................Unused parameters passed to Capybara::Queries::SelectorQuery : ["http://en.wikipedia.org/wiki/Magic_bean"]
> ........................
> [snip]
> Failures:
>   1) Planting a crop Marking a planting as finished
>      Failure/Error: page.should have_selector('ul.ui-autocomplete li.ui-menu-item a')
>        expected to find css "ul.ui-autocomplete li.ui-menu-item a" but there were no matches
>      # ./spec/support/feature_helpers.rb:11:in `select_from_autocomplete'
>      # ./spec/features/plantings/planting_a_crop_spec.rb:186:in`block (2 levels) in <top (required)>'
>      # /home/travis/.rvm/gems/ruby-2.2.4@global/gems/bundler-1.12.4/lib/bundler/cli/exec.rb:63:in `load'
[...]

PR #976, opened 2016-06-08
Remove "show" methods that just do the default

pozorvlak:
We can just fall back to the default implementation for these.

See comments on #972 for background.

### DIFF
--- a/app/controllers/alternate_names_controller.rb
+++ b/app/controllers/alternate_names_controller.rb
@@ -13,17 +13,6 @@ class AlternateNamesController < ApplicationController
     end
   end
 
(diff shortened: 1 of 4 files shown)


maco:
since our coverage is now suspect, we should check that html and ison are both covered by the super show

pozorvlak:
@maco: excellent point.

pozorvlak:
Oh look, a "PhantomJS client died" error: https://travis-ci.org/Growstuff/growstuff/jobs/136173245#L2088 Paging #901...

CloCkWeRX:
Will merge after tests go green

Issue #919, opened 2016-05-26
Replace all hard-coded pluralize("crop") and similar with Crop.model_name.human()


maco:
We hard-code strings for each model and then tell the system to pluralize them. This means these model names aren't automatically translated. We should instead rely on `.model_name.human(count)` which will pull from config/locale/en.yml (or es.yml or jp.yml etc) automatically, depending on which language of the site visitors are viewing.

## To fix this:

1. Search the whole codebase for `pluralize(`
2. If it is *not* using one of our model names (from `app/models/`) in quotes as the second thing in the parentheses, just skip over it. That's a more advanced case.
2. When you find it, check to see if the model being used is defined in the `activerecord` section [of the English translation file](https://github.com/Growstuff/growstuff/blob/dev/config/locales/en.yml)
3. If it isn't, add it to that file, just like the others (specifying `one` with the singular spelling and `other` with the plural)
4. Replace the `pluralize(count, "model")` call with `Model.model_name.human(:count => count)`

## Example:

The Activity section on the Member page has:

```
link_to pluralize(member.posts.size, "post"), posts_by_author_path(:author => member)
```

It should be:

```
link_to Post.model_name.human(:count => member.posts.size), posts_by_author_path(:author => member)
```

lucasnogueira:
Hello. I was just checking this task and I got a question. Changing the methods will make the actual number of objects disappear. For exemple:

`pluralize(2, 'foo')` equals `'2 foos'` while `Foo.model_name.human(:count => 2)` will result in just `'foos'`.

Is this the expected behaviour?

maco:
Oops, my instructions are not quite right, then. Still put in the number!

lucasnogueira:
So I suppose we should keep the `pluralize` but  replace the hard coded strings with the `Model.model_name.human` right?

maco:
Possibly something like this:

`link_to "#{member.posts.size} #{Post.model_name.human(:count => member.posts.size)}", posts_by_author_path(:author => member)`

So we're making a string for the link text, and that string is made up of:

- the number
- the correctly pluralized noun in whatever language the visitor is using

maco:
(Does that make sense?)

lucasnogueira:
Yeah it makes! But is there any problem with the `pluralize`? This option came to my mind but it feels a little bit repetitive. `pluralize(member.posts.size, Post.model_name.human)` would have the same result and would take care of the hardcoded string (that I believe is the problem here).

[...]

PR #980, opened 2016-06-09
Replace pluralize calls

lucasnogueira:
[#919](https://github.com/Growstuff/growstuff/issues/919)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -50,5 +50,13 @@ module ApplicationHelper
       default: :identicon
     })
   end
+
+  # Returns a string with the quantity and the right pluralization for a
+  # given collection and model.
+  def localize_plural(collection, model)
+    size       = collection.size
+    model_name = model.model_name.human(:count => size)
+    "#{size} #{model_name}"
+  end
 end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -22,7 +22,7 @@ module ApplicationHelper
     return link_to "(convert)",
       link,
       target: "_blank"
-  end 
+  end
 
   # Produces a cache key for uniquely identifying cached fragments.
   def cache_key_for(klass, identifier="all")
@@ -55,7 +55,7 @@ module ApplicationHelper
   # given collection and model.
   def localize_plural(collection, model)
     size       = collection.size
-    model_name = model.model_name.human(:count => size)
+    model_name = model.model_name.human(count: size)
     "#{size} #{model_name}"
   end
 end
--- a/app/views/forums/index.html.haml
+++ b/app/views/forums/index.html.haml
@@ -7,7 +7,7 @@
 - @forums.each do |forum|
   %h2= forum
   %p
-    = pluralize(forum.posts.size, "post")
+    = localize_plural(forum.posts, Post)
     |
     =link_to "Visit forum", forum
     |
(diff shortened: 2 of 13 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -43,4 +43,32 @@
       end
     end
   end
+
+  describe '#localize_plural' do
+    let(:post) { create(:post) }
+
+    context 'with a popullated colection' do
```
pozorvlak: Typo: "populated".

spec/helpers/application_helper_spec.rb
```diff
@@ -43,4 +43,32 @@
...
+      context 'with more than one element' do
+        before { create_list(:comment, 2, post: post) }
+
+        it 'returns a string with the quantity and the plural of the model' do
+          expect(localize_plural(post.comments, Comment)).to eq '2 comments'
+        end
+      end
+    end
+
+    context 'without a popullated colection' do
```
pozorvlak: Same here: "populated collection"

pozorvlak:
Looks good apart from the spelling nitpicks! I'm re-running the build, since it looks like you've run into issue #901 :-(

pozorvlak:
The build failure [appears not to be transient](https://github.com/growstuff/growstuff/issues/901#issuecomment-224861654), though it seems unrelated to your code - I'll dig into it a bit before merging. Sorry about this!

spec/helpers/application_helper_spec.rb
lucasnogueira: Sorry! This kind of things happen when English is not your mother tongue and you are coding up to late.

pozorvlak: No problem! English spelling can be [difficult even for native speakers](https://aeon.co/essays/why-is-english-so-weirdly-different-from-other-languages).

app/helpers/application_helper.rb
```diff
@@ -50,5 +50,13 @@ def avatar_uri(member, size = 150)
       default: :identicon
     })
   end
+
+  # Returns a string with the quantity and the right pluralization for a
+  # given collection and model.
+  def localize_plural(collection, model)
+    size       = collection.size
+    model_name = model.model_name.human(:count => size)
```
CloCkWeRX: We swapped over to ruby 2.0 syntax recent, so don't be afraid to use it :) Rubocop offers ways to automatically fix this for you too

app/views/forums/index.html.haml
```diff
@@ -7,7 +7,7 @@
 - @forums.each do |forum|
   %h2= forum
   %p
-    = pluralize(forum.posts.size, "post")
+    = localize_plural(forum.posts, Post)
```
CloCkWeRX: So, we have a collection of Posts already passed in; do we need to inject the model explicitly? I imagine 'yes' or 'its difficult' for empty collections, but for others do we have the ability to look at ```.first.class.name```?

app/helpers/application_helper.rb
lucasnogueira: Oh, good. The style link was down so I used what I saw around :P

app/views/forums/index.html.haml
lucasnogueira: My first approach was not to use the model name but a lot of tests broke because of empty collections. I assumed that this could be the case in every scenario so that is why I ended up doing this.

CloCkWeRX: Fair enough! I wonder what magic rails is doing under the hood to know what class name to .create or .build. Brief googling did not assist

app/helpers/application_helper.rb
maco: In theory, Code Climate should've pointed out the hash syntax as a "new issue". I wonder if it did. (Not sure I can see past results though)

PR #984, opened 2016-06-09
Rubocop config (ignore the red x)

maco:
Turn on all the Rubocop style checks so we can stop manually nitpicking. This might over time need some finessing.

Yes, this utterly fails the Code Climate check. That's because this commit makes it more strict, so now it's noticing all the stuff currently in the rest of the codebase.


CloCkWeRX:
Could we maybe start in the other direction; so we can incrementally fix issues across the entire codebase?

IE I think checks like 'space after comma' are or 'inconsistent indentation' are a lot easier to collectively agree on than 'line length exceeded', 'class missing documentation' or 'method has too many lines' (which is only 10 and rather arbitrary/might actually be made less understandable by refactoring).

maco:
Yeah, the conversation over in IRC was "uh, so should I just s/false/true/
on Code Climate's default Rubocop config and see what happens?" "Sure!" So
this answers "see what happens."

Ok, I did also shift the method lines from 30 being CC's default to 10
because that's what the Ruby style guide _actually_ says.

pozorvlak:
1,197 new issues - not as many as I'd expected, TBH :-) I'd request we're not too strict about line length - I write Python at work, and PEP8's 79-character limit is the bane of my life. Well, that and flaky CI :-/

maco:
Ok, I'm sitting here with the line length check turned off. Keeping in mind that we *can* tell Code Climate "just ignore it this one time" about any particular commit breaking any particular check, do we want to keep the 10 line function a la Ruby's style guide, put it at Code Climate's default 30, or what? We probably do want *some* limit there to avoid nasty code smells.

pozorvlak:
Put it at 30, I think. That's enough to detect the really egregious cases;
we can ratchet it down later if we want to.

[I have seen a [1600-line function](
https://gist.github.com/pozorvlak/1579912) in production C code before. The
horror, the horror...]

pozorvlak:
Put it at 30, I think. That's enough to detect the really egregious cases; we can ratchet it down later if we want to.

[I have seen a [1600-line function](https://gist.github.com/pozorvlak/1579912) in production C code before. The horror, the horror...]

maco:
That knocks it down to 889 issues found

homu-gs:
:umbrella: The latest upstream changes (presumably #1018) made this pull request unmergeable. Please resolve the merge conflicts.

cesy:
Let's get this in - if you can fix the conflicts, I'm happy to merge, and we can work on fixing the issues gradually.

maco:
Ok, I'll poke at this in a bit. Also, it turns out there's a CC integration
for GH Issues, so we can click on particular CC issues and turn them into
beginner bugs.

Br3nda:
Is this still in progress? I'd like send some robocop fixes, but master has a different code style config. 

If my vote counts, I'd like this merged, than we start chipping away at the style file by file.

cesy:
Yeah, it'll get merged as soon as the conflicts are fixed.

cesy:
Okay, this adds in SparkPost as well, but I'm happy to let that go in.

maco:
Darn it! I thought I removed that in the merge!

PR #964, opened 2016-06-02
Update rails to 4.2.0+

CloCkWeRX:
And a few other minor bits and pieces.

We get access to all of this:
http://edgeguides.rubyonrails.org/4_2_release_notes.html

... which is vaguely handy.

Specifics:
 * Upgrade activemerchant and unvendor, so we're no longer pinned to a 2 year old version
 * Upgrade rails
 * Fix deprecation errors relating to activerecord.find (we could certainly apply those separately if its cleaner)
 * Fix deprecation errors with active_job vs deliver (defaults to deliver_now)
 * Fix deprecation errors with view tests using old selectors with ```assert_select```, now that things are based on nokogiri
 * Rewrite 2-3 tests slightly, to relax expectations around rendering specific asset names as img src="..." or to better focus on the behaviour over implementation
 * Fix devise 4.2 deprecation errors. I have no idea why I didn't notice these in the previous pull request / didn't think we were affected, but they were appearing frequently.

PR #986, opened 2016-06-10
Update various gems - nokogiri, mime-types, parser, guard-rspec

CloCkWeRX:
First bits of #964 being redone in isolation


CloCkWeRX:
Provided tests are still happy, I'll shift onto the rails 4.2 update itself after this is merged

PR #987, opened 2016-06-10
Upgrade to rails 4.2.0+ (round 2)

CloCkWeRX:
Just like #964 only more mergable, working.


config/application.rb
```diff
@@ -112,5 +112,7 @@ class Application < Rails::Application
     # didn't work for us.
     config.cloudmade_key = '29a2d9e3cb3d429490a8f338b2388b1d'
 
+
```
pozorvlak: Please remove the extra whitespace.

config/environments/test.rb
```diff
@@ -15,7 +15,7 @@
   config.cache_classes = false
 
   # Configure static asset server for tests with Cache-Control for performance
-  config.serve_static_assets = true
+  config.serve_static_files = true
```
pozorvlak: Should we be setting this in `development.rb`?

db/migrate/20120903092956_devise_create_users.rb
```diff
@@ -34,7 +34,7 @@ def change
       # t.string :authentication_token
 
 
-      t.timestamps
+      t.timestamps null: true
```
pozorvlak: Are we currently creating null timestamps? We should probably fix that, if so.

config/environments/test.rb
CloCkWeRX: "config.serve_static_files configures Rails itself to serve static files. Defaults to true, but in the production environment is turned off as the server software (e.g. NGINX or Apache) used to run the application should serve static assets instead. "

So, should be ok to leave implicit

db/migrate/20120903092956_devise_create_users.rb
CloCkWeRX: Went with what the deprecation message suggested. It's unclear if we will set created_at = Time.now && updated_at = nil.

pozorvlak: OK. Let's go with this for now, and raise an issue to disallow NULL values
for timestamps (may require data migrations).

pozorvlak:
Please fix the whitespace issue, and then this is good to merge.

CloCkWeRX:
Re-running tests

Issue #950, opened 2016-06-01
Branded error pages


cesy:
Add Growstuff branding to the error pages currently in /public, particularly the Error 500 and 404 ones.

Include links back to the homepage, or ideally the whole standard header and footer. This can be done by copying the code from app/views/layout or app/views/home. You'll need to spell it out in full instead of just using HAML, because the error pages are used when Rails isn't fully working.

Prashanth261993:
Hi, I would like to work on this one. This is my first contribution, so if you could hep me with how to change its assignee, that would be great. Thanks!

maco:
Github is silly and prevents people from assigning, even to themself,
unless they already have commit (I think?) access. In the next week or so,
I'll have a bot setup to make it possible, but in the meanwhile, your
comment here is your claim.

Prashanth261993:
Oh thanks a lot. I'm unable to signup though. The confirmation seems to be redirecting to localhost:8080. I am guessing its a staging or production server. Is there a password for one of these members : test1, admin1?

maco:
Ah, yes, the password for test1 is password1 and for test2 is password2
etc. Similarly, wrangler1's password is wrangler1, and admin1's password is
password1

For future reference, if you look in db/seeds.rb you can search for a
username that's in the test db and find its password.

Prashanth261993:
The header and footer use bootstrap-css and bootstrap-sprockets. Is there an easy way to integrate them to files under public folder?

If not, which of the following should be done:
  i) Write custom css, js under <style> and <script> respectively
  ii) Do something similar to : [http://stackoverflow.com/questions/24235805/rails-4-how-do-i-create-a-custom-404-page-that-uses-the-asset-pipeline]

Prashanth261993:
Thanks @pozorvlak . Will definitely try.

Could you also please explain what cms_snippet_content(:footer1) does? 
It is part of _footer.html.haml.

It seems to be populating content in production but nothing under development. I can't find any code that initializes the content.

maco:
We have a CMS (content management system) called Comfy Mexican Sofa in use.
The footer content is stored in the DB and configured using the web
browser. I wouldn't worry about having much in the way of footer and all,
just making it matching in terms of color, font, etc.

Prashanth261993:
Hi, I have fixed the issue. I created a new branch, committed my changes and then merged dev into my branch. But when I try to push it, permission is denied.

Prashanth261993:
[...]

PR #988, opened 2016-06-13
Handled Issue #950 (Branded Error Pages)

Prashanth261993:
Moved error related views from public to errors (asset pipeline can be used).
Header and footer were included along with the error message.


CloCkWeRX:
Ooops, a few merge conflicts - can you git pull to get the latest?

Prashanth261993:
@CloCkWeRX  Pulled the growstuff-dev and merged with mine and resolved the conflict (CONTRIBUTORS.md). Will this do?

pozorvlak:
Merging, but noting for future reference: we were expecting this issue to be fixed by adding to the static HTML pages in `public/`, rather than replacing them with HAML but, it [looks like it's impossible to use site styles in static pages under Rails 4](https://mattbrictson.com/dynamic-rails-error-pages). I don't know how much we should worry about the "I can't report errors, because [my error-reporter is broken](http://scholar.harvard.edu/files/mickens/files/thenightwatch.pdf)" scenario - @cesy?

pozorvlak:
@Prashanth261993 congratulations on landing your first PR with us :-) I'd have used a summary message "Add branding to error pages (issue #950)" in line with [best practice](http://chris.beams.io/posts/git-commit/) (use the imperative, focus on what the change *does*), but that's a micro-nit.

pozorvlak:
Actually, I'd probably have said "Add branding to error pages" and then put "Fixes #950" somewhere in the body of the commit message, but we're well into [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality) at that point.

pozorvlak:
I just checked `http://staging.growstuff.org/wilma` and saw a blank page instead of a 404 error page. So I'm afraid this needs more work - reverting for now.

PR #989, opened 2016-06-13
update to SparkPost's mailserver

maco:
I've already pressed the buttons in Heroku to make it go. There is an open question of whether the API key Heroku generated will work. If not, I have the one SP generated, and we can swap that out in Heroku's config without needing another code push. That's something to check on staging.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -69,15 +69,15 @@ Growstuff::Application.configure do
   # Growstuff configuration
   config.action_mailer.default_url_options = { host: 'growstuff.org' }
 
-  config.action_mailer.smtp_settings = {
-      port: '587',
-      address: 'smtp.mandrillapp.com',
-      user_name: ENV['GROWSTUFF_MANDRILL_USERNAME'],
-      password: ENV['GROWSTUFF_MANDRILL_APIKEY'],
-      domain: 'heroku.com',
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -69,15 +69,15 @@
...
 
-  config.action_mailer.smtp_settings = {
-      port: '587',
-      address: 'smtp.mandrillapp.com',
-      user_name: ENV['GROWSTUFF_MANDRILL_USERNAME'],
-      password: ENV['GROWSTUFF_MANDRILL_APIKEY'],
-      domain: 'heroku.com',
-      authentication: :plain
+  ActionMailer::Base.smtp_settings = {
+      port:                 ENV['SPARKPOST_SMTP_PORT'],
```
pozorvlak: We've been using the convention that all Growstuff-related environment variables (which should at least include our SparkPost username and password) should have the prefix GROWSTUFF_.

Also, these environment variables should be documented in config/environment.yml.example.

maco: These var names are auto-generated when you turn on the SparkPost plugin in Heroku. Should I still go rename them all?

pozorvlak: Hmmm, maybe not. But please add a followup patch documenting them in application.yml.example.

PR #990, opened 2016-06-15
don't override our environment email settings with plain old sendmail

maco:
We were actually trying to send from the local server instance's install of sendmail

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -82,12 +82,6 @@ module Growstuff
       g.javascripts      false
     end
 
-    config.action_mailer.delivery_method = :sendmail
-      config.action_mailer.sendmail_settings = {
-        location: '/usr/sbin/sendmail',
-        arguments: '-i -t',
-        openssl_verify_mode: 'none'
-      }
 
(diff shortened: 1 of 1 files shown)


pozorvlak:
Looks like that code's been in there since 2012 - how did email ever work?

maco:
How email ever worked: our SPF record appears to say something along the
lines of "and any IP address heroku could ever possibly use is fine too" so
uh...i'll be fixing that.

pozorvlak:
I'm still confused, though:

```
$ heroku run --app=growstuff-staging ls -l /usr/sbin/sendmail
ls: cannot access /usr/sbin/sendmail: No such file or directory
$ heroku run --app=growstuff-prod ls -l /usr/sbin/sendmail
ls: cannot access /usr/sbin/sendmail: No such file or directory
```

So there isn't a `/usr/sbin/sendmail` on either our production or staging environments. Unless that's only become true recently, I don't think we can have been using sendmail to send out email.

maco:
Maybe heroku did recently remove it *shrug*

PR #979, opened 2016-06-08
Install PhantomJS 2.1.1 on Travis-CI

pozorvlak:
I've found it to be much more stable locally - hopefully it will help with our flaky CI issues (#901).

The travis_phantomjs directory is cached, so we should only have to download and unpack PhantomJS the first time.

Code mostly cargo-culted from
https://github.com/mitchlloyd/ember-orbit/commit/9d3afe3d0434d9e3572e018d65ea96857fe5fa25

PR #996, opened 2016-06-17
Block external URLs in feature tests

pozorvlak:
[StackOverflow](http://stackoverflow.com/a/36003988) and [GitHub commenters](https://github.com/teampoltergeist/poltergeist/issues/375#issuecomment-42620085) suggested that this might help with our intermittent "Request to http://localhost:8081 failed to reach server, check DNS and/or server status" errors in CI - the theory being that the error message is misleading, and it's actually a daughter request to an external service timing out. Seems worth a shot. It also ought to speed up testing, but the effect is slight (10s out of 3.5 minutes) on my machine.

I've blacklisted the URLs I could find in app/views or by tcpdumping a test run, because I couldn't get URL whitelisting to work.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -101,4 +101,14 @@ RSpec.configure do |config|
 
   # Allow just create(:factory) instead of needing to specify FactoryGirl.create(:factory)
   config.include FactoryGirl::Syntax::Methods
+
+  # Prevent Poltergeist from fetching external URLs during feature tests
+  config.before(:each, js: true) do
+    page.driver.browser.url_blacklist = [
+      'gravatar.com',
+      'mapbox.com',
+      'okfn.org',
+      'googlecode.com',
+    ]
+  end
 end


pozorvlak:
I'm going to re-run this until one of the following happens:

 - it becomes clear that we haven't fixed the problem
 - we demonstrate that we have, at at least the 5% significance level
 - I get bored.

Results:
1. passed!
2.
3.
4.
5.
6.
7.
8.
9.
10.
11.
12.
13.
14.
15.

pozorvlak:
OK, this isn't a total cure. But that's a [much lower false-negative rate than we had before](https://github.com/Growstuff/growstuff/pull/979), so I think it's still worth merging this.

pozorvlak:
Remembered that I'd failed to block `googlecode.com`. Aaargh. Really not sure I can be arsed to re-run the tests twenty times again.

pozorvlak:
1. passed!
2.

CloCkWeRX:
I'm good with this - for those services, if anything is critical, we should either build or find an existing fake we can use.

Issue #1006, opened 2016-06-20
readme irc server needs clarification


ctlewitt:
Specifying "irc.freenode.net" instead of "Freenode" in the README.md will help clarify for new IRC users where to find the chat server.

PR #1007, opened 2016-06-20
issue #1006: change "Freenode" to "irc.freenode.net" in README.md

ctlewitt:
Changed reference to "Freenode" to "irc.freenode.net" in README.md for clarification for newcomers to IRC.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -69,3 +69,4 @@ submit the change with your pull request.
 - DV Dasari / [dv2](https://github.com/dv2)
 - Eric Tillberg / [Thrillberg](https://github.com/Thrillberg)
 - Lucas Nogueira / [lucasnogueira](https://github.com/lucasnogueira)
+- Charley Lewittes / [ctlewitt](https://github.com/ctlewitt)
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ encourage participation from people of all backgrounds and skill levels.
 * [Issues](http://github.com/Growstuff/growstuff/issues) (features we're
   working on, known bugs, etc)
 * [Discussion forums](http://talk.growstuff.org/) (design ideas, planning releases)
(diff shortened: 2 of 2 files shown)


CloCkWeRX:
WE HAVE A FREENODE CHANNEL!?!!?!?!

More seriously, want to make that a link ( irc://irc.freenode.net/growstuff ) or something similar as well; so it's even easier?

pozorvlak:
@CloCkWeRX we didn't tell you about it? \*smacks forehead\* We thought you knew about it but didn't use IRC...

@ctlewitt Good catch, but as @CloCkWeRX says a link would be even better. And please add yourself to CONTRIBUTORS.md!

cesy:
Adding cross-reference in the comments because for some reason Github doesn't pick it up from the subject of a pull request, only the description or comments: #1006.

ctlewitt:
Okay, so I'm super new to IRC.  Do most people use a web browser to view IRC?  Like this:
https://webchat.freenode.net/
(The link above didn't work for me, so I ended up with a Google search that took me there.  Does  irc://irc.freenode.net/growstuff or something related work for you?)  

Otherwise, I found downloading an IRC client and then going to channel #growstuff on the irc.freenode.net network helpful.  

I think the former has a lower barrier to entry, but  neither seems to give me any history if I log off and then log on again later.

cesy:
People vary - some use a web browser, many use a client of some sort. For history, some people leave a client running all the time, or do something like screen+irssi, others have a bot that logs channel history somewhere. The web link has the lowest barrier to entry, as you say, and Gitter is also a low barrier to entry for people who aren't used to IRC.

maco:
If you're running a Linux desktop, the irc:// type link will probably
launch the default IRC client. On Windows and Mac, though, IRC clients
aren't part of a default install and so much less commonly used, that the
browsers don't have irc:// handlers configured by default. The https://
link is better since it'll work for anyone who doesn't have an IRC client
set up yet.

cesy:
Build failed with localhost intermittent issue, so restarted it.

PR #1008, opened 2016-06-21
Check for "forgot to commit Gemfile.lock" in CI

pozorvlak:
Quick-and-dirty script to catch an error we've missed a couple of times in code review, but which is not (AFAICT) covered by any of our existing checkers. Inspired by the following posts:

 - https://zachholman.com/posts/how-github-writes-blog-posts/
 - https://zachholman.com/talk/move-fast-break-nothing/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ rvm:
 before_script:
 - psql -c 'create database growstuff_test;' -U postgres
 script:
+- script/gemfile_check
 - bundle exec rake db:migrate --trace
 - bundle exec rspec spec/
 services:
--- a/script/gemfile_check
+++ b/script/gemfile_check
@@ -1,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require 'set'
+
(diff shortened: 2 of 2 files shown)


pozorvlak:
Tested by #1009, which should fail CI if this works.

pozorvlak:
#1009 failed CI, as expected.

pozorvlak:
BTW, it's occasionally legit to commit changes to `Gemfile.lock` but not to `Gemfile` - this happens when you run `bundle upgrade`.

pozorvlak:
Note that we're hard-coding the target branch to `dev`. This shouldn't be a problem since all PRs to `master` should first go through `dev`, but we may in future want to get the name of the target branch for such checks. This turns out to be [annoyingly nontrivial](https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/).

pozorvlak:
Though on the upside, that would also give us access to lots of information about the pull request, enabling checks for things like "is the author listed in `CONTRIBUTORS.md`?"

PR #1009, opened 2016-06-21
Test "test gemfile commits" - DO NOT MERGE!

pozorvlak:
Test of #1008. Should fail CI - please don't merge it even if it doesn't :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ rvm:
 before_script:
 - psql -c 'create database growstuff_test;' -U postgres
 script:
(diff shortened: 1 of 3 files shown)


pozorvlak:
OK, that failed as expected. I'm not sure if continuing to run the full test suite after a failure is the Right Thing, though - perhaps we should split out a separate "static checks" entry of the build matrix?

cesy:
Yeah, we should have static checks separately at the beginning, ideally. But this is still an improvement.

PR #1013, opened 2016-06-21
WIP Code cleanup

maco:
(based on #984 )

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -79,6 +79,7 @@ group :production, :staging do
   gem 'memcachier'
   gem 'rails_12factor' # supresses heroku plugin injection
   gem 'bonsai-elasticsearch-rails'  # Integration with Bonsa-Elasticsearch on heroku
+  gem 'sparkpost_rails'
 end
 
 group :development do
--- a/app/controllers/crops_controller.rb
+++ b/app/controllers/crops_controller.rb
@@ -11,11 +11,11 @@ class CropsController < ApplicationController
     @sort = params[:sort]
     if @sort == 'alpha'
       # alphabetical order
-      @crops = Crop.includes(:scientific_names, {plantings: :photos})
+      @crops = Crop.includes(:scientific_names, { plantings: :photos })
       @paginated_crops = @crops.approved.paginate(page: params[:page])
     else
       # default to sorting by popularity
-      @crops = Crop.popular.includes(:scientific_names, {plantings: :photos})
+      @crops = Crop.popular.includes(:scientific_names, { plantings: :photos })
       @paginated_crops = @crops.approved.paginate(page: params[:page])
     end
 
@@ -77,14 +77,14 @@ class CropsController < ApplicationController
   # GET /crops/1
   # GET /crops/1.json
   def show
-    @crop = Crop.includes(:scientific_names, {plantings: :photos}).find(params[:id])
+    @crop = Crop.includes(:scientific_names, { plantings: :photos }).find(params[:id])
     @posts = @crop.posts.paginate(page: params[:page])
 
     respond_to do |format|
       format.html # show.html.haml
       format.json do
         render json: @crop.to_json(include: {
-          plantings: { include: { owner: { only: [:id, :login_name, :location, :latitude, :longitude] }}}
+          plantings: { include: { owner: { only: [:id, :login_name, :location, :latitude, :longitude] } } }
         })
       end
     end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,25 +1,24 @@
 module ApplicationHelper
 
   def price_in_dollars(price)
-    return sprintf('%.2f', price / 100.0)
+    sprintf('%.2f', price / 100.0)
   end
 
   # 999 cents becomes 9.99 AUD -- for products/orders/etc
   def price_with_currency(price)
-    return sprintf('%.2f %s', price / 100.0,
-        Growstuff::Application.config.currency)
+    sprintf('%.2f %s', price / 100.0, Growstuff::Application.config.currency)
   end
 
   def parse_date(str)
     str ||= '' # Date.parse barfs on nil
-    return str == '' ? nil : Date.parse(str)
+    str == '' ? nil : Date.parse(str)
   end
 
   def forex_link(price)
     pid = price_in_dollars(price)
     currency = Growstuff::Application.config.currency
     link = "http://www.wolframalpha.com/input/?i=#{pid}+#{currency}"
-    return link_to "(convert)",
+    link_to "(convert)",
       link,
       target: "_blank"
   end
--- a/app/helpers/harvests_helper.rb
+++ b/app/helpers/harvests_helper.rb
@@ -5,13 +5,13 @@ module HarvestsHelper
     weight = display_weight(harvest)
 
     if human_quantity && weight
-      return "#{human_quantity}, weighing #{weight}"
+      "#{human_quantity}, weighing #{weight}"
     elsif human_quantity
-      return human_quantity
+      human_quantity
     elsif weight
-      return weight
+      weight
     else
-      return 'not specified'
+      'not specified'
     end
   end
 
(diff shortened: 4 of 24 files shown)


app/models/member.rb
```diff
@@ -235,32 +229,31 @@ def update_newsletter_subscription
...
+    return true if Rails.env.test? && !testing
     gb = Gibbon::API.new
-    res = gb.lists.subscribe({
-      id: Growstuff::Application.config.newsletter_list_id,
-      email: { email: email },
-      merge_vars: { login_name: login_name },
-      double_optin: false # they already confirmed their email with us
-    })
+    gb.lists.subscribe({
+                          id: Growstuff::Application.config.newsletter_list_id,
```
CloCkWeRX: Not a huge fan of this style of indentation, as it runs counter to the rationale behind line length guidance (all roughly readable without too much scrolling as opposed to fits on a punchcard) and is pretty readable in the former format. If it's a specfic check as opposed to generic whitespace checks; might be worth turning it off.

Gemfile
```diff
@@ -79,6 +79,7 @@ group :production, :staging do
   gem 'memcachier'
   gem 'rails_12factor' # supresses heroku plugin injection
   gem 'bonsai-elasticsearch-rails'  # Integration with Bonsa-Elasticsearch on heroku
+  gem 'sparkpost_rails'
```
CloCkWeRX: Extra bits from another change?

app/helpers/application_helper.rb
```diff
@@ -1,25 +1,24 @@
...
-    return str == '' ? nil : Date.parse(str)
+    str == '' ? nil : Date.parse(str)
   end
 
   def forex_link(price)
     pid = price_in_dollars(price)
     currency = Growstuff::Application.config.currency
     link = "http://www.wolframalpha.com/input/?i=#{pid}+#{currency}"
-    return link_to "(convert)",
+    link_to "(convert)",
```
CloCkWeRX: Want to maybe cram this onto one line?

Gemfile
maco: gahhh that came in with the merge from the rubocop.yml -editing branch which i apparently made off of a wrong branch or something. *sigh* (that was part of experimenting with using SP's API instead of just using SMTP)

app/models/member.rb
maco: Yep, I can turn off the parameter alignment check. That's around like 400 in the .rubocop.yml

app/helpers/harvests_helper.rb
```diff
@@ -20,20 +20,20 @@ def display_human_quantity(harvest)
       if harvest.unit == 'individual' # just the number
         number_to_human(harvest.quantity, strip_insignificant_zeros: true)
       elsif ! harvest.unit.blank? # pluralize anything else
-        return pluralize(number_to_human(harvest.quantity, strip_insignificant_zeros: true), harvest.unit)
+        pluralize(number_to_human(harvest.quantity, strip_insignificant_zeros: true), harvest.unit)
```
CloCkWeRX: With these if-elsif-else and returns; we should probably try to return early for the error conditions/edge cases
```
def foo()
    return nil if impossible?
    return nil if boring_dont_care_about?

    pluralize(number_to_human(harvest.quantity, strip_insignificant_zeros: true), harvest.unit)
end
```

If find that cleans up a lot of the intent of the method

app/mailers/notifier.rb
```diff
@@ -4,7 +4,7 @@ class Notifier < ActionMailer::Base
 
   def verifier()
     if ENV['RAILS_SECRET_TOKEN']
-      return ActiveSupport::MessageVerifier.new(ENV['RAILS_SECRET_TOKEN'])
+      ActiveSupport::MessageVerifier.new(ENV['RAILS_SECRET_TOKEN'])
```
CloCkWeRX: Another one I'd flip

```
def verifier
  raise "Areeegghhj" unless ENV['RAILS_SECRET_TOKEN']

  ActiveSupport::MessageVerifier.new(ENV['RAILS_SECRET_TOKEN'])
end
```

app/controllers/crops_controller.rb
```diff
@@ -77,14 +77,14 @@ def search
...
-    @crop = Crop.includes(:scientific_names, {plantings: :photos}).find(params[:id])
+    @crop = Crop.includes(:scientific_names, { plantings: :photos }).find(params[:id])
     @posts = @crop.posts.paginate(page: params[:page])
 
     respond_to do |format|
       format.html # show.html.haml
       format.json do
         render json: @crop.to_json(include: {
-          plantings: { include: { owner: { only: [:id, :login_name, :location, :latitude, :longitude] }}}
+          plantings: { include: { owner: { only: [:id, :login_name, :location, :latitude, :longitude] } } }
```
CloCkWeRX: This one is probably neater with line breaks rather than just spaces alone

app/models/post.rb
```diff
@@ -16,12 +16,12 @@ class Post < ActiveRecord::Base
     self.body.scan(Haml::Filters::GrowstuffMarkdown::MEMBER_REGEX) do |m|
       # find member case-insensitively and add to list of recipients
       member = Member.where('lower(login_name) = ?', $1.downcase).first
-      recipients << member if member and not recipients.include?(member)
+      recipients << member if member and ! recipients.include?(member)
```
CloCkWeRX: Might want to rewrite the logic for these, it reads funny... maybe a .uniq! at the end instead of checking the collection?

CloCkWeRX:
I'm 50-50 on these as a whole - 100% yes to b50127c16647fbd82f13527fc8c8ed25bf3e1ae5 61afcd450318dd097302bbde9fde0a26ab172e55 8fe1c173875587fc8b1411690853574d8c8b27e0 d5fbbad4137bf39a48382f254e840dce8887d72c

Others I think are an indicator we should rework the logic - a87785d e8f9f768fd341d45186fafca8c03f7b20387ea6e

app/models/member.rb
CloCkWeRX: Let's see what others think and nuke if there's consensus

maco:
@CloCkWeRX agreed on there being logic to rework scattered about. Maybe make the changes you've suggested here, merge that (which'll pull in the rubocop config changes too), then do like one file at a time of clean up over the next...while? Rubocop will also tell us about logically-complex functions; I was just sticking to style changes for now. (I'm a little afraid to do much in the way of logic rewrites depending on the test coverage for a given file)

CloCkWeRX:
Yeah we can cherry pick the good to go commits I think... the other ones we can ticket for refactoring and merge the style updates; so we don't lose track of them (and add unit test coverage as part of refactoring!)

maco:
Cherry picking is a git skill level I have not yet attained.

CloCkWeRX:
Ah, it's very easy!

It basically is a clever copy/paste of an entire commit. Unlike a merge, it cuts out just that change.

Try this:

```
$ git checkout dev
$ git checkout -b code_cleanup2_electric_boogaloo
$ git cherry-pick b50127c 
$ git cherry-pick 61afcd4
$ git cherry-pick 8fe1c17
$ git cherry-pick d5fbbad
$ git push origin code_cleanup2_electric_boogaloo
```

Another way that can be useful to ensure you get clean feature branches:
```
$ git remote add upstream git@github.com:Growstuff/growstuff.git
```
then you can just do
```
git fetch upstream
git checkout upstream/dev
git checkout -b my_new_feature_branch
```

homu-gs:
:umbrella: The latest upstream changes (presumably #1018) made this pull request unmergeable. Please resolve the merge conflicts.

CloCkWeRX:
Applied 
$ git cherry-pick 61afcd4
$ git cherry-pick 8fe1c17
$ git cherry-pick d5fbbad

manually. Will revisit the && / and changes in a sec

CloCkWeRX:
Added a few more of these commits with the cleanup. Still some bits unaddressed; but should tackle seperately.

PR #1020, opened 2016-07-08
Release 13

CloCkWeRX:
We should probably put this out even if it means bumping a lot of stuff from the roadmap.


pozorvlak:
Agreed, let's get a release out soon even if it doesn't include all the stuff we'd like. @Cesy, @maco?

pozorvlak:
Build failed with an "ElasticSearch::ServiceUnavailable" error; re-running.

cesy:
Yes, agreed.

pozorvlak:
I'm away camping this weekend and likely to have limited Internet access; I
won't be able to help with the deploy until Monday, but feel free to
proceed without me :-)

CloCkWeRX:
Staging looks briefly ok, though I only clicked around a little given how little user facing changes are in here...

Issue #904, opened 2016-05-23
Top nav on mobile layout obscures content


CloCkWeRX:
The searchbar obscures too much on mobile (galaxy s6/staging)

Br3nda:
Can you add a screen show of where you mean?

CloCkWeRX:
@Br3nda I think you might have fixed this at some point, the searchbar used to be in the header and overflow weirdly.

PR #1023, opened 2016-07-28
Top nav on mobile layout obscures content #904

polveenomials:
Hi. I fixed the search bar and added a button on the side for the convenience of the mobile users. I also made the image responsive wherein if the site is opened in mobile, the image that'll appear would just be the icon. Hope this can help. 😄


CloCkWeRX:
Ah nice work, @polveenomials but I may have beaten you to it a bit with #1019 (check out how it behaves on staging).

Did you have a screenshot of what the responsive logo looks like? That's potentially worth adding in by itself

polveenomials:
Here's how it looks like in mobile. @CloCkWeRX 
screen shot 2016-07-28 at 12 48 45 pm

PR #1024, opened 2016-07-28
#981 Add mapbox v4

### LINKED ISSUE #981: Update to Mapbox API v4
We appear to be using the [deprecated v3 API](https://c.tiles.mapbox.com/v3/0/2/1/1.png), and this [may be causing build failures](https://github.com/Growstuff/growstuff/pull/980), though maps are rendering fine on staging.

CloCkWeRX:
Fixes #981 Adds an access token, but from my free account (limited to 50k views, in use elsewhere).


cesy:
We should use our paid account in production - can you update it to use an environment variable, and fall back to your free one only if that fails?

CloCkWeRX:
I opted to skip the ENV approach for now, as we would be rendering that token out as part of the link - its not exactly secret.

PR #1028, opened 2016-08-15
Minor gem upgrades for August 2016

CloCkWeRX:
Devise 4.2 is probably the biggest one here, the rest are minor bits and pieces.

Still no rails 5 :(


cesy:
For the future, we try to avoid merging our own pull requests unless it's critically urgent - so the active record one was fine, but ideally this should have waited for someone else.

Thanks for doing the updates, though.

PR #1032, opened 2016-08-18
Release 14

cesy:
Ready for release.


CloCkWeRX:
Well, those tests seem happy. Staging deploy?

cesy:
Staging already has the latest, since it takes from dev rather than master, so if you can merge this, I'll deploy to prod.

CloCkWeRX:
I can still see a bug fixed in https://github.com/Growstuff/growstuff/pull/1032/commits/58452204aa3ea6523e4465688f49c206b050cf06#diff-37bb8b73ce97a0974da8119c0e40416fR97 on http://www.staging.growstuff.org/crops/anaheim-pepper ; so I don't think its quite at the latest

cesy:
Thanks - it looks like it only took one build that day. I've now updated the way it deploys to use the Github-Heroku automatic integration, and kicked off a fresh deploy. We'll need to remember to keep a close eye on any scripts that need running or database migrations, but we were doing that anyway.

CloCkWeRX:
I'm fairly happy with it after a noodle around on staging.

Checked:
 - Layout on Galaxy s6, search bar, etc - works well
 - Ebay/Google/Gardenate links for Lettuce, Anaheim Peppers. It works fairly well, though the Gardenate 404 page is a bit annoying :(
 - Member page layout on small devices. The 'location not set' imagery and layout is a bit funny still on mobile, but not a showstopper.
 - Signin/sign out
 - Imagery of crops/plantings (very briefly)
 - Mark as finished labelling appears OK

cesy:
Great, and deployed!

PR #1034, opened 2016-08-22
Rails update

cesy:
We'll work our way up to 4.2.7 to check it's working at each stage, before going to 5.0, by which time 5.0 will be more stable.


CloCkWeRX:
It might be worth just leaping to '~> 4.2.7' ; very little seemed to go wrong with similar upgrades in a number of other.projects

PR #1036, opened 2016-09-11
allow registration and signin/out on the API

Br3nda:
I wanted to sign in/out over the json API. I told devise to allow repsonding in json in both registration and session controller.

There might be another way to do this already, but the API docs link in the readme is all 404-ed pages.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -71,3 +71,4 @@ submit the change with your pull request.
 - Lucas Nogueira / [lucasnogueira](https://github.com/lucasnogueira)
 - Charley Lewittes / [ctlewitt](https://github.com/ctlewitt)
 - Kristine Nicole Polvoriza / [polveenomials](https://github.com/polveenomials)
+- Brenda Walalce / [br3nda](http://github.com/br3nda)
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,4 +1,5 @@
 class RegistrationsController < Devise::RegistrationsController
+  respond_to :json
 
   def edit
     @twitter_auth  = current_member.auth('twitter')
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,4 +1,6 @@
 class SessionsController < Devise::SessionsController
+  respond_to :json
+
   def create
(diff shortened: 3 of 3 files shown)


CloCkWeRX:
Out of curiosity, what are you doing that wants registration/login/logout? 

I'm not 100% these being opened up, as it makes it easier to trawl through logins or create spam accounts (not that we've added captcha or similar on the webui, but that's not too tricky to do). 

Wouldn't be as fussed if we had API tokens that could be revoked; so that external apps could act on behalf of another user.



The tests are happy, its only:
```
The command "bundle exec script/check_contributors_md" exited with 1.
```
which is easily fixed by adding your name to Contributors.md

Br3nda:
Started a little android app. 
https://github.com/Br3nda/RobotGarden

I started by posting to the html signin (without .json suffix) which works but doesn't let me work out which user they logged in as. So, if someone wants to trawl, they can do that now to the html response.

I need the json response to tell me their login_name, so i can next request their gardens (the api seems to require i know this, i can't just request my own). 

Are the docs somewhere? All i found was 404s where api docs used to be.

Br3nda:
oh cool - i see where the docs are now.
https://github.com/Growstuff/growstuff/commit/3ea8e2294eb2b701eb9c452e3085b889e0f1f093

cesy:
I'd much prefer proper API tokens, too.

Br3nda:
Ideally the whole api would adhere to jsonapi.org. Then i wouldn't need as much custom code to interact with the api.

Is my patch being rejected in favour of instead a new json api method that returns a token on sign in?

CloCkWeRX:
> Ideally the whole api would adhere to jsonapi.org.

I think thats a quite workable idea for us to move towards, we can mix in json-ld as well if we want to.

CloCkWeRX:
> Is my patch being rejected in favour of instead a new json api method that returns a token on sign in?

Lets not say no to it outright, we can always merge then refactor.

CloCkWeRX:
So, I've started stuff in #1040 for the next version of the API, which we can have output json-api format data and add in tokens and what have you. Feel free to throw suggestions galore onto that pull request; I'll knock 'em off piece by piece

This might take a bit to get deployed to staging.growstuff.org ; but it should go up soon.

PR #1040, opened 2016-09-14
Add api v1: Documentation, namespace, routes, etc

CloCkWeRX:
Docs viewable at http://127.0.0.1:3000/api/

This is really just copy/paste of the controllers listed on https://github.com/Growstuff/growstuff/wiki/API-Version-0 ; with respond_to assuming JSON only.

No tests etc yet, just wanted to put it up for discussion. Will DRY the more complex stuff later.

Remaining:
 * Introduce authentications controller as per discussion on #1036
 * Introduce API tokens
 * Adjust output to be json-api style


cesy:
Is it best practice to have separate controllers for the API now, rather than reusing the ones for the main site? I would have thought we wanted the logic just in one place, so if we decide to change who's allowed to do what, we don't have to remember to update the API separately.

CloCkWeRX:
So the main reason for it is that the API concerns and the site concerns are going to be different over time. Right now, they are very similar as I've just duplicated a bunch of code but removed HTML view specific concerns (/new, /edit, etc).

However, if we start presenting content as jsonapi (or jsonld or geojson) it becomes a very different presentation of the underlying resource. 
Rails doesn't really have view-models, which would probably solve this in other frameworks - there are things like active model serializers or rabl, but its all a bit piecemeal.

You can see a bit of that where some of the HTML controllers and the current .json output introduce all sorts of extra lookups for the html presentation only - stuff like ```@title``` variables; while the API version has a much more interpretation of things.

Alternatively, look at our RSS and CSV respond_to behaviours; or places we use to_json and more or less describe a 'template' - it would make sense to go turn those into a ViewModel/Renderer or similar style class that takes data and spits out a resource.

We also end up with questions like "How do we completely change the modelling of resource X without breaking existing integrations?" (versioning helps, thus the routes), "Can I measure traffic specifically to the API?" (specific routes, tokens), "Can I find an abusive API consumer easily?" (Application token), "Do I allow different security behaviour that what I would for the UI; ie CORS?"

> if we decide to change who's allowed to do what, we don't have to remember to update the API separately.

You are dead right, the next logical step is to DRY.
 
If there's anything that makes send to be stuck in the model, we should do that as much as possible. 

For the things that don't make sense - ie, updating a whole lot of related models then firing off an email when a user signs up - we should introduce service objects (https://blog.engineyard.com/2014/keeping-your-rails-controllers-dry-with-services).

I held off doing that kind of refactoring as it doesn't tend to make sense until you have the same behaviour in 2-3 places, each with their own subtle tweaks - say a rake task that runs periodically (so handles command line args a bit), stuff that happens through the UI (with flash messages mixed in) and API calls.

That really lets you pull out the most relevant application/domain behaviour (planting and growing crops!) from the stuff describing how it interfaces with the world.

cesy:
Great, that makes sense. Do we want to merge this as-is or wait for some of the other bits?

CloCkWeRX:
Let's maybe merge as is, that way the security token parts of it are easier to review/we can find gaps in the design with @Br3nda on the staging site

cesy:
We can ignore code climate complaining about non-DRY stuff, but can you check through the other errors it's thrown, please?

Br3nda:
This gem is useful for speeding up api dev and ensuring you're adhering to specification
https://rubygems.org/gems/jsonapi-resources

Br3nda:
I've had a play, and while this is an api that does json, it's not json api.

I've send a PR to this PR :-) 
over at https://github.com/CloCkWeRX/growstuff/pull/1

Br3nda:
I think we need tokens, and some request specs next.

Br3nda:
i added some request specs here:
https://github.com/CloCkWeRX/growstuff/pull/2

How far should we spike out on this feature? Permissions and tokens are a big piece

CloCkWeRX:
@Br3nda did you want to just put your PR in directly, there's a few conflicts at the moment on my branch. I might close this off in favour of yours.

PR #1045, opened 2016-09-18
harvest thumbnail title linking to the harvest itself

Br3nda:
I found it confusing that "bob's harvest" doesn't link to Bob's harvest. It instead links to Bob. I've made this mistake many times trying to get a the harvest#show from harvest#index.

This patch makes that link lead to the harvest itself.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -71,4 +71,4 @@ submit the change with your pull request.
 - Lucas Nogueira / [lucasnogueira](https://github.com/lucasnogueira)
 - Charley Lewittes / [ctlewitt](https://github.com/ctlewitt)
 - Kristine Nicole Polvoriza / [polveenomials](https://github.com/polveenomials)
-- Brenda Walalce / [br3nda](http://github.com/br3nda)
+- Brenda Wallace / [br3nda](http://github.com/br3nda)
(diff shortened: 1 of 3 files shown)


Br3nda:
example page where this is found:
http://www.growstuff.org/harvests
I cannot see how to get to harvest#show from there.

cesy:
Good catch, this is definitely worth fixing. Can you add a test for it, though, please?

cesy:
Ah, and it's failing because your name isn't yet in the Contributors file - please can you fix that, and then I'll merge this one while we're discussing the API? Thanks.

pozorvlak:
@Shiny, @Cesy: how do you feel about us testing CONTRIBUTORS.md membership
in CI? Helpful, or annoying?

An alternative (which would be more work), would be to have a bot that
automatically opened PRs for new contributors, but not everyone's name is
easily discoverable from their GitHub profile.

cesy:
Helpful in CI, for me.

cesy:
These poltergeist failures are really annoying. I've restarted it again.

Br3nda:
I don't know why the contributors was failing. @br3nda was already in there and passes on my checkout.

I don't mind the small failures. The poltergeist test is looking for something on port 8081, and i don't know what that is. Maybe the tests need better names, to assist when they fail.

CloCkWeRX:
The network errors are an unfortunate feature of our CI environment, and it's really not clear why. I just whacked the restart build button, which should give us a clean test run hopefully.

Br3nda:
Everything is passing except the funky poltergeist on CI which isn't involved in this change.
Can we merge?

PR #1048, opened 2016-10-07
Prevent FactoryGirl from trying to connect to ElasticSearch in tests

### LINKED ISSUE #1047: Failed to open TCP connection to localhost:9200
when i run the specs i get 100+ tests failing with  Failed to open TCP connection to localhost:9200
Is this for elastic search? the comments in application.yml.example lead me to thing this isn't needed to run the app or the tests.

I have config/application.yml, with GROWSTUFF_ELASTICSEARCH: "false"

Is it supposed to be possible to run the specs without elastic search?
Are there instructions somewhere on setting up elastic search for dev,test,ci?

---
p.s. typed with a small child pulling on my arm. forgive typoss.

pozorvlak:
Fixes #1047.

### DIFF
--- a/app/models/crop.rb
+++ b/app/models/crop.rb
@@ -50,44 +50,46 @@ class Crop < ActiveRecord::Base
 
   ####################################
   # Elastic search configuration
-  include Elasticsearch::Model
-  include Elasticsearch::Model::Callbacks
-  # In order to avoid clashing between different environments,
-  # use Rails.env as a part of index name (eg. development_growstuff)
-  index_name [Rails.env, "growstuff"].join('_')
-  settings index: { number_of_shards: 1 },
-    analysis: {
-      tokenizer: {
-        gs_edgeNGram_tokenizer: {
-          type: "edgeNGram",  # edgeNGram: NGram match from the start of a token
-          min_gram: 3,
-          max_gram: 10,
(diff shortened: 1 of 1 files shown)


cesy:
Have you set the new env variable in Heroku stage and prod? Do we need to document it anywhere else, e.g. putting it to false in our example .yml so other people can use our code without getting caught out by it?

Travis failures were just Poltergeist, so this is otherwise good to go.

pozorvlak:
I didn't add an environment variable, I just added an additional check for
one we were already using.

CloCkWeRX:
Oophf, I get why it works, I just get sad that we have hooks like this :( 

Longer term, are we better off explicitly calling update_index from a controller; rather than relying on after_* ?

In my experience, this initially sounds like a bad idea but ends up being only a handful of places where we are being explicit in an controller or two - and that tends to be a better spot to look at config/env vars to change behaviour (wiring)

Maybe if we get bitten a few more times, worth another look

pozorvlak:
Agreed, this is pretty ugly, and your suggestion is definitely worth
bearing in mind for the future.

Honestly, though, I think the best solution is "make setting up an
environment containing ElasticSearch trivially easy, and then remove all
the non-ES code".

PR #1049, opened 2016-10-11
moving where clauses into scopes to keep that contained in models

maco:
just pushing things down to the model


app/controllers/notifications_controller.rb
```diff
@@ -5,7 +5,7 @@ class NotificationsController < ApplicationController
 
   # GET /notifications
   def index
-    @notifications = Notification.where(recipient_id: current_member).page(params[:page])
+    @notifications = Notification.find_by_recipient(current_member).page(params[:page])
```
CloCkWeRX: Might be better with the 'by_recipient' instead of find_by_recipient, as it looks like the normal find_by method that would only return one record; not a .where

app/models/notification.rb
```diff
@@ -11,6 +11,10 @@ class Notification < ActiveRecord::Base
   before_create :replace_blank_subject
   after_create :send_email
 
+  def self.find_by_recipient(recipient)
```
CloCkWeRX: Is it better to use the scope syntax for these?

```
scope :by_recipient, ->(recipient) { where(recipient_id: recipient) }
```

CloCkWeRX:
Maybe some other candidates:

We do similar to this in a few places
```
@crops = Crop.popular.includes(:scientific_names, {plantings: :photos})
```

A few variants of this are around:
```
@owner.harvests.includes(:owner, :crop)
```

And a few of these

```
@crop.plantings.includes(:owner, :crop, :garden)
```

maco:
At my day job they're standardized on only using scope syntax for single
line with no parameters. Rubocop doesn't have an opinion. I'm fine with
using scope syntax, but figure we should talk about what's our preference
for readability.

Thanks for pointing out those other ones. I just did a search for where.
Hmm, I should check for order too.

As to why move even simple ones to the model, just trying to see where the
model can do the work instead of the controller. My Rails knowledge is very
ad hoc, but I'm trying to get this fat model, thin controller thing
internalized and figure poke GS while I do. (Also I'm currently reading
Rails Antipatterns)

CloCkWeRX:
> As to why move even simple ones to the model, just trying to see where the
model can do the work instead of the controller.

Oh yeah, I'm not saying never do it; particularly if its something used repeatedly or more concise naming makes more sense :)

> fat model, thin controller

That one is a bit of a balancing act - http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/ is a worthwhile followup read; which gives you a bit more of a vocabulary to solve things like "update a whole lot of related database entries; update elasticsearch and send an email at the end" without getting stuck trying to figure out which model should own what.

Br3nda:
dev branch now has rubocop checking, so it'd be best to merge dev into here, and trigger a travis run.

cesy:
@maco when you can update this, it would be great to merge

app/controllers/orders_controller.rb
```diff
@@ -4,7 +4,7 @@ class OrdersController < ApplicationController
 
   # GET /orders
   def index
-    @orders = Order.where(member_id: current_member.id)
+    @orders = Order.by_member_id(current_member.id)
```
pozorvlak: I'd prefer `by_member(current_member)`.

app/models/order.rb
```diff
@@ -12,6 +12,10 @@ class Order < ActiveRecord::Base
 
   before_save :standardize_referral_code
 
+  def self.by_member_id(member_id)
```
pozorvlak: Scope syntax here as well, I think.

pozorvlak:
I've requested a couple of changes, but they're both very minor and I'm not too bothered about them if you want to talk me out of them :-)

maco:
oh uh whoops, I'd also made another change...I made a private function called `popular_crops` to make it clearer what's happening in index so the comments become unneeded. And this is where my question about why our `.includes()` vary by format comes in.

CloCkWeRX:
That dismiss button was probably the wrong one, oh well.

PR #1055, opened 2016-11-12
Add rubocop to travis ci

Br3nda:
this does three things

1) include rubocop in the bundle
  Rubocop adds new offenses in each version, and sometimes removes them. The rubocop config file changes between rubocop versions too. We need to track which version this project should be checked against, and manage the upgrades

2) Add all existing rubocop offenses to a todo list
  This is the output of `rubocop --auto-gen-config`, and can be updated as the code base style gets more consistent

3) Add rubocop checking to the travis script.

We no longer need to disable all check then enable those we have fixed against. Instead we have a todo list of where we need to work - and can then fix these one check at a time, or one file at a time.


Br3nda:
Code climate is failing because it found the word "todo"
I think this is appropriate word for this,  (and also the standard filename for a rubocop autoconfig file)

Br3nda:
Contributor check is failing. I've fixed in #1056

.rubocop.yml
```diff
@@ -1,1156 +1,9 @@
...
-Metrics/LineLength:
-  Description: 'Limit lines to 80 characters.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#80-character-limits'
-  Enabled: false
-
-Metrics/MethodLength:
-  Description: 'Avoid methods longer than 30 lines of code.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#short-methods'
-  Enabled: true
-  Max: 30
```
maco: This is a custom number (it defaults to 10). Can you preserve that?

.rubocop.yml
```diff
@@ -1,1156 +1,9 @@
...
-
-Style/SpecialGlobalVars:
-  Description: 'Avoid Perl-style global variables.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-cryptic-perlisms'
-  Enabled: true
-
-Style/StringLiterals:
-  Description: 'Checks if uses of quotes match the configured preference.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#consistent-string-literals'
-  Enabled: true
```
maco: I'm not sure what "match the configured preference" works out to, but in discussion we decided that since there's no performance hit for unnecessarily using double-quotes, we don't care about single versus double-quotes. Can you make sure it's not being overly strict in the autogenerated config?

.rubocop.yml
Br3nda: I've changed to to have a comment saying it should be set to 30 once the long lines are addressed in the code.

PR #1056, opened 2016-11-12
Contributors check case-insentive

Br3nda:
My PRs are being failed by travis, because it thinks my username is 'Br3nda', and the contributors file says 'br3nda'

This makes the grep for contributors ignore case.

### DIFF
--- a/script/check_contributors_md
+++ b/script/check_contributors_md
@@ -28,7 +28,7 @@ Please set it using
   end
 end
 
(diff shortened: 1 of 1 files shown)


Br3nda:
Also noticed we were allowing github user names unescaped into shell. Small risk, but patched for best practise since i was there anyways.

pozorvlak:
Good catch, thanks!

PR #1058, opened 2016-11-14
Removed duplicate wiki_linktext

Br3nda:
same key appears again (with different value) 5 lines below


Br3nda:
```     Capybara::Poltergeist::StatusFailError:
       Request to 'http://localhost:8081/forums/permaculture' failed to reach server, check DNS and/or server status
```
I don't think the i8n file did that

PR #1061, opened 2016-11-17
Wrapped all lines > 120 chars.

Br3nda:


Br3nda:
"script/gemfile_check
Looks like you committed changes to Gemfile but not Gemfile.lock
The command "script/gemfile_check" exited with 1."

The changes are whitespace for rubocop compliance

app/controllers/crops_controller.rb
```diff
@@ -215,6 +215,15 @@ def destroy
   private
 
   def crop_params
-    params.require(:crop).permit(:en_wikipedia_url, :name, :parent_id, :creator_id, :approval_status, :request_notes, :reason_for_rejection, :rejection_notes, scientific_names_attributes: [:scientific_name, :_destroy, :id])
+    params.require(:crop).permit(:en_wikipedia_url,
```
pozorvlak: I really hate this indentation style, and would much prefer

```
params.require(:crop).permit(
  :en_wikipedia_url,
  :name,
  # etc
)
```
But I'm prepared to be out-voted. What do other people think?

spec/models/crop_spec.rb
```diff
@@ -597,8 +597,17 @@
   end
 
   context "crop rejections" do
-    let!(:rejected_reason) { FactoryGirl.create(:crop, name: 'tomato', approval_status: 'rejected', reason_for_rejection: 'not edible') }
-    let!(:rejected_other) { FactoryGirl.create(:crop, name: 'tomato', approval_status: 'rejected', reason_for_rejection: 'other', rejection_notes: 'blah blah blah') }
+    let!(:rejected_reason) do
+      FactoryGirl.create(:crop, name: 'tomato',
```
pozorvlak: Again, I'd prefer an indentation style that didn't introduce vast acres of horizontal whitespace, but I see what you were going for here. Interesting.

pozorvlak:
The Travis CI failures are just the Gemfile check, as @Br3nda says, so this is good to merge. It looks like the `--ignore-all-space` and `--name-only` options to `git diff` don't play nicely together, so it's not straightforward to fix `script/gemfile_check` to handle this case. I vote we don't bother; this seems like a pretty rare situation.

app/controllers/crops_controller.rb
maco: I don't have a strong opinion either way, but it seems like the way @Br3nda did it here is more likely to result in nested things hitting the line length limit anyway. I've certainly hit the 80 char limit in Python doing it this way before. (And I do have the impression the say it's done here is what's common in those languages, so I can only guess the people who wrote PEP-8 and whatever Ruby's version is are shaking their fists and saying "don't nest so deep!" at me)

spec/models/crop_spec.rb
Br3nda: i agree on less space - but enforcing this indentation style also makes it clearer that's it's one level - that we're not passing arrays of hashes of arrays. I think we'll catch some existing bugs as we make the contents of long lines come in to view.

maco:
I don't think you'd lose that clarity by indenting 2 spaces then 4 spaces
etc (like coffee script json hashes)

Br3nda:
@maco I think that's Style/AlignParameters
SupportedStyles: with_first_parameter, with_fixed_indentation

It would look like this
```
578	+	      FactoryGirl.create(:crop,
579	+	        name: 'tomato',
580	+	        approval_status: 'rejected',
581	+	        reason_for_rejection: 'not edible')
```

Br3nda:
I personally find it easier to read like this:
```
	584	+	      FactoryGirl.create(:crop,
585	+	                         name: 'tomato',
586	+	                         approval_status: 'rejected',
587	+	                         reason_for_rejection: 'other',
588	+	                         rejection_notes: 'blah blah blah')
```

which is the rubocop default (a quasi standard of it's own).

Br3nda:
:-D I gotta go dig up some spuds. 

Lemme know if i should set with_first_parameter (rubocop default) or with_fixed_indentation . I can probably sort that today.

Br3nda:
I'm trying to keep the PRs to one subject, so that's mere humans can feasibly review (instead of all the codestyle fixes at once)

The next PR i send can be a fix for the inconsistent indentation in the code base. 

Is that a good compromise to get this one merged?

maco:
Sure

Br3nda:
Can I get this merged so on go onto indentation PR next?

pozorvlak:
I'd like to do release 15 first - will try to get that out today.

pozorvlak:
@Br3nda Release 15 is now out, so I'm keen to start merging all your PRs! Can you please rebase or merge this with dev to fix the conflict?

Br3nda:
Request to 'http://localhost:8081/members/member49' failed to reach server, check DNS and/or server status

app/controllers/crops_controller.rb
```diff
@@ -215,6 +215,15 @@ def destroy
...
+                                 :name,
+                                 :parent_id,
+                                 :creator_id,
+                                 :approval_status,
+                                 :request_notes,
+                                 :reason_for_rejection,
+                                 :rejection_notes, scientific_names_attributes: [:scientific_name,
```
pozorvlak: Regardless of what we decide about indentation style, I think you probably want a newline before `scientific_names_attributes` here. Depending on the outcome of the indentation style debate, you may or may not need one after `[`.

pozorvlak:
So we're merging this and then bikeshedding over indentation style in a separate PR? Works for me.

pozorvlak:
My vote's for `with_fixed_indentation` - I think that should make nesting level clear (hashes of arrays of hashes of etc get another level of indentation, right?) but not introduce too much extra horizontal whitespace. As I said, though, I'm happy to be outvoted.

PR #1062, opened 2016-11-19
Reduce complexity in photos controller

Br3nda:
according to rubocop this method had the highest perceived complexity 
Moved some to another method.


Br3nda:
```
 1) account types admin user deleting account type
     Failure/Error: visit account_type_path account_type
     Capybara::Poltergeist::StatusFailError:
       Request to 'http://localhost:8081/account_types/2' failed to reach server, check DNS and/or server status
     # ./spec/features/admin/account_types_spec.rb:52:in `block (3 levels) in <top (required)>'
```

*sigh*.. again.

app/controllers/photos_controller.rb
```diff
@@ -129,4 +115,23 @@ def photo_params
...
+  def find_or_create_from_flickr_photo(flickr_photo_id)
+    photo = Photo.find_by(flickr_photo_id: flickr_photo_id)
+    photo = Photo.new(photo_params) unless photo
+    photo.owner_id = current_member.id
+    photo.set_flickr_metadata
+    photo
+  end
+
+  def which_collection?
+    case params[:type]
```
CloCkWeRX: Should really swap this to something like
```
return unless %w(garden planting harvest).index(planting[:type])

@photo.send(planting[:type].plualize.to_sym)
```

Br3nda:
code climate is complaining of code duplication in growstuff.rake. 
So i fixed that too...

app/controllers/photos_controller.rb
Br3nda: :+1: will do

Br3nda: neat! same list is used when finding item later.

Br3nda:
Brakeman is unhappy with our 2 liners, and brakeman is right.
I'll make it the more verbose, but stronger white-listed way.

PR #1063, opened 2016-11-20
Use phantomjs 211 in travis

Br3nda:
This is pretty much the same as #979. I forked, and merged in dev branch, and got Green build reliably on travis.
Hoping for the same here.


Br3nda:
It's failing script/gemfile_check
This is because the commit that fixes the codestyle makes that script unhappy when merged in here.

PR #1065, opened 2016-11-23
Reduced code duplication in growstuff.rake

Br3nda:
codeship was failing un-related PRs because of this duplication.

### DIFF
--- a/lib/tasks/growstuff.rake
+++ b/lib/tasks/growstuff.rake
@@ -4,18 +4,20 @@ namespace :growstuff do
   # usage: rake growstuff:admin_user name=skud
 
   task admin_user: :environment do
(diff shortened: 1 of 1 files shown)


lib/tasks/growstuff.rake
```diff
@@ -4,18 +4,20 @@ namespace :growstuff do
...
-    admin  = Role.find('admin')
-    member.roles << admin
+    add_role_to_member! ENV['name'], 'admin'
   end
 
   desc "Add a crop wrangler user, by name"
   # usage: rake growstuff:cropwrangler_user name=skud
 
   task cropwrangler_user: :environment do
-    member = Member.find_by_login_name(ENV['name']) or raise "Usage: rake growstuff:cropwrangler_user name=whoever (login name is case-sensitive)"
```
CloCkWeRX: Any change you can retain the raise 'how to use this' messaging or shift it into the desc?

Br3nda: :+1:

PR #1066, opened 2016-11-24
Auto-deploy to production on successful merge

pozorvlak:
I *think* this is all that's necessary, though it's a bit hard to test :-)


cesy:
Let's give it a go!

PR #1075, opened 2016-11-26
Bugfix: Link to garden, not member

Br3nda:
"[Member]'s garden" should link to the garden, not the member

image

### DIFF
--- a/app/views/gardens/_thumbnail.html.haml
+++ b/app/views/gardens/_thumbnail.html.haml
@@ -1,7 +1,7 @@
 .panel.panel-success
   .panel-heading
     %h3.panel-title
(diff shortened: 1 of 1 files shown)


app/views/gardens/_thumbnail.html.haml
```diff
@@ -1,7 +1,7 @@
 .panel.panel-success
   .panel-heading
     %h3.panel-title
-      = link_to "#{garden.owner.login_name}'s garden", garden.owner
+      = link_to "#{garden.owner.login_name}'s garden", garden
```
CloCkWeRX: I tackled something similar with 80f826421bdd72e67182a193e07519bdb0cd83e8 I think - could swear there was a bug report open for it, but can't find it.

PR #1077, opened 2016-11-27
Narrow-ed regex on wikipedia url

Br3nda:
the regexp was missing a \z t end it, so allowed anything you want afterwards. 
These are handled safely afterwards, but the regex still needed a fix.


Br3nda:
codeclimate likes me. it says "Fixed: Insufficient validation for 'en_wikipedia_url' using /\Ahttps?:\/\/en.wikipedia.org\/wiki/. Use \A and \z as anchors"

PR #1078, opened 2016-11-27
Limit garden names to one line

Br3nda:


Br3nda:
code climate likes this too: "Fixed: Insufficient validation for 'name' using /\S/. Use \A and \z as anchors"

maco:
Do you have the option to press the restart button in failing tests when
you view then in Travis? (Login with Github)

Br3nda:
@maco No, the option seems only appear in Travis if you have write/commit access to the repo,. and I don't.

PR #1083, opened 2016-11-28
Improve crop model

maco:


app/models/crop.rb
```diff
@@ -354,6 +335,23 @@ def self.search(query)
     end
   end
 
+  private
+
+  def add_names_to_list(names_to_add, list_name, col_name)
+    cropbot = Member.find_by_login_name('cropbot')
+    names_to_add.each do |n|
+      if self.send("#{list_name}_names").exists?(["#{col_name} LIKE ?", n])
```
maco: If anyone knows a better way to do this `exists?` check using a string?

CloCkWeRX: It could do with a comment or two to explain it clearly, took me a bit to get my head around the whole method.

Only tweaks I can really think of is stuff like:
```
actual_names_to_add = names_to_add.select {|name| !self.send("#{list_name}_names").exists?(["#{col_name} LIKE ?", name]) }
```
... which kills off one conditional.

Maybe the SQL could be tweaked to be ```SELECT DISTINCT column_name FROM list_name WHERE LOWER(column_name) IN(?)``` and pass in lowercase input; so you do one query for all existing instead of multiple.
Then just ```names_to_add - existing_names_query```

CloCkWeRX:
Any chance of doing a new branch and just cherry picking the relevant commits to clean up history?

```
git checkout dev
git checkout -b fix-model-two-electric-boogaloo
git cherry-pick 864fc9815e674fc6fac5b9ce5608fc6ff6528447
```

app/models/crop.rb
maco: If alternate_names and scientific_names both just called their column `name` this'd be SO much easier. 

The reason I'm wondering about a better way is that Code Climate's warning about SQL injection. Now, I don't think injection is *actually* an issue here, since this is in a private function in the model operating on a local csv file, but...figured I'd check.

CloCkWeRX: Not sure you can avoid that without creating a view or similar (```SELECT foo as bar FROM blah```) - that quickly is going to upset activerecord for an edge case.

You *might* be able to use ```alias_attribute :name :scientific_name``` and swap it for ```.where(name: ...)```

PR #1084, opened 2016-11-29
Improve crop model

maco:
hey @CloCkWeRX I made it more readable. It'll still claim to have potential for SQL injection, which I still think it's wrong about (because private method operating on local CSV file), so maybe can tell Code Climate to ignore...

### DIFF
--- a/app/models/crop.rb
+++ b/app/models/crop.rb
@@ -263,43 +263,24 @@ class Crop < ActiveRecord::Base
       logger.warn("Warning: no scientific name (not even on parent crop) for #{self}")
     end
 
+    cropbot = Member.find_by_login_name('cropbot')
+
     if names_to_add.size > 0
-      cropbot = Member.find_by_login_name('cropbot')
       raise "cropbot account not found: run rake db:seed" unless cropbot
 
-      names_to_add.each do |n|
-        if self.scientific_names.exists?(scientific_name: n)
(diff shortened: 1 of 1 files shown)


CloCkWeRX:
Might even make things a bit less flakey with the if elasticsearch logic :)

maco:
Wait, I'm confused about that elasticsearch stuff. I did the cherrypick thing you said, but I don' know what happened there...? It looks like maybe some of @Br3nda 's changes got screwy when I did the cherry picking? (Or maybe when I committed a change, uncommitted it, and added more stuff to it....) I was meaning to only touch below line 340 and the `add_*_names_from_csv` functions

CloCkWeRX:
TBH, that change is worthwhile enough I think to just leave in :)

ie; we run two sets of tests and one network call tends to fail out of the two sets. We've reduced that by ~50% now most likely.

maco:
Oh it looks like when you did the merge, that all came out in the wash, so...weird merge weirdness is all

maco:
And hey, crop.rb's grade moved from D to C 😛

pozorvlak:
Yeah, looks like the `if ENV['GROWSTUFF_ELASTICSEARCH']` bit came from ce2141dd, and must have been merged in to your branch somehow. It's definitely worth keeping, not least because that commit was a fix for #1047 :-)

pozorvlak:
Also, \o/ for improving grades!

Issue #1085, opened 2016-11-29
Fix "Similar code" issue in app/models/crop.rb


maco:
Similar code found in 1 other location (mass = 23)

https://codeclimate.com/github/Growstuff/growstuff/app/models/crop.rb#issue_583ccb95fe9fa300010001dd

PR #1086, opened 2016-11-29
deduplicate sunniness and planted_from functions (Fixes: #1085)

maco:


app/models/crop.rb
```diff
@@ -152,27 +152,15 @@ def default_photo
...
-    end
-    return sunniness
+    count_uses_of_property 'sunniness'
   end
 
   # crop.planted_from
   # returns a hash of propagation methods (seed, seedling, etc),
   # key: propagation method (eg. 'seed')
   # value: count of how many times it's been used by plantings
   def planted_from
```
CloCkWeRX: Where are these count methods called from?

They may be better suited to a ```SELECT COUNT(0), planted_from FROM plantings WHERE crop_id =?``` or similar

maco:
screen shot 2016-11-28 at 8 41 02 pm

PR #1088, opened 2016-11-29
rename scientific_names.scientific_name to scientific_names.name

maco:
this mirrors alternate_names.name and allows making code more DRY


app/models/crop.rb
```diff
@@ -139,11 +137,11 @@ def default_scientific_name
   # later we can choose a default photo based on different criteria,
   # eg. popularity
   def default_photo
-    return photos.first if photos.any?
```
CloCkWeRX: We want this return here, otherwise it'll keep going to the next line

db/schema.rb
```diff
@@ -433,15 +433,15 @@
...
-  create_table "scientific_names", force: true do |t|
-    t.string   "scientific_name", null: false
-    t.integer  "crop_id",         null: false
+  create_table "scientific_names", force: :cascade do |t|
+    t.string   "name",       null: false
+    t.integer  "crop_id",    null: false
     t.datetime "created_at"
     t.datetime "updated_at"
     t.integer  "creator_id"
   end
```
CloCkWeRX: Its interesting the lack of indexes on crop_id, etc in the schema. Should fix that in another ticket probably

app/controllers/crops_controller.rb
```diff
@@ -169,19 +169,15 @@ def update
...
           params[:alt_name].each do |index, value|
-            alt_name = @crop.alternate_names.create(name: value, creator_id: current_member.id)
-          end
-
-          @crop.scientific_names.each do |sci_name|
-            sci_name.destroy
+            create_name('alternate', value)
           end
+        end
+        if !params[:sci_name].nil?
```
Br3nda: Could `unless`be better than `if!`

pozorvlak:
@Br3nda: you're right, the "what the reviewers think" summary view is really useful!

maco:
Since @Br3nda says 👍 and I fixed the things @CloCkWeRX mentioned, I'm going to merge it.

Issue #1079, opened 2016-11-27
My growstuff.org profile is all http 500 errors


Br3nda:
I added a new planting (lavender).
After submitting the form i get a http 500 error at http://www.growstuff.org/plantings/shiny-beside-the-garage-lavender

If i go to my plantings page, i also get a http 500 error.
http://www.growstuff.org/plantings/owner/shiny

Same error page on my profile:
http://www.growstuff.org/members/shiny

This happens when signed in, and when not signed in.

![image](https://cloud.githubusercontent.com/assets/12266/20646136/f6f40278-b4d9-11e6-9dd8-ad6b8fce2a5a.png)

Other parts of the site seem fine.

Br3nda:
The garden page is also broken now http://www.growstuff.org/gardens/shiny-beside-the-garage

![image](https://cloud.githubusercontent.com/assets/12266/20646702/5efddd20-b4e5-11e6-9f8b-d2841f9720b6.png)

CloCkWeRX:
Can anyone throw me access to newrelic, easier to hunt down errors like this via that.

pozorvlak:
Logs from hitting http://growstuff.org/plantings/shiny-beside-the-garage-lavender:

```
2016-11-29T09:27:24.695957+00:00 app[web.1]: Started GET "/plantings/shiny-beside-the-garage-lavender" for 213.175.37.10 at 2016-11-29 09:27:24 +0000
2016-11-29T09:27:24.701516+00:00 app[web.1]: Processing by PlantingsController#show as HTML
2016-11-29T09:27:24.701568+00:00 app[web.1]:   Parameters: {"id"=>"shiny-beside-the-garage-lavender"}
2016-11-29T09:27:24.798810+00:00 app[web.1]:   Rendered plantings/show.html.haml within layouts/application (32.3ms)
2016-11-29T09:27:24.799200+00:00 app[web.1]: Completed 500 Internal Server Error in 97ms (ActiveRecord: 20.8ms)
2016-11-29T09:27:24.804171+00:00 app[web.1]: ActionView::Template::Error (undefined method `+' for nil:NilClass):
2016-11-29T09:27:24.804169+00:00 app[web.1]: 
2016-11-29T09:27:24.804172+00:00 app[web.1]:     43:         = " (#{sunniness})"
2016-11-29T09:27:24.804175+00:00 app[web.1]:     46:       %dd= "#{display_days_before_maturity(@planting)}"
[...]

PR #1091, opened 2016-11-29
Fix PlantingHelper.display_days_before_maturity

pozorvlak:
- fixes the bug that caused #1079
 - fixes an off-by-one error in the calculation
 - always return a string from the helper.

I'm not totally sold on 589a31d; it's possible I've spent too much time coding in statically-typed languages recently and that this isn't an issue in Ruby. But it tripped me up while working on the other two commits.

### DIFF
--- a/app/helpers/plantings_helper.rb
+++ b/app/helpers/plantings_helper.rb
@@ -3,11 +3,11 @@ module PlantingsHelper
     if planting.finished?
       0
     elsif !planting.finished_at.nil?
-      ((p = planting.finished_at - DateTime.now).to_i) <= 0 ? 0 : p.to_i
+      ((p = planting.finished_at - Date.current).to_i) <= 0 ? 0 : p.to_i
     elsif planting.planted_at.nil? || planting.days_before_maturity.nil?
       "unknown"
     else
-      ((p = (planting.planted_at + planting.days_before_maturity) - DateTime.now).to_i <= 0) ? 0 : p.to_i
+      ((p = (planting.planted_at + planting.days_before_maturity) - Date.current).to_i <= 0) ? 0 : p.to_i
     end
   end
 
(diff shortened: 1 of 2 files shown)


config/database.yml
```diff
@@ -1,10 +1,16 @@
...
+  host: localhost
+  user: growstuff
+  password: growstuff
 
 test:
   adapter: postgresql
   database: growstuff_test
+  host: localhost
+  user: growstuff
+  password: growstuff
```
maco: looks like these changes broke it, maybe?

pozorvlak: Gah! Will fix.

app/helpers/plantings_helper.rb
```diff
@@ -1,13 +1,15 @@
 module PlantingsHelper
   def display_days_before_maturity(planting)
     if planting.finished?
-      0
+      "0"
     elsif !planting.finished_at.nil?
-      ((p = planting.finished_at - DateTime.now).to_i) <= 0 ? 0 : p.to_i
+      ((p = planting.finished_at - Date.today).to_i) <= 0 ? "0" : p.to_i.to_s
```
maco: rubocop is grumpy about `Date.today` without a timezone. Perhaps you want `Date.current`?

Br3nda:
sorry i missed this already existed. I did a simpler fix over at #1092
it doesn't fix any other bugs - but i'm hoping it can be deployed hotfix to production and make my growstuff.org account not broken anymore.

app/helpers/plantings_helper.rb
pozorvlak: Done.

config/database.yml
pozorvlak: I've now fixed my local DB setup so I don't need to keep applying and stashing these changes :-)

PR #731, opened 2015-03-16
#578 More extensive database seeding - bulk creating members, gardens and plantings

oshiho3:
Made some improvement on the database seeding process, covering some items from issue #578. Just did this as I wanted to test performance on geo related processing.

This will let you...
- Create N number of members and 
- Populate member's location
- Populate garden's location
- Creates a planting from various crops, sunniness and planted_from for each user.

The default number of members is still 3, but you can specify the number of members as...
<code>rake db:seed member_size=100</code>
or
<code>rake db:reset member_size=100<code>
Tested up to 1000 worked fine.

PR #1094, opened 2016-11-29
Useless assignments fixed, and rails 4 stuff.

Br3nda:
Fixes
* use of dynamic find_by
* before_filter is now before_action in rails 4
* allows two special files to call puts
* fixed up useless assignments


db/seeds.rb
```diff
@@ -87,7 +87,7 @@ def load_test_users
       suburb_file.pos = 0 if suburb_file.eof?
       row = CSV.parse(suburb_file.readline)
 
-      suburb, country, state, latitude, longitude = row[0]
+      suburb, _country, _state, latitude, longitude = row[0]
```
maco: what're the underlines?

app/controllers/seeds_controller.rb
```diff
@@ -49,7 +49,7 @@ def new
     @seed = Seed.new
 
     # using find_by_id here because it returns nil, unlike find
-    @crop = Crop.find_by_id(params[:crop_id]) || Crop.new
+    @crop = Crop.find_by(id: params[:crop_id]) || Crop.new
```
maco: Code Climate is pointing out that alternate_names_controller.rb and ....something else... have practically this same chunk of code in `new()` and I'm left wondering, given existence of `load_and_authorize_resource` and `load_resource` in cancan, do we need to find these by id manually? 

Ok, I'm also wondering why we'd be creating a seed for a crop that doesn't exist.

db/seeds.rb
Br3nda: They indicate that the variable is not going to be used. That's it's okay for linter to ignore this assignment to a useless variable.

app/controllers/seeds_controller.rb
Br3nda: I haven't used cancan much  but i thought load_and_authorize_resource would only load the `@seed` on the seeds controller. The `@crop` and `@owner` needs to be loaded by the method that needs it.

maco:
If I remember right, `load_resource` can be used to get other related
models too. I'll have to check.

Br3nda:
i forgot the `--rails` when i tested this on my laptop.

sorting these last rubocop issues

db/seeds.rb
pozorvlak: @maco: yep, "variable starting with underscore" is a fairly widespread convention for either "private variable" or "variable that can be ignored".

@Br3nda: Good question. This code was merged in #731, and the discussion there doesn't explain where the file `suburbs.csv` came from. It looks like it's a list of Australian suburbs generated for some other purpose and included wholesale - an Australian government database, perhaps? We should probably know more about its origin, to check we're not causing any licensing issues. @oshiho3, where did that file come from?

maco:
Oooh, I don't know about copyright stuff around the world, but in the US
data wouldn't be copyrightable at all (which is why that lawsuit against
the timezone database was bunk).

maco:
I vote ignoring those 3 code climate issues about duplication in controllers, at least for now. We do have a comment somewhere suggesting we look at RABL, and https://github.com/plataformatec/responders is also a possibility when tackling "wow, our controllers are just piles of scaffolding" becomes a high priority.

app/controllers/seeds_controller.rb
CloCkWeRX: Its the
> using find_by_id here because it returns nil, unlike find
bit that makes it different.

spec/features/unsubscribing_spec.rb
```diff
@@ -52,7 +52,7 @@
     # visit /members/unsubscribe/somestring ie.parameter to the URL is a random string
     visit unsubscribe_member_url("type=send_planting_reminder&member_id=#{member.id}")
     expect(page).to have_content "We're sorry, there was an error"
-    updated_member = Member.find(member.id) # reload the member
+    Member.find(member.id) # reload the member
```
CloCkWeRX: That seems unused

spec/helpers/gardens_helper_spec.rb
```diff
@@ -23,7 +23,6 @@
         description: 'a' * 130
       )
       result = helper.display_garden_description(garden)
-      link = link_to("Read more", garden_path(garden))
       expect(result).to eq 'a' * 130
```
CloCkWeRX: Thats a really weird expectation; like something was lost in refactoring

spec/models/garden_spec.rb
```diff
@@ -90,7 +90,7 @@
 
   context 'ordering' do
     it "should be sorted alphabetically" do
-      z = FactoryGirl.create(:garden_z)
+      FactoryGirl.create(:garden_z)
       a = FactoryGirl.create(:garden_a)
```
CloCkWeRX: This one might have been better fixed by fleshing out the other expectation (garden.last == z)

db/seeds.rb
oshiho3: Unfortunately, I can't remember where I pulled the suburbs.csv. The purpose of the file was to improve the speed of generating test users. I remember there was a view that shows a number of growers/gardens in bubbles on a map. I think geolocation was normally pulled from external geocoding service when an address was entered in on a user record. Doing this for all test users was causing long wait.
I hope this information helps.

Br3nda: We could remove that file and instead use the Faker gem.

maco:
Faker is a great gem

app/controllers/seeds_controller.rb
Br3nda: Can you sum up this? Should it be changed back? Is it okay like this?

CloCkWeRX: Its OK as is (maybe adjust comment), it was more why we can't use load_resource or similar there

app/models/member.rb
```diff
@@ -196,6 +196,10 @@ def interesting?
     return false
   end
 
+  def Member.login_name_or_email(login)
```
CloCkWeRX: Later on, should swap that to pure scope syntax

CloCkWeRX:
Still LGTM

app/models/member.rb
Br3nda: I started with scopes - but the railsguide on how to do scopes says a class method is preferrable to a scope with a variable pass in. 
http://guides.rubyonrails.org/active_record_querying.html#passing-in-arguments

> However, this is just duplicating the functionality that would be provided to you by a class method.
> ...
> Using a class method is the preferred way to accept arguments for scopes.

Br3nda:
I'd like someone who understands the growstuff SSO to test this - there are changes in the warden lookup, so someone checking we're still secure is wise.

(I don't use facebook, cos their T&S aren't for me)

CloCkWeRX:
> I'd like someone who understands the growstuff SSO to test this - there are changes in the warden lookup, so someone checking we're still secure is wise.

I must have missed that, unless you mean the before_filter => before_action changes. http://stackoverflow.com/questions/16519828/rails-4-before-filter-vs-before-action covers that

Did you mean something else?

Br3nda:
@clockwerx I mean the query change in find_first_by_auth_conditions

CloCkWeRX:
Ah thats all fine.

PR #1095, opened 2016-11-30
DRY up photo & model interaction

maco:
* move stuff in models into a concern
 * create combined collection for all models that have photos

I'm marking this WIP because I think a constants file that maps garden/gardens/Garden would make it possible to minimize repetition in `photos_controller.rb` too. I want to minimize how many places we need to touch on adding a new model-that-uses-photos because I know missing a copy&paste has introduced bugs before. (The particular past bug I'm thinking of is handled by this commit, though, with the `self.destroy unless relationships.size > 0`)

### DIFF
--- a/app/constants/photo_models.rb
+++ b/app/constants/photo_models.rb
@@ -1,0 +1,37 @@
+module Growstuff
+  module Constants
+    class PhotoModels
+      PLANTING = { type: 'planting', class: 'Planting', relation: 'plantings' }.freeze
+      HARVEST = { type: 'harvest', class: 'Harvest', relation: 'harvests' }.freeze
+      GARDEN = { type: 'garden', class: 'Garden', relation: 'gardens' }.freeze
+
+      ALL = [PLANTING, HARVEST, GARDEN].freeze
+
+      def self.types
+        ALL.map do |model|
+          model[:type]
+        end
+      end
+
+      def self.relations
+        ALL.map do |model|
+          model[:relation]
+        end
+      end
+
+      def self.get_relation(object, type)
+        relation = ALL.select do |model|
+          model[:type] == type
+        end[0][:relation]
+        object.send(relation)
+      end
+
+      def self.get_item(type)
+        class_name = ALL.select do |model|
+          model[:type] == type
+        end[0][:class]
+        class_name.constantize
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


Br3nda:
just needs the rubocop fixes.

maco:
I kind of want to add a constants file that makes a hash with garden/gardens/Garden etc so that in `photos_controller.rb` we can look up class_name versus relation_name etc. and eliminate the case statements (which are a place I'd expect to see a bug crop up with adding photos to another model). What's here so far does at least eliminate having that bug we had before: forgot to list the new model in `destroy_if_unused`

maco:
Ok, rubocop doesn't like that I'm using `has_and_belongs_to_many` in photo_capable.rb and wants me to do `has_many ___, through: ____` instead. And I *could* have it use the class name to work out the join table model's name (and create those models)...**if** it weren't for the fact that they're named PhotosPlantings, GardensPhotos, and HarvestsPhotos. Yeah, oops. They're in opposite orders (I think skud maybe said do them alphabetically?). So, if I rename the PhotosPlantings table to PlantingsPhotos, and add 3 models, then I could make rubocop happy here...or I could tell rubocop to look away.

Thoughts?

Br3nda:
you can ignore on that one line
`# rubocop:disable Rails/HasAndBelongsToMany`

maco:
I finally figured it out!!! 🎉

pozorvlak:
Possible flaky test failure:

```
) CropsController GET crop search fetches the crop search page
     Failure/Error: return response.records.to_a
     Elasticsearch::Transport::Transport::Errors::ServiceUnavailable:
       [503] {"error":"SearchPhaseExecutionException[Failed to execute phase [query_fetch], all shards failed]","status":503}
     # ./app/models/crop.rb:314:in `search'
     # ./app/controllers/crops_controller.rb:68:in `search'
     # ./spec/controllers/crops_controller_spec.rb:45:in `block (3 levels) in <top (required)>'
```

Rerunning now. I *really* wish Travis kept logs around from jobs that have been re-run.

pozorvlak:
The build [passed on re-running](https://travis-ci.org/Growstuff/growstuff/jobs/180104523) :-(

maco:
That's bad news for our CI but good news for me

PR #1097, opened 2016-11-30
Precompile assets in test

pozorvlak:
According to https://github.com/teampoltergeist/poltergeist/issues/677#issuecomment-222919584, a possible cause of the "Request failed to reach server" problem (#901) is that assets are being compiled on-demand, causing enough of a slowdown for PhantomJS to think the connection has failed. This PR turns off lazy asset compilation in the test environment and precompiles assets before testing on Travis.

We should also turn off lazy asset compilation in production - see http://stackoverflow.com/questions/8821864/config-assets-compile-true-in-rails-production-why-not Our Heroku deployments already run `rake assets:precompile`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ before_install:
   - phantomjs --version
 before_script:
   - bundle exec rake db:create db:migrate db:test:prepare
+  - bundle exec rake assets:precompile
 script:
   - bundle exec rubocop --display-cop-names --rails
   - script/gemfile_check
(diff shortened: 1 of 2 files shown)


pozorvlak:
CI did not fail because of a "Request failed to reach server" error!

```
Failures:

  1) Notification doesn't send email to people who don't want it

     Failure/Error: ActionMailer::Base.deliveries.last.to.should_not == [notification.recipient.email]     
     NoMethodError: undefined method `to' for nil:NilClass

     # ./spec/models/notification_spec.rb:41:in `block (2 levels) in <top (required)>'
```

I'll look in to this.

PR #1098, opened 2016-11-30
Run tests with `rake` when DB does not yet exist

pozorvlak:
Previously, `rake spec` ran the tests whether or not the `growstuff_test` database existed or not, but `rake` crashed unless it had already been created. The problem was that `rake` was trying to load the environment, which required the presence of a DB; `rake spec` was running this with the `dev` environment, so succeeded if `growstuff_dev` existed, but `rake` was running everything with the `test` environment, so would fail if `growstuff_test` did not exist.

This patch adds `db:create` to the dependencies of the `spec` task, forcing an unadorned `rake` to create all necessary databases first.

tl;dr you should now be able to just type `rake` to run specs locally.

### DIFF
--- a/lib/tasks/testing.rake
+++ b/lib/tasks/testing.rake
@@ -2,7 +2,7 @@ require 'rake'
 begin
   require 'rspec/core/rake_task'
   task(:spec).clear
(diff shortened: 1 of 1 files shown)


maco:
(agh i'm finding points of confusion in github's UI. don't mind me)

PR #1101, opened 2016-12-01
Updating a few gems

cesy:
Let's have some December updates


cesy:
Looks like we need to make some changes for CodeClimate and SimpleCov. Anyone want to have a go? Otherwise I'll look into it later.

Gemfile.lock
```diff
@@ -147,17 +149,17 @@ GEM
       json
       thread
       thread_safe
-    elasticsearch (2.0.0)
```
CloCkWeRX: This one we want to avoid, the don't use semver. Right now its at elasticsearch 2.0.0; indicating API compatibility with elasticsearch 2.0.0. Upgrading to 5 warrants a corresponding elasticsearch upgrade.

It probably won't break anything too badly in terms of BC, but there are some things to watch out for (or at least that was my experience from 1.4 -> 2.0, a BC break every other release)

Gemfile.lock
```diff
@@ -421,41 +423,37 @@ GEM
       sprockets (>= 2.8, < 4.0)
       sprockets-rails (>= 2.0, < 4.0)
       tilt (>= 1.1, < 3)
-    selenium-webdriver (2.53.4)
+    selenium-webdriver (3.0.3)
```
CloCkWeRX: Lets be narrower with webdriver, capybara, etc updates vs flakey CI (upgrade in individual commits so we can pinpoint regressions)

CloCkWeRX:
I might do another version of this one today, pinning us to the right elasticsearch etc - mind if I close?

CloCkWeRX:
Opened #1126

PR #1103, opened 2016-12-01
Ruby update

cesy:
Updating to 2.3.2. If we like it, we can then do 2.3.3.


Gemfile
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 source 'https://rubygems.org'
 
-ruby '2.3.1'
+ruby '2.3.2'
```
CloCkWeRX: Best to jump straight to 2.3.3 - https://www.ruby-lang.org/en/news/2016/11/21/ruby-2-3-3-released/

PR #1105, opened 2016-12-01
expand wikipedia regex to include -, (, and )

maco:
because some have - or () in name

example:

- https://en.wikipedia.org/wiki/Urad_(bean)
- https://en.wikipedia.org/wiki/Curry-leaf_Tree

(Fixes: #1104)

### DIFF
--- a/app/models/crop.rb
+++ b/app/models/crop.rb
@@ -42,7 +42,7 @@ class Crop < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
   ## Wikipedia urls are only necessary when approving a crop
   validates :en_wikipedia_url,
     format: {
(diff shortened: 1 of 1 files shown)


maco:
If anyone can figure out how to get it to just be that it also allows `-()` that'd be great, but I went for `[:punct:]` because it seems you can't do escaping inside `[ ]`

> To include a backslash as a character without any special meaning inside a character class, you have to escape it with another backslash. [\\x] matches a backslash or an x. The closing bracket ], the caret ^ and the hyphen - can be included by escaping them with a backslash, or by placing them in a position where they do not take on their special meaning. The POSIX and GNU flavors are an exception. They treat backslashes in character classes as literal characters. So with these flavors, you can't escape anything in character classes.

from: http://www.regular-expressions.info/charclass.html

Issue #1107, opened 2016-12-01
cannot find photo constants


maco:
Tests worked, but not browser, go figure. I couldn't seed the DB to try it out thanks to #1104, and now I see I need to add some `require` lines. 

Currently blocked by #1106

Br3nda:
Fixed a while ago

Issue #1106, opened 2016-12-01
cannot connect flickr account


maco:
This looks like a side effect of moving from /cms/foo URLs to /foo URLs

<img width="489" alt="screen shot 2016-12-01 at 12 37 01 pm" src="https://cloud.githubusercontent.com/assets/401264/20804944/e2a001c8-b7c2-11e6-8d40-e73653c8ecef.png">

maco:
Using /members/auth/flickr instead sends you to Flickr properly, but then the callback lands at /members/auth/flickr/callback which doesn't exist

cesy:
So should we just updating our routes.rb to point /auth to the right place?

maco:
I think so I'm just not sure how

Br3nda:
I had a go at this, and noticed this error
```W, [2017-02-05T12:09:49.405681 #4929]  WARN -- : You are setting a key that conflicts with a built-in method OmniAuth::Strategy::Options#display defined in Kernel. This can cause unexpected behavior when accessing the key via as a property. You can still access the key via the #[] method.
```

Br3nda:
I've had a couple goes at this. I'm not sure it has anything to do with the cms. Removing the cms doesn't help.

maco:
The cms is just catching any route that's not already defined, so I guess
that makes sense. It used to only catch /cms/* rather than /* so I was
thinking they're related.

Br3nda:
Works in release 14
broken in release 15.

Br3nda:
did some testing, and i think connecting a flickr account was broken in this commit
c60609d50708cb890be54f2bdcf7eca2e7ea80b3

maco:
huh. I imagine it's one of the gem updates, but... wow.

Br3nda:
possibly the rails update. But there's adding of facebook auth about the same time too.

PR #1108, opened 2016-12-01
WIP: add require_relative to photo_capable concern for constants

maco:
starts #1107 
need to add require changes in photo.rb and photos_controller.rb but can't really test until #1106 is fixed

### DIFF
--- a/app/models/concerns/photo_capable.rb
+++ b/app/models/concerns/photo_capable.rb
@@ -1,3 +1,4 @@
+require_relative '../../constants/photo_models.rb'
 module PhotoCapable
   extend ActiveSupport::Concern
 


Br3nda:
This does fix the error.
Maybe a feature spec for the plantings#new page will catch this in the future.

maco:
@Br3nda ok, so photo.rb and photos_controller.rb are fine as they are? That seems surprising.

Br3nda:
i'm getting this error from dev branch
```
NameError - uninitialized constant Growstuff::Constants:
  activesupport (4.2.7.1) lib/active_support/dependencies.rb:533:in `load_missing_constant'
  activesupport (4.2.7.1) lib/active_support/dependencies.rb:184:in `const_missing'
  app/models/photo.rb:4:in `<class:Photo>'
  app/models/photo.rb:1:in `<top (required)>'
```

the specs fail too...

cherry-pick this PR's commit fixes it.

maco:
There's probably some way to make it autoload so it's not repeatedly loaded to make that warning, but merging this so we go from error to warning is probably a good step

CloCkWeRX:
Perhaps 13dd20c8ff603d4155978b2b863c985a4bfdc43d kills the warning a bit better (still no autoload)

PR #1109, opened 2016-12-01
add photos to seeds

### LINKED ISSUE #495: Seed photos
You should be able to attach photos to seeds, as you do for plantings, harvests, etc.

maco:
Fixes #495 

untested, waiting on #1106 and #1107

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161129021533) do
+ActiveRecord::Schema.define(version: 20161201154922) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


Br3nda:
Does the photo_mappings table not replace the photo_seeds photo_crops tables etc?

maco:
@Br3nda after talking to @CloCkWeRX i learned that having a table that has fake foreign keys to multiple tables based on a column is a crappy database performance thing so we should stick to ordinary join tables. Thus, did not do `photo_mappings` after all.

Br3nda:
@maco and  create table photo_mappings is still in this pr.

db/schema.rb
```diff
@@ -373,6 +381,11 @@
     t.integer "planting_id"
   end
 
+  create_table "photos_seeds", id: false, force: :cascade do |t|
```
Br3nda: Should this table still be in here?

db/schema.rb
```diff
@@ -355,6 +355,14 @@
     t.integer "product_id"
   end
 
+  create_table "photo_mappings", force: :cascade do |t|
```
Br3nda: Should this table still be here?

maco: oooo good catch

CloCkWeRX:
Hey, crappy is too strong of a word. if we take a db centric view, its bad.
if we take an oop/orm view, its good. Was highlighting tradeoffs.

PR #1113, opened 2016-12-02
Homepage redesign redux (updated)

### LINKED ISSUE #628: Homepage redesign, 2015

CloCkWeRX:
Replaces #697

From the original:

Fixes #628

Replaces #674

~~@Skud :~~
~~> I think after you select something from that search UI it should take you direct to the crop page in question (eg. /crops/eggplant) not to search results (unless there's not an exact match).~~

~~This should work in production where we're using Elasticsearch but it doesn't seem to work in development. See inline note.~~

@Skud :

> Why is the crop search not an autosuggest? If I type "pea" or "berry" for instance I don't see the list.

Good call. I need to refactor autosuggest a bit. Should I do that now or break into a new story?

**Question**: Should the [signed in user homepage redesign](http://talk.growstuff.org/t/major-homepage-redesign/168) be part of this PR?

**Screenshot** (updated March 1, 2015)
screen shot 2015-03-01 at 5 07 09 pm

Updated screenshot of homepage:
image
image


CloCkWeRX:
So... this is a bit more mergable, but I think its a bit less nice that the current logged in user homepage.

It might be better to redo this to mix in the map feature to the existing content or otherwise split into 'logged in experience' vs 'new user journey'

pozorvlak:
Lots of `Missing template home/_blurb` errors in CI: did you forget to add a file?

CloCkWeRX:
The merge was quite a pain, it was probably eaten at some point. Will restore in a bit

Br3nda:
The screenshots look great.  Need any help getting this mergable?

Br3nda:
How much of this can be saved and used now?

I can't access the talk website anymore -- There seems to be some views missing from the branch, so I can't just checkout that branch and have a look.

cesy:
The talk website was taken down as it was costing us extra money every month, but I have a backup I can send you, or let me know which page you need and I can try and dig it out. I don't think we're wedded to any particular design now, so anything that's an improvement is great.

Br3nda:
Closing this after a year. But if anyone has markup of the design somewhere we can revisit

PR #1115, opened 2016-12-03
Bw/codeclimate rubocop

Br3nda:
Our use of .rubocop_todo.yml is causing codeclimate to ignore all the issues on our todo list. It's giving our code a much higher score than it should have.

This proposed change moves the config to .rubocop_codeclimate.yml so all the remaining issues continue to be flagged by codeclimate, and any newly introduced ones are flagged on the PR

p.s. Code climate is going to really not like this PR - which is the goal here.


maco:
You've got a merge conflict, and the line of `<<<<<<` is making the test crash

Br3nda:
The more time i spend on it, the less sense this advice from codeclimate makes.
Let's leave this for another time.

PR #1119, opened 2016-12-06
Fixed parentheses that are interp-ed as grouped expression

Br3nda:


app/mailers/notifier.rb
```diff
@@ -16,7 +16,7 @@ def notify(notification)
     @reply_link = reply_link(@notification)
 
     # Encrypting
-    @signed_message = verifier.generate ({ member_id: @notification.recipient.id, type: :send_notification_email })
+    @signed_message = verifier.generate(member_id: @notification.recipient.id, type: :send_notification_email)
```
maco: was it intentional to ditch the `{}`?

Br3nda: So, i need I'm really a python dev who pretends to know ruby.
Rubocop complained about the {} being unnecessary, hinting that ruby treats it as a hash regardless. I took it at it's word.

Br3nda: read up some - and you're right. Good catch. Fixing this now

maco:
It was totally possible I was also being Pythonic, but it seemed worth
checking

Br3nda:
Could someone with the powers fix up coveralls? I would love to be able to check that CI touched the lines that changed

image

CloCkWeRX:
There's something funny going on - its clearly there on dev via git.

I got an email notification just yesterday from data@gitshield.com - 	Tue, Dec 6, 2016 at 5:22 PM (thats before you authored the PR apparently though)
```

File Name
 
/notifier.rb

 
Pattern Matched
 
Possible Ruby secret key(s)
```

That's clearly wrong, but I'm wondering if a similar thing is happening with other apps and they think they are protecting us.

PR #1120, opened 2016-12-06
Yet another wikipedia url char added to regex

Br3nda:
wikipedia urls can have a fullstop on them. Infact we have one in our db:seeds
https://en.wikipedia.org/wiki/Capsicum_annuum_var._glabriusculum


pozorvlak:
This is starting to feel like another game of whack-a-mole: what would be the problem with reverting to [:punct:]?

Also, we could use Ruby's [flexible quote syntax](https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Literals#The_.25_Notation) to avoid leaning-toothpick syndrome in the regex.

PR #688, opened 2015-02-01
Bar graphs -- initial draft pull request

Marlena:
This is a draft pull request of bar charts, suggested by @tygriffin, for sunniness and plantings and is not meant to be merged in, but just to show what's been done.  This post will go over some of the big pieces in the pull request and also detail some of the work that's left.

Here is a link to the initial [Tracker story](https://www.pivotaltracker.com/story/show/53219527).
## To see the chart in action

Open a crops page.

screen shot 2015-02-01 at 7 58 05 am
## Design

This example is the MVP I've been working towards.  

horizontal_chart_example
## Javascript

These charts are being written in javascript with an emphasis on declarative over imperative style.  This is why information about what the charts should look like is stored in json data.  The javascript was written with the help of local pair, Ryan Dy ([rdy](https://github.com/rdy)) who writes javascript, node and react.js code all day, everyday for Pivotal Labs.  Big thanks to rdy!
## Component Architecture

While the higher level structure is a horizontal bar graph, the underlying pieces could be used for other types of graphs.  Some refactoring is needed as I was developing my understanding of component based architecture while building this.
## Usage

The charts are designed to be easy to reuse in different areas of Growstuff.  See crops.js.erb for usage.
## D3

The charts are rendered as svg by D3.  While D3 is a widely used library for visualization, it does not necessarily promote good software engineering practices.  Often, the code for a chart looks like one messy chain of method calls and it's hard to tell what part each call plays in the chart.  I did a lot of work to break the code for the chart into more digestible objects and to make what each object is doing clearer for easier maintenance.  The side effect of this is that you can also see how to approach OO in javascript with classes and objects.
## Jasmine Tests & TDD

Although this started as a messy spike in a personal repo, all of the code in the Growstuff repo was test driven with [Jasmine](http://jasmine.github.io/).

To run the tests: rake jasmine

Jasmine will then print out a URL you can paste into a browser to see the tests.  By clicking on a test, you can see Jasmine run that one test.  If you open the browser inspector, you can see the page content under test rendered in the jasmine-content div.
## Still working on
- Numbers indicating the total for each category
- Currently, the data you see on the chart is fake.  There will need to be access to the sunniness and planting data
- What to do about the empty category:  Some people don't fill out either the sunniness or planting data.  Does this category belong on the chart?
- Style of this area on the crops page and location on the crops page

[...]

PR #1121, opened 2016-12-06
Bw/bar graphs

Br3nda:
I think PR #688 is awesome, so here is it with the merge conflicts fixed.


CloCkWeRX:
> Color literals like #ffa500 should only be used in variable declarations; they should be referred to via variable everywhere else.

I don't think that's used anywhere else (its... orange); though we do have
{code}
$beige: #f3f1ee
$brown: #413f3b

$green:  #5f8e43
$blue:   #2f4365
$red:    #8e4d43
$orange: #b2685c
$yellow: #b2935c
{code}

Weirdly; #b2685c is completely not orange. I didn't quite see where it was used in the UI looking at the code only

CloCkWeRX:
I  think with the d3 'similar code', its just going to be something we ignore. Specifically:
```
Similar code found in 1 other location (mass = 73)
NEW
    return d3.append('g')
      .attr("class", "bar-label")
      .selectAll("text")
      .data(bars.map(function(bar){ return bar.name;}))
      .enter()
```

```
To ignore this duplication issue modify your .codeclimate.yml file and add 16dbcb58d6caa7ccfe241417831ecfa6 to exclude_fingerprints. An example is shown below.

.codeclimate.yml
engines:
  # ... CONFIG CONTENT ...
  duplication:
    enabled: true
    # ... CONFIG CONTENT ...
    exclude_fingerprints:
    - 16dbcb58d6caa7ccfe241417831ecfa6
# ... CONFIG CONTENT ...
```

CloCkWeRX:
```
for (i; i < data.bars.length; i++){
      barValues.push(data.bars[i].value)
    }
```
Kind of keen to see if we can ```barValues.push(data.bars.map(function(bar) { return bar.value; });```

spec/javascripts/support/jasmine.yml
```diff
@@ -0,0 +1,124 @@
+# src_files
```
CloCkWeRX: I wonder if spec/javascript will execute in travis automagically, or if we have to invoke something else.

Generally like jasmine for complex JS testing

pozorvlak: That would be *awesome*, but it's not the end of the world if we add another line to .travis.yml.

In general, I'm +10000 to us adding JavaScript unit tests via some means, and I'm not too bothered about which framework we use.

CloCkWeRX: Let's make #1125 also be 'make js tests work in travis'

Br3nda:
Css isn't my thing. Anyone who can quickly work out how to appease code climate?

pozorvlak:
The one-line duplication between `bar_group.js` and `width_scale.js` looks trivial, but actually it's one identical line embedded in two functions which are logically identical but just different enough to fool CodeClimate. So it would be good to factor that out. The function in question boils down to `bars.map(function(bar) { return bar.value; })`, which I see you're using in the *other* "similar code found" issue - can we rely on the `map` method existing?

The "similar code found between bar_group.js and bar_label_group.js" issue looks like a false positive to me - sure, we could factor out some common code, but the similarity is superficial so we wouldn't gain anything thereby. To ignore it, add `16dbcb58d6caa7ccfe241417831ecfa6` to `engines.duplication.exclude_fingerprints` in .codeclimate.yml, as follows:

```
engines:
  # ... CONFIG CONTENT ...
  duplication:
    enabled: true
    # ... CONFIG CONTENT ...
    exclude_fingerprints:
    - 16dbcb58d6caa7ccfe241417831ecfa6
# ... CONFIG CONTENT ...
```

spec/javascripts/support/jasmine.yml
pozorvlak: It doesn't look like the tests are being run in Travis; if I run `rspec spec/javascripts` locally I get 0 tests executed. @Br3nda, how are you invoking the tests locally?

pozorvlak:
Oh, @CloCkWeRX already said everything I did a fortnight ago. D'oh. Oh well, at least I'm somewhat up to speed with the issues now.

CloCkWeRX:
> can we rely on the map method existing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map - mostly.


Feature	Chrome	Firefox (Gecko)	Internet Explorer	Opera	Safari
Basic support	(Yes)	1.5 (1.8)	9	(Yes)	(Yes)

pozorvlak:
We can use [jQuery.map](http://api.jquery.com/jquery.map/): we're already using jQuery.

Br3nda:
Superceded

PR #1122, opened 2016-12-07
Gardens query was running twice in contoller

Br3nda:


CloCkWeRX:
In theory, that just built an SQL string until you read from the enumerable put into ```@gardens```; so that shouldn't have been a huge perf deal.

LGTM, once the CC complaint is sorted

pozorvlak:
Looks fine apart from the CC issue; I'm worried by the Travis failure, though that looks unrelated - @maco, any idea what's going on there?

Br3nda:
@pozorvlak i've fixed the rubocop complaint and re-run the travis ci

PR #1123, opened 2016-12-07
Less busy gardens page

Br3nda:
I'm expecting this to be contentious.. but...

the gardens#index page is really busy. It has lots of the attributes of a garden including all the placeholder text like "no description provided." and " not specified".
There were 3 links to every garden, the header, the name, and the "show more plantings".

I pruned lots of it - as i want it to be just an index, used to find a garden, not show it all.

### DIFF
--- a/app/assets/stylesheets/overrides.sass
+++ b/app/assets/stylesheets/overrides.sass
@@ -319,8 +319,11 @@ $state-success-bg: lighten($green, 50%)
     text-overflow: ellipsis
     overflow: hidden
 
-#gardens_panel_body
-  height: 20em
+.garden-plantings
+  list-style-type: none
+  font-size: 90%
+
+.gardens_panel_body
 
 .form-group.required .control-label:before
   content: "* "
--- a/app/views/gardens/_thumbnail.html.haml
+++ b/app/views/gardens/_thumbnail.html.haml
@@ -1,36 +1,21 @@
 .panel.panel-success
   .panel-heading
     %h3.panel-title
-      = link_to "#{garden.owner.login_name}'s garden", garden
+      = link_to garden.name, garden
       - if can? :edit, garden
         %a.pull-right{:href => edit_garden_path(garden), :role => "button", :id => "edit_garden_glyphicon"}
           %span.glyphicon.glyphicon-pencil{:title => "Edit"}
-  .panel-body{:id => "gardens_panel_body"}
+  .panel-body{:class => "gardens_panel_body"}
     .row
       .col-md-4
         = link_to image_tag((garden.default_photo ?  garden.default_photo.thumbnail_url : 'placeholder_150.png'), :alt => garden.name, :class => 'img'), garden
       .col-md-8
-        %dl.dl-horizontal
-          %dt Name :
-          %dd= link_to garden.name, garden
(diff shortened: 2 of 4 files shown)


Br3nda:
Before:
image

After:
image

CloCkWeRX:
I'm +1 for cleaning up the card UI styling issues (ie, the appearance of the UL items), but -1 for taking away content.

That said, there might be a strong case for splitting out 'user garden index' and 'garden index' into two different kinds of UI.

IE; if this were my gardens page:

image

app/views/gardens/_thumbnail.html.haml
```diff
@@ -1,36 +1,21 @@
 .panel.panel-success
   .panel-heading
     %h3.panel-title
-      = link_to "#{garden.owner.login_name}'s garden", garden
+      = link_to garden.name, garden
```
CloCkWeRX: I prefer this; with maybe a fallback for garden name = blank (or we add model validations to require it if not already there)

app/views/gardens/_thumbnail.html.haml
```diff
@@ -1,36 +1,21 @@
...
       .col-md-4
         = link_to image_tag((garden.default_photo ?  garden.default_photo.thumbnail_url : 'placeholder_150.png'), :alt => garden.name, :class => 'img'), garden
       .col-md-8
-        %dl.dl-horizontal
-          %dt Name :
-          %dd= link_to garden.name, garden
-          %dt Location :
+        %dl
+          - if garden.location.present?
+            %dt Location:
```
CloCkWeRX: I do prefer "Foo:" to "Foo :" very much, we should keep this - or even drop the ":" and rely on CSS to convey visual seperation

app/assets/stylesheets/overrides.sass
```diff
@@ -319,8 +319,11 @@ $state-success-bg: lighten($green, 50%)
     text-overflow: ellipsis
     overflow: hidden
 
-#gardens_panel_body
-  height: 20em
+.garden-plantings
+  list-style-type: none
```
CloCkWeRX: +1

Br3nda:
The only attribute really removed is the garden.description. Everything else is now only on the page once, or has the "not specified" placeholders  gone.

Br3nda:
I want to send another PR after this, where you can default to not seeing the inactive gardens. I've got gardens from a house i live in years ago and I don't think they should be all over the UI unless i am looking for inactive gardens.

CloCkWeRX:
From what I understand, the main reason for the placeholder text is to try to ensure all of the cards are the same height - it can be really noticable on http://www.growstuff.org/gardens if things are a bit skewiff.

How would you feel if we did the CSS cleanup changes from this PR as a first step - I think they have value - and we see what the consensus is on the rest?

I reckon the inactive gardens hidden by default is also a good one to do; as it'll encourage pages like http://www.growstuff.org/gardens to show only the most interesting stuff (particularly if we can hide 'draft' gardens or gardens with no attributes as well)

Br3nda:
My plan is to split this PR up in to each change (and the newly suggested changes)

maco:
Can't we set a height on them, then?

PR #1124, opened 2016-12-07
Cancan already loads the resource

Br3nda:
removed lots of unnecesary loading of a controller's resource. The `load_and_authorize_resource` already does this for us.

### DIFF
--- a/app/controllers/account_types_controller.rb
+++ b/app/controllers/account_types_controller.rb
@@ -13,8 +13,6 @@ class AccountTypesController < ApplicationController
 
   # GET /account_types/1
   def show
-    @account_type = AccountType.find(params[:id])
-
     respond_to do |format|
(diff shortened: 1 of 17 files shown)


Br3nda:
This creates 16 new places where the code is now very similar. I say this is good news, as we can reduce some of those to re-usable code next.

CloCkWeRX:
It lgtm, but I get wary of too much implicit behaviour vs explicit - this
is non obvious unless you know devise/cancan

Could you add to a Contributing.md or similar? We have that plus 'avoid
view tests for feature tests', I'm sure there are more assumptions we have
made somewhere.

CloCkWeRX:
Added some notes to https://github.com/Growstuff/growstuff/wiki/Coding-style-guide

PR #1127, opened 2016-12-14
Guard clauses

Br3nda:
also less code than before.


Br3nda:
Side note, if you rely on load_and_authorize_resource to do the authorizing, and then load the controller's model again after, you may have loaded something different to what you check authorizing on.

I spent longer than i wanted to checking we were secure. It seems like a trap for the unwary.

Br3nda:
I'm stuck because my usual technique is controller specs. You submit posts and puts that you're not supposed to, (e.g. post gardens#update to someone else's garden) and verify the authorize stops you.

But this project has a "feature specs only" rule. Is there an example of how to submit a http post to a controller methods without a web form, using feature specs?

Br3nda:
I check now, and can't see any feature gardens where we try to edit someone else's garden. Is this even possible to do in feature specs?

maco:
@pozorvlak and I have talked a few times about bringing back controller
tests

Br3nda:
@maco i would be in favour of that. Controller specs serve a different function to feature specs - especially now that we have APIs in apps these days.

pozorvlak:
Yeah, just write a controller spec.

PR #1136, opened 2016-12-24
Controller garden specs for member doing what they shouldn't

Br3nda:


app/controllers/gardens_controller.rb
```diff
@@ -45,8 +45,7 @@ def edit
   # POST /gardens
   # POST /gardens.json
   def create
-    params[:garden][:owner_id] = current_member.id
-    @garden = Garden.new(garden_params)
+    @garden.owner_id = current_member.id
```
Br3nda: ALSO -- we don't need to call `Model.new` in our controllers. the `load_and_authorize_resource` already does this on `#new`

Br3nda:
Note: i changed from using login_member to instead including the devise examples, because creating a member causes a Garden.new to be called, and was messing with the `expect(Garden).to receive(:new)` exceptions.

spec/controllers/gardens_controller_spec.rb
```diff
@@ -12,11 +12,72 @@
...
+  include Devise::Test::ControllerHelpers
+  let(:valid_params) { { name: 'My second Garden' } }
 
-  def valid_attributes
-    member = FactoryGirl.create(:member)
-    { name: 'My Garden', owner_id: member.id }
+  context "when not signed in" do
+    let(:garden) { double('garden') }
+    describe 'GET new' do
+      before { get :new, id: garden.to_param }
```
pozorvlak: It looks like you're consistently using

```
describe "METHOD endpoint" do
   before { make request }
   it { assert something }
end
```

when you could use

```
it "METHOD endpoint" do
    make request
    assert something
end
```

which is what I'm used to seeing - the extra blocks tripped me up on first reading this code. Or am I missing something?

Br3nda: i use `before` so that all the asserts happen -- it means when you break something, you get more clues on the first failed test run. it will say which are failing, instead of just the first to fail in an `it` block. 
e.g.
```
describe "METHOD endpoint" do
   before { make request }
   it { assert something }
   it { assert more }
   it { assert yet more }
end
```

instead of 
```
describe "METHOD endpoint" do
   it do
     make request
     assert something
     assert more
     assert yet more
  end
end
```

Br3nda: basically, i find the more descriptive errors greatly speed up debugging.

pozorvlak: Right - I was confused because you were using `before` in cases where there was only one assertion.

PR #1138, opened 2017-01-02
Garden's name in link

Br3nda:
uses the garden's name in the link to the garden.
also truncates it to 50 chars, because really long names muck with the display on many pages.

this was one of the changes in #1123

### DIFF
--- a/app/helpers/gardens_helper.rb
+++ b/app/helpers/gardens_helper.rb
@@ -9,6 +9,10 @@ module GardensHelper
     end
   end
 
+  def display_garden_name(garden)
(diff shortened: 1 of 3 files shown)


app/helpers/gardens_helper.rb
```diff
@@ -9,6 +9,10 @@ def display_garden_description(garden)
     end
   end
 
+  def display_garden_name(garden)
+    truncate(garden.name, length: 50, separator: ' ', omission: '... ')
```
CloCkWeRX: http://www.w3schools.com/cssref/css3_pr_text-overflow.asp is also another option for preventing overflow issues which tends to work a bit better regardless of text size

Issue #1089, opened 2016-11-29
proposal: turn on review requirements


Br3nda:
Github now  supports requiring a number of reviews and checks passing before making the merge button green.

It looks like this on my other projects:
![github](https://cloud.githubusercontent.com/assets/12266/20695441/cf1e82a0-b650-11e6-89b0-a0b4e6c69114.png)

Let's turn it on for growstuff. A PR could require 2 reviews from commiters, and for Travis to pass. 

Essentially it makes the button look a bit scary if the criteria isn't met, and makes the button a comforting green if it's good.

maco:
:+1: to this from me.

CloCkWeRX:
I'm for leaving it as optional - its useful to add feedback in a structured way; but I don't think we should rigidly enforce it.
Its rare that we push without review and we haven't done so in a way that Breaks Everything

maco:
Well, I don't know about requiring 2 reviewers since sometimes some of us
wander off for a week or whatever. At my work the way we're using it is
that approve/change indicates whether the reviewer is saying "this is a
nitpick and not totally necessary, but..." or "this has to change."

cesy:
Yeah, we should set it to one reviewer as that's what we currently require,
but let's turn it on and play with it.

pozorvlak:
I'm OK with trying this out, but currently our problem is that we're not
fast enough at reviewing and merging things, and AFAICT this has the
potential to make that problem worse. So, let's try it for a bit, and turn
it off again if it's not working.

Br3nda:
This wouldn't enforce. You can still press the button. The change only makes the button scary looking when nobody has reviewed or tests have failed.

Br3nda:
I've found this can make it faster. We don't have to keep checking and re-reading if the pr is reviewed as good. It's down there ar the bottom.

Br3nda:
I have turned this on. I also ticked that the branch should be up-to-date with the latest changes in the branch you propose merge into. 

![image](https://cloud.githubusercontent.com/assets/12266/20728709/d7693fac-b6e3-11e6-917c-0b02bf9c6beb.png)

maco:
Thanks. I noticed the out of date warning. I tried to find the checkbox to
do this proposal and couldn't. Where'd ya find the setting?

Br3nda:
It's under "protect  branches". I turned it on for dev and master

PR #1140, opened 2017-01-03
descriptive spec

Br3nda:
This pushes the model name out into the rspec output.

One of the models was failing, but i couldn't tell which - and the tests stopped on the first failure.
This change makes the example tested for all models in the list, and prints the route in the output as it goes.


Br3nda:
I also reduced the code duplication in the file

pozorvlak:
I'm confused about what our merge requirements are now, and reading the discussion on https://github.com/Growstuff/growstuff/issues/1089 doesn't really enlighten me. The "needs approval" light is great for showing "a reviewer has noticed problems with this PR, don't merge it", which is great (thanks, @Br3nda!). But if I approve of a PR, and nobody has disapproved of it, can I go ahead and merge it? That's how we used to do things, and seems sensible given that we've had lots of problems with review velocity.

Anyway, I'm going to go ahead and merge this PR because it's simple and clearly a Good Thing.

Br3nda:
I think yes, if the PR has been there a few days (or is trivial) then go ahead and press the green button.

PR #1141, opened 2017-01-04
Show only active plantings by default

Br3nda:
This removes all the plantings that finished years ago from the plantings#index - can be found again from one button push


Br3nda:
I'd like to do the same to gardens

app/views/plantings/index.html.haml
```diff
@@ -8,11 +8,15 @@
...
-        = link_to "View everyone's plantings", plantings_path, :class => 'btn btn-default'
+        - content_for :buttonbar do
+          = link_to 'My Active Plantings', plantings_by_owner_path(:owner=>current_member.slug), :class => 'btn btn-default'
+          = link_to 'All My Plantings', plantings_by_owner_path(:all=>1, :owner=>current_member.slug), :class => 'btn btn-default'
+          = link_to "Everyone's plantings", plantings_path, :class => 'btn btn-default'
     - else # everyone's plantings
       = link_to 'Plant something', new_planting_path, :class => 'btn btn-primary'
       - if current_member
         = link_to 'View your plantings', plantings_by_owner_path(:owner => current_member.slug), :class => 'btn btn-default'
+      = link_to 'Show all (including inactive)', plantings_path(:all=>1)
```
pozorvlak: Maybe "Show all" if `all == 0` and "Show only active" if `all == 1`?

pozorvlak:
+1 to doing this for gardens too.

app/controllers/plantings_controller.rb
```diff
@@ -5,16 +5,19 @@ class PlantingsController < ApplicationController
...
-    @owner = Member.find_by(slug: params[:owner])
-    @crop = Crop.find_by(slug: params[:crop])
+    @owner = Member.find_by(slug: params[:owner]) if params[:owner]
+    @crop = Crop.find_by(slug: params[:crop]) if params[:crop]
     @plantings = if @owner
-                   @owner.plantings.includes(:owner, :crop, :garden).paginate(page: params[:page])
+                   @owner.plantings
                  elsif @crop
-                   @crop.plantings.includes(:owner, :crop, :garden).paginate(page: params[:page])
+                   @crop.plantings
```
pozorvlak: Do we want "all of @pozorvlak's catnip plantings"? I guess that's a separate issue even if we do.

Br3nda: It should match the  existing behaviour.

app/views/plantings/index.html.haml
Br3nda: Okay. Can you explain how that is better?

app/controllers/plantings_controller.rb
pozorvlak: I think we're talking at cross-purposes. Here's what I meant to say:

- The code here matches the existing behaviour.
- That means we can filter by owner *or* crop, but not both.
- Maybe it would be useful to filter by both.
- That's out of scope for this PR, though.
- So the code above is fine.

app/views/plantings/index.html.haml
pozorvlak: Sure. Right now, the "Everyone's plantings, including inactive" view looks like this:
screenshot from 2017-01-05 12-12-08

The "Show all (including inactive)" link takes you to the page you're already looking at, so it isn't useful. AFAICT, if you want to see only active plantings, you need to manually edit the URL bar, or perhaps hit "Back" if that's where you came from. I'm suggesting making that link into a toggle rather than a one-way journey.

Br3nda: Oh I see now.  Yes, let's make that better.

Br3nda: It's now a check box (along with a small refactor)

app/controllers/plantings_controller.rb
```diff
@@ -5,15 +5,10 @@ class PlantingsController < ApplicationController
...
-                   @owner.plantings.includes(:owner, :crop, :garden).paginate(page: params[:page])
-                 elsif @crop
-                   @crop.plantings.includes(:owner, :crop, :garden).paginate(page: params[:page])
-                 else
-                   Planting.includes(:owner, :crop, :garden).paginate(page: params[:page])
-                 end
+    @owner = Member.find_by(slug: params[:owner]) if params[:owner]
+    @crop = Crop.find_by(slug: params[:crop]) if params[:crop]
+    @show_all = params[:all] == '1'
+    @plantings = plantings
```
Br3nda: Moved this out, as the method go too big (and rubocop insisted).

app/views/plantings/index.html.haml
```diff
@@ -1,21 +1,10 @@
 - content_for :title, @owner ? t('.title.owner_plantings', owner: @owner) : @crop ? t('.title.crop_plantings', crop: @crop.name) : t('.title.default')
+
+= render 'nav', owner: @owner, show_all: @show_all
```
Br3nda: This makes the navigation look consistent across the 3 ways to view a plantings#index.
I want to do the same to gardens and crops too

Br3nda:
image

image

image


image

Br3nda:
@pozorvlak can you take another look now?

pozorvlak:
Looks good!

Issue #1137, opened 2016-12-26
Linking harvests to a planting 


Br3nda:
I want to track what harvest I get from a planting. Can we link them in data and in the ui?

PR #1142, opened 2017-01-04
Link a harvest to a planting, and display

Br3nda:
This links a harvest to a planting (and thus through to a garden)

Initially only displaying on plantings - but we can link it through to see what the yield from a garden is, season by season, year to year. 

We could also link harvest it to companion plantings, and optional crop rotation - and other events we implement later.

This is for #1137

image
image
image

### DIFF
--- a/app/controllers/harvests_controller.rb
+++ b/app/controllers/harvests_controller.rb
@@ -26,10 +26,15 @@ class HarvestsController < ApplicationController
     end
   end
 
+  def show
+    @planting = @harvest.planting if @harvest.planting_id
+  end
+
   # GET /harvests/new
   # GET /harvests/new.json
   def new
     @harvest = Harvest.new('harvested_at' => Date.today)
+    @planting = Planting.find_by(slug: params[:planting_id]) if params[:planting_id]
 
     # using find_by_id here because it returns nil, unlike find
     @crop = Crop.find_or_initialize_by(id: params[:crop_id])
@@ -42,14 +47,13 @@ class HarvestsController < ApplicationController
 
   # GET /harvests/1/edit
   def edit
+    @planting = @harvest.planting if @harvest.planting_id
   end
 
   # POST /harvests
   # POST /harvests.json
   def create
-    params[:harvest][:owner_id] = current_member.id
-    params[:harvested_at] = parse_date(params[:harvested_at])
-    @harvest = Harvest.new(harvest_params)
+    @harvest.crop_id = @harvest.planting.crop_id if @harvest.planting_id
 
     respond_to do |format|
       if @harvest.save
@@ -90,7 +94,10 @@ class HarvestsController < ApplicationController
   private
 
   def harvest_params
-    params.require(:harvest).permit(:crop_id, :harvested_at, :description, :owner_id,
-      :quantity, :unit, :weight_quantity, :weight_unit, :plant_part_id, :slug, :si_weight)
+    params.require(:harvest)
+      .permit(:planting_id, :crop_id, :harvested_at, :description,
+        :quantity, :unit, :weight_quantity, :weight_unit,
+        :plant_part_id, :slug, :si_weight)
+      .merge(owner_id: current_member.id)
   end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -92,6 +92,8 @@ class Ability
     can :create,  Harvest
     can :update,  Harvest, owner_id: member.id
     can :destroy, Harvest, owner_id: member.id
+    can :update,  Harvest, owner_id: member.id, planting: { owner_id: member.id }
+    can :destroy, Harvest, owner_id: member.id, planting: { owner_id: member.id }
 
     can :create, Photo
     can :update, Photo, owner_id: member.id
--- a/app/models/harvest.rb
+++ b/app/models/harvest.rb
@@ -7,6 +7,7 @@ class Harvest < ActiveRecord::Base
   belongs_to :crop
   belongs_to :owner, class_name: 'Member'
   belongs_to :plant_part
+  belongs_to :planting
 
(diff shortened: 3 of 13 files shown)


Br3nda:
I need tests.

app/controllers/harvests_controller.rb
```diff
@@ -65,6 +72,8 @@ def create
   # PUT /harvests/1
   # PUT /harvests/1.json
   def update
+    planting_is_mine!
```
Br3nda: I don't know why it doesn't check owner id on update. It's an allowed param. So I think it should.

app/controllers/harvests_controller.rb
```diff
@@ -26,10 +26,15 @@ def index
     end
   end
 
+  def show
+    @planting = Planting.find(@harvest.planting_id) if @harvest.planting_id
```
CloCkWeRX: ```@planting = @harvest.planting``` ?

app/controllers/harvests_controller.rb
```diff
@@ -90,7 +99,14 @@ def destroy
...
   def harvest_params
-    params.require(:harvest).permit(:crop_id, :harvested_at, :description, :owner_id,
+    params.require(:harvest).permit(:planting_id, :crop_id, :harvested_at, :description, :owner_id,
       :quantity, :unit, :weight_quantity, :weight_unit, :plant_part_id, :slug, :si_weight)
   end
+
+  def planting_is_mine!
+    return unless @harvest.planting_id
+    @planting = Planting.find(@harvest.planting_id)
+    raise "Planting owner does not match" unless @planting.owner_id == current_member.id
```
CloCkWeRX: So we model this with
```
      can :create,  Planting
      can :update,  Planting, garden: { owner_id: member.id }
      can :destroy, Planting, garden: { owner_id: member.id }
```

It should be checked by something like
```can? :update @planting``` or ```authorise! :update @planting```
... though for some reason we have only really done this kind of checking at the view layer, and rarely in the controller layer.

Its probably because it happens implicitly in a lot of places:
> ... you can use the load_and_authorize_resource method in your controller to load the resource into an instance variable and authorize it automatically for every action in that controller.

File: app/controllers/harvests_controller.rb (hunk 2 above)
Comment (Br3nda) [inline app/controllers/harvests_controller.rb]: it didn't load. how do i convince rails to do that?

Comment (Br3nda) [inline app/controllers/harvests_controller.rb]: `belongs_to :planting`
got it

**Comment by Br3nda** [comment]
@CloCkWeRX can you give an example? All our :create in ability.rb does not check owner. It checks in the controller.

**Comment by Br3nda** [comment]
Reading the docs, i can expand it out to 
```
    can :create,  Harvest do |harvest|
      if harvest.planting
        harvest.planting.owner_id == member.id && harvest.owner_id == member.id
      else
        true
      end
    end
    can :update, Harvest do |harvest|
      if harvest.planting
        harvest.planting.owner_id == member.id && harvest.owner_id == member.id
      else
        harvest.owner_id == member.id
      end
    end
    can :destroy, Harvest, owner_id: member.id
```

This seems harder to read than putting it in the controller.

**Comment by Br3nda** [comment]
can :create,  Harvest do |harvest|
      if harvest.planting
        harvest.planting.owner_id == member.id && harvest.owner_id == member.id
      else
        true
      end
    end
    can :update, Harvest do |harvest|
      if harvest.planting
        harvest.planting.owner_id == member.id && harvest.owner_id == member.id
      else
        harvest.owner_id == member.id
      end
    end
    can :destroy, Harvest, owner_id: member.id

File: app/models/ability.rb [hunk 4]
```diff
@@ -92,6 +92,8 @@ def initialize(member) # rubocop:disable Metrics/AbcSize
     can :create,  Harvest
     can :update,  Harvest, owner_id: member.id
     can :destroy, Harvest, owner_id: member.id
+    can :update,  Harvest, owner_id: member.id, planting: { owner_id: member.id }
+    can :destroy, Harvest, owner_id: member.id, planting: { owner_id: member.id }
```
Br3nda: I'm really not sure this is correct. But the tests pass.

app/models/harvest.rb
```diff
@@ -7,6 +7,7 @@ class Harvest < ActiveRecord::Base
   belongs_to :crop
```
pozorvlak: Do we have any guarantee that `harvest.crop == harvest.planting.crop` where both are defined?

app/models/ability.rb
pozorvlak: AFAICT the tests cover all the cases of interest (no planting specified, planting owned by member, planting not owned by member), so I think we're good.

app/models/harvest.rb
Br3nda: No, but I think this is the least worst option with our current datamodel. 

I do question, long term, why track harvests of thing you do to have planted in your garden?

maco:
I can only guess something about how you might harvest dandelion leaves for
salad when you didn't plant dandelions and they just happened to show up in
your garden/lawn.

app/models/harvest.rb
Br3nda: Later we could say "you're harvesting something you never planted. Would you like to create a planting?"

Br3nda: We could delegate havest.crop to the planting when present.

Br3nda:
I think we addressed everything....  Are we good to merge?

pozorvlak:
I'm now +1 on this: @CloCkWeRX, are you satisfied?

CloCkWeRX:
Merge away

Br3nda:
YAY! i can see my harvests on my plantings.
This is gonna be very useful
http://www.growstuff.org/plantings/shiny-vege-gardens-blueberry

PR #1133, opened 2016-12-20
Refactor test-runners

pozorvlak:
I'm aiming for something along the lines laid out in #1125; the idea is that it should be as easy as possible to duplicate a CI run locally.

- [X] factor out duplicated env vars in .travis.yml
- [x] create a script to invoke all static checks
- [x] create a new build-matrix entry that runs only static checks
- [x] make `rake` run specs and static checks
- [x] add a `spec:unit` task that runs non-feature tests
- [x] run JavaScript tests with rake
- [x] run JavaScript tests from Travis
- [x] tag tests that need ElasticSearch
- [x] split out elastic and inelastic tests into separate tasks
- [x] only run elastic tests under GROWSTUFF_ELASTICSEARCH=true?
- [x] put test-running instructions in CONTRIBUTING.md

PR #1143, opened 2017-01-04
[WIP] bar graphs

pozorvlak:
I also think #688 is awesome, so here it is with the codeclimate problems fixed, hopefully. Currently mostly untested.

**Edit**: OK, I've now done some very preliminary testing, and there are two problems with this that I can see.

1. I can't work out how to run the JS unit tests. I'd been thinking of rebasing #1133 on top of this and using `jasmine` rather than `jasmine-rails`, but the [test-running instructions for Jasmine](https://github.com/jasmine/jasmine-gem/blob/master/README.markdown) aren't working for me on this code, so maybe it should be the other way round.
2. The actual data appears to be hardcoded (see screenshot). We should fix that before merging and deploying this :-) **Edit** - fixed in 7fa0d22.

screenshot from 2017-01-05 11-08-55


pozorvlak:
Huzzah!

screenshot from 2017-01-09 15-13-41

Code rather horrible and inefficient, though - see  7fa0d22.

cesy:
Looking exciting, but needs a rebase/dev merge before it can go in, from the look of things.

Br3nda:
This looks mergable. Is it still WIP?

CloCkWeRX:
I think there was "hook it up to real data" remaining.

pozorvlak:
I've hooked it up to real data, albeit in a horrible and inefficient way. I wanted to write more unit tests, but I can do that in a later PR if you're OK with merging this as-is.

CloCkWeRX:
Then lets merge I reckon!
(after tests pass)

Issue #1128, opened 2016-12-14
crop#create dies in sys.exit


Br3nda:
on growstuff.org, and in git dev branch:

at line 140 of crops_controller:
```
          Role.crop_wranglers.each do |w|
            Notifier.new_crop_request(w, @crop).deliver_later!
          end
```
`deliver_later!` doesn't return. it exits the ruby process. the web container spits a http 500.
I imagine only one crop wrangler will be getting the emails right now.
This may have come in at a gem upgrade recently.

p.s. would love someone to approve "boysenberry".

CloCkWeRX:
We can maybe swap to ```deliver_now!``` to rule out sidekiq/activejob/etc configuration without much effort.  If its sidekiq, its probably fe137a11bb363d6332b5c55e603bb870930e666a

PR #921, opened 2016-05-26
Use standardized & translateable strings for the photo show page

maco:
Fixes #876

PR #1145, opened 2017-01-06
Descriptive links from photo page

Br3nda:
The links from photos page don't made descriptions. see example here:
http://www.growstuff.org/photos/756
There are multiple links called:
"
    A harvest by shiny
    A harvest by shiny
    A harvest by shiny
    A harvest by shiny
    A harvest by shiny
    A harvest by shiny
"

This PR add description titles to the links, so you can tell them apart

BEFORE:
image

AFTER:
image

It also add links to seeds.

The `if` isn't necessary.

### DIFF
--- a/app/models/planting.rb
+++ b/app/models/planting.rb
@@ -64,7 +64,15 @@ class Planting < ActiveRecord::Base
   end
 
   def planting_slug
-    "#{owner.login_name}-#{garden.name}-#{crop.name}".downcase.gsub(' ', '-')
+    if garden.present? && crop.present?
+      "#{owner.login_name}-#{garden.name}-#{crop.name}"
+    elsif garden.present?
+      "#{owner.login_name}-#{garden.name}-null"
+    elsif crop.present?
+      "#{owner.login_name}-null-#{crop.name}"
+    else
+      "#{owner.login_name}-null-null"
+    end.downcase.gsub(' ', '-')
   end
 
   # location = garden owner + garden name, i.e. "Skud's backyard"
--- a/app/models/planting.rb
+++ b/app/models/planting.rb
@@ -6,6 +6,7 @@ class Planting < ActiveRecord::Base
   belongs_to :garden
   belongs_to :owner, class_name: 'Member', counter_cache: true
   belongs_to :crop, counter_cache: true
+  has_many :harvests, -> { order(harvested_at: :desc) }, dependent: :destroy
 
   default_scope { order("created_at desc") }
   scope :finished, -> { where(finished: true) }
--- a/app/models/planting.rb
+++ b/app/models/planting.rb
@@ -73,7 +73,7 @@ class Planting < ActiveRecord::Base
 
   # stringify as "beet in Skud's backyard" or similar
   def to_s
-    self.crop_name + " in " + self.location
+    I18n.t('plantings.string', crop: crop.name, garden: garden.name, owner: owner)
   end
 
   def default_photo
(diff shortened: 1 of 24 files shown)


pozorvlak:
This is basically a revert of #921, which aimed to make the link titles translatable. I think your screenshot makes a good argument that having the same link title for every harvest/garden/planting is bad, so I'd be +1 to this once you've fixed up the failing tests (possibly by just reverting 58f4ae32). But down the road, we should think if we can achieve translatability without losing so much information. The `to_s` methods of `Harvest` and `Planting` currently do a lot of ad-hoc string munging which deeply assumes we're outputting English :-(

Br3nda:
I'd greatly like translations back. I'll rig this up to preserve translations.

Br3nda:
I've updated to use I18n. I left the harvest untranslated for now. It has a lot going on with weights and quantity. It can be in a seperate PR.

Br3nda:
don't merge yet. it's getting "uninitialized constant Growstuff::Constants" on a freshly restarted rails.

Br3nda:
I fixed the Garden::constants error with 88b1f02574cbf9305098b48a494077d656ccbd07. Not sure why this happened.

maco:
That file is how we define @photo.crop and @photo.seed etc, and I guess
crop.rb and photo.rb are loading in the opposite order of how they do when
we attach a photo to a crop.

>

Br3nda:
I'm not sure why the last test is failing. It's during new crop submission, which is broken in dev and prod too. On dev and prod it sys.exit(), but on test it doesn't, so this is a second error.

pozorvlak:
I think the test failure indicates a genuine bug in this PR - when I check out this branch and try to create a planting with an unknown crop, I get the following:

screenshot from 2017-01-09 13-07-49

On the current dev branch, I get

screenshot from 2017-01-09 13-08-19

I'm not sure what you mean by "new crop submission... is broken in dev and prod": I just created a new crop on http://staging.growstuff.org, and got the following:

screenshot from 2017-01-09 13-10-53

...which looks fine to me.

app/models/planting.rb
```diff
@@ -64,17 +64,17 @@ def finished_must_be_after_planted
   end
 
   def planting_slug
-    "#{owner.login_name}-#{garden}-#{crop}".downcase.gsub(' ', '-')
+    "#{owner.login_name}-#{garden.name}-#{crop.name}".downcase.gsub(' ', '-')
```
pozorvlak: OK, I see what's going on - the test is now breaking because we're looking up the `name` field on a nonexistent crop. Previously, we'd create a slug `pozorvlak-allotment-nil`, which was fine, and then we'd throw it away because the planting was invalid. So I guess we need to move the test to before the slug is generated, or make the slug-generation more robust.

Br3nda:
The broken crop suggestion is on #1128

pozorvlak:
@Br3nda: cool, thanks. Still, I don't think that's related to the test failure here.

Br3nda:
I've reproduced the old behaviour in eba397117c7ac69643a0b79af7a9fb5842b0e289
It will still produce errors if we try to make a planting slug before we assign an owner.

We could chase down why we are creating slugs when the planting isn't valid yet -- but that would be a bigger scope than this PR.

app/models/planting.rb
```diff
@@ -64,7 +64,15 @@ def finished_must_be_after_planted
   end
 
   def planting_slug
-    "#{owner.login_name}-#{garden.name}-#{crop.name}".downcase.gsub(' ', '-')
+    if garden.present? && crop.present?
```
pozorvlak: I'd have written something like

```
garden_name = garden.present? ? garden.name : "null"
crop_name = crop.present? ? crop.name : "null"
"#{owner.login_name}-#{garden_name}-#{crop_name}".downcase.gsub(' ', '-')
```

but this is fine.

PR #1147, opened 2017-01-08
Bundle alphabetical order to comply with stricter rubocop

Br3nda:
Same gems, but each cluster is sorted


Br3nda:
"The command "script/gemfile_check" exited with 1."

otherwise green

PR #1150, opened 2017-01-14
Rubocop fix up for the models 

Br3nda:


app/models/crop.rb
```diff
@@ -103,18 +103,19 @@ class Crop < ActiveRecord::Base
     end
   end
 
-  def as_indexed_json(options = {})
-    self.as_json(
+  def as_indexed_json(_options = {})
```
Br3nda: note unused variable here

app/models/crop.rb
```diff
@@ -103,18 +103,19 @@ class Crop < ActiveRecord::Base
...
         alternate_names: { only: :name }
-      })
+      }
+    )
   end
 
   # update the Elasticsearch index (only if we're using it in this
   # environment)
-  def update_index(name_obj)
+  def update_index(_name_obj)
```
Br3nda: note unused variable here

Br3nda:
I'll leave the planting model for later, as there is a PR that will conflict with it.

PR #968, opened 2016-06-04
 like feature for posts

CloCkWeRX:
Was #696 

Screenshots:
like_ui
like_ui2

PR #1153, opened 2017-01-14
Release 18

cesy:
Since we have a pile of good changes, let's get some live.


cesy:
Needs an approval from someone other than me before we can merge and deploy.

Br3nda:
It would be good to get #968 in here too

cesy:
We'll do that in the next release just to get this live, and let's keep trying to do releases more often.

PR #1155, opened 2017-01-21
2017 01 21 gem updates

CloCkWeRX:
Updates to rubocop, others; plus pinning us to certain gems to avoid installation problems.


CloCkWeRX:
https://travis-ci.org/Growstuff/growstuff/jobs/193959684 was flaky, which was interesting; but since it seems to be pretty low occurrence I opted to ignore it.

PR #1158, opened 2017-01-22
more controller translations

Br3nda:


Br3nda:
Codeclimate is complain of "Similar code found in 1 other location"
I'll give it a tackle.

PR #1159, opened 2017-01-22
Fix display of gardens on smaller screens

Br3nda:
before 
image

after
image


Br3nda:
I went looking for why this is here, and found commit message of:
```
    switch from less to sass
    * dependency hell + bit rot on upstream libraries  prevents installation on OS X
```
on 8659ebca2dae32ce42a61be60d06775f776b3251

Issue #1139, opened 2017-01-03
"Follow" member broken on growstuff.org


Br3nda:
![image](https://cloud.githubusercontent.com/assets/12266/21602111/8b1f8a16-d1f5-11e6-9b19-29520da2244b.png)

Br3nda:
On dev i get this

2017-01-03T07:49:33.818Z 10645 TID-gp19st2nw ERROR: Your Redis configuration uses the namespace 'app3_sidekiq_development' but the redis-namespace gem is not included in the Gemfile.Add the gem to your Gemfile to continue using a namespace. Otherwise, remove the namespace parameter.
   (0.2ms)  ROLLBACK
Completed 500 Internal Server Error in 77ms (ActiveRecord: 4.5ms)

PR #1160, opened 2017-01-22
change deliver_later to deliver_now, so we don't sys.exit() crash

Br3nda:
Production (and my dev) is crashing out on a sys.exit at:
* new crop request / approval
* following another user
* planting reminders

this changes to deliver_now


Br3nda:
also fixes #1128

Br3nda:
and fixes #1139

PR #1168, opened 2017-01-24
Reviving Likeable

Br3nda:
This looks ready after a year of good work. 

Was #968


CloCkWeRX:
Closing the other PR, I think there's a few other tiny bits of feedback I completely missed.

Br3nda:
some haml-lint, method length, and comments about todo/hack could be gone. But i'd prefer to merge this rather than wait

PR #1174, opened 2017-02-01
rubocop - reduce complexity

Br3nda:


app/controllers/seeds_controller.rb
```diff
@@ -109,4 +103,14 @@ def seed_params
       :days_until_maturity_min, :days_until_maturity_max, :organic, :gmo,
       :heirloom, :tradable_to, :slug)
   end
+
+  def seeds
+    if @owner
```
pozorvlak: I'd be tempted to further simplify this into

```
s =
  if @owner then @owner.seeds
  elsif @crop then @crop.seeds
  else Seed end
s.includes(:owner, :crop).paginate(page: params[:page])
```

app/models/harvest.rb
```diff
@@ -82,32 +82,38 @@ def harvest_slug
...
-      string += "#{number_to_human(quantity.to_s, strip_insignificant_zeros: true)} "
-      string += if unit == 'individual'
-                  'individual '
-                elsif quantity == 1
-                  "#{unit} of "
-                else
-                  "#{unit.pluralize} of "
-                end
-    end
+    "#{quantity_to_human} #{unit_to_human} #{crop_name_to_human} #{weight_to_human}".strip
```
pozorvlak: This is *much* better. Awesome!

app/models/member.rb
```diff
@@ -221,17 +221,10 @@ def self.nearest_to(place)
...
-      newsletter_subscribe
-    elsif confirmed_at # i.e. after member's confirmed their account
-      if newsletter_changed? # edited member settings
-        if newsletter
-          newsletter_subscribe
-        else
-          newsletter_unsubscribe
-        end
-      end
+    if newsletter
```
pozorvlak: This is shorter, but I find it a lot harder to follow than the previous version. Was the problem that rubocop was complaining about the if-clauses being nested too deeply? We could easily trim one level by changing L226 to `if confirmed_at && newsletter_changed?`.

app/models/member.rb
```diff
@@ -221,17 +221,10 @@ def self.nearest_to(place)
...
+      newsletter_subscribe confirmed_at_changed? || confirmed_at && newsletter_changed?
     end
+    newsletter_unsubscribe if confirmed_at && newsletter_changed? && !newsletter
```
pozorvlak: You could put this in an `else`-clause to the previous `if` and avoid checking `newsletter` again.

app/models/member.rb
```diff
@@ -221,17 +221,10 @@ def self.nearest_to(place)
...
-    elsif confirmed_at # i.e. after member's confirmed their account
-      if newsletter_changed? # edited member settings
-        if newsletter
-          newsletter_subscribe
-        else
-          newsletter_unsubscribe
-        end
-      end
+    if newsletter
+      newsletter_subscribe confirmed_at_changed? || confirmed_at && newsletter_changed?
```
pozorvlak: Should there be an `if` between `newsletter_subscribe` and `confirmed_at_changed`? The first argument of `newsletter_subscribe` shouldn't be a boolean.

I'd also prefer brackets in the condition to make the precedence clear.

app/models/planting.rb
```diff
@@ -112,6 +112,20 @@ def planted?(current_date = Date.current)
     planted_at.present? && current_date.to_date >= planted_at
   end
 
+  def finished_at_date_passed?
```
pozorvlak: You're right, this code definitely belongs on the model.

app/models/member.rb
Br3nda: Wow. Yes. I'll add more tests too here

Br3nda:
I'm gonna split these out into PRs

app/models/harvest.rb
Br3nda: I've moved this into it's own pr here  #1177

app/controllers/seeds_controller.rb
Br3nda: i've done this, and removed the `s` variable -- it's on PR  #1176

Br3nda:
I'm gonna rebase this -- hoping the comments don't get lost by github

PR #1175, opened 2017-02-01
Reduce complexity of plantings controller

Br3nda:
the code was writing to @plantings over and over, and then returning it to be written into @plantings again.

this is better

### DIFF
--- a/app/controllers/plantings_controller.rb
+++ b/app/controllers/plantings_controller.rb
@@ -124,16 +124,14 @@ class PlantingsController < ApplicationController
   end
 
   def plantings
-    @plantings = if @owner
-                   @owner.plantings
-                 elsif @crop
-                   @crop.plantings
-                 else
-                   Planting
-                 end
-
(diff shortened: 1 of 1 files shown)


app/controllers/plantings_controller.rb
```diff
@@ -124,16 +124,14 @@ def update_days_before_maturity(planting, crop_id)
...
-                   @owner.plantings
-                 elsif @crop
-                   @crop.plantings
-                 else
-                   Planting
-                 end
-
-    @plantings = @plantings.current unless @show_all
-    @plantings = @plantings.includes(:owner, :crop, :garden).order(:created_at).paginate(page: params[:page])
-    @plantings
```
CloCkWeRX: This set up a bit of shared state; do any views or similar refer to ```@plantings``` that could have accidentally relied on this?

Br3nda: none would rely on that behaviour, because i made this method only a few days ago in another PR.

PR #1176, opened 2017-02-01
Reduce complexity of seeds controller

Br3nda:


app/controllers/seeds_controller.rb
```diff
@@ -109,4 +103,14 @@ def seed_params
       :days_until_maturity_min, :days_until_maturity_max, :organic, :gmo,
       :heirloom, :tradable_to, :slug)
   end
+
+  def seeds
```
CloCkWeRX: Might be clearer to pass these as arguments (owner = nil, crop = nil), or shift this onto the model as something "seeds_by_topic"

Br3nda: @CloCkWeRX updated

pozorvlak:
```
Failures:

  1) forums as an admin user editing forum

     Failure/Error: expect(current_path).to eq forum_path(forum)

     

       expected: "/forums/permaculture"

            got: "/forums/permaculture/edit"

     

       (compared using ==)

     # ./spec/features/admin/forums_spec.rb:52:in `block (3 levels) in <top (required)>'
```

on one build but not the other. Page reloading too slowly? I'm re-running the build.

PR #1178, opened 2017-02-02
PlantingsHelper simplification

Br3nda:


app/helpers/plantings_helper.rb
```diff
@@ -1,14 +1,19 @@
...
+    return "0" if planting.finished_at.present? && planting.finished_at_date_passed?
+
+    # days until finished_at
+    return planting.days_until_finished.to_s if planting.finished_at.present?
+
+    return "unknown" if planting.planted_at.nil? || planting.days_before_maturity.nil?
+
+    return "0" if planting.days_until_mature <= 0
+
+    planting.days_until_mature.to_s
```
Br3nda: i think i can make this much clearer....  more changes coming

app/models/planting.rb
```diff
@@ -112,6 +112,21 @@ def planted?(current_date = Date.current)
     planted_at.present? && current_date.to_date >= planted_at
   end
 
+  def finished_at_date_passed?
```
pozorvlak: This method doesn't seem to be used anywhere.

app/models/planting.rb
```diff
@@ -112,6 +112,21 @@ def planted?(current_date = Date.current)
...
+    finished_at.present? && (finished_at - Date.current).to_i <= 0
+  end
+
+  def days_until_finished
+    return 0 if finished?
```
pozorvlak: Should this be `if finished_at_date_passed?`?

app/models/planting.rb
Br3nda: You're right, i factored it out in the end. I will remove

app/models/planting.rb
Br3nda: No, it's a boolean column on that table of the database. 
```
    t.boolean  "finished",             default: false
```
So, this is checking, is that set to true -- which seems to be yet another way a planting record can be finished.

app/models/planting.rb
Br3nda: gone

app/models/planting.rb
pozorvlak: Oh dear, that's not very DRY :-(

PR #1181, opened 2017-02-03
Te grand haml cleanup of 2017

Br3nda:


Br3nda:
To make this feasible to review, I'll stop here and continue on another PR later.

PR #1184, opened 2017-02-05
Haml clean up -- crops views

Br3nda:


Br3nda:
Code climate found an unescaped model attribute. It's a calculated one (plant_part frequency) so it's safe, but i've cleaned up that code anyway

app/models/crop.rb
```diff
@@ -167,7 +167,7 @@ def popular_plant_parts
     harvests.each do |h|
       popular_plant_parts[h.plant_part] += 1 if h.plant_part
     end
-    popular_plant_parts
+    popular_plant_parts.sort_by { |_s, freq| freq }.reverse
```
CloCkWeRX: I think this is is what upset the tests; not really sure if they are stale expectations or legit regression

Br3nda: The method is only being called in one place, so it's easy to fix up.

app/models/crop.rb
```diff
@@ -163,11 +163,11 @@ def planted_from
   # key: plant part (eg. 'fruit')
   # value: count of how many times it's been used by harvests
   def popular_plant_parts
-    popular_plant_parts = Hash.new(0)
-    harvests.each do |h|
-      popular_plant_parts[h.plant_part] += 1 if h.plant_part
-    end
-    popular_plant_parts
```
Br3nda: I've changed the structure of what it returns -- but i think it's better now because it's doing the group/count and ordering in the database.
See: 8408fdc845b98721366bb678e1b74e65bf9cdeaa

Issue #1182, opened 2017-02-04
The tests fail if i run them at lunch time


Br3nda:
The code isn't handling timezones correctly. I'm in UTC+13. If I run the specs between 12pm and 1pm, I get failures like this:

  1) PlantingsController GET new sets the date of the planting to today
     Failure/Error: assigns(:planting).planted_at.should == Date.today
     
       expected: February 04, 2017
            got: February 03, 2017 (using ==)
       Diff:
       @@ -1,2 +1,2 @@
       -February 04, 2017
       +February 03, 2017

CloCkWeRX:
Expectation seems wrong, the controller is doing:
```'planted_at' => Time.zone.today```

PR #1199, opened 2017-02-10
Use timezone when creating dates

Br3nda:


Br3nda:
I suspect this will help with #1182

PR #1200, opened 2017-02-10
Simplified logic in planting.planting_slug

Br3nda:


Br3nda:
How about we use the  primary key id instead of this "null" strings?

pozorvlak:
@Br3nda good idea, though I'm wondering when the "null" versions will ever be used - is it possible to create a planting without an owner, garden or crop?

PR #1202, opened 2017-02-11
Simplified alternate names controller.

Br3nda:
Move respond to.
Removed some json responders that aren't being used


maco:
Oh, thank goodness. I was sure there was a way to get rid of all that `respond to format` stuff

Br3nda:
so -- i'll chip away and get the rest of the controllers doing that then...

Br3nda:
YAY 
image

PR #1203, opened 2017-02-11
Added cancan ability for likes (and DRY-ed)

Br3nda:


Br3nda:
Actually this fixes destroy -- but breaks create.

PR #1206, opened 2017-02-12
don't allow owner in params

Br3nda:
because when you create/update a garden , you don't get to choose who the owner is


maco:
do we need a test for this?

Br3nda:
There's good coverage of this scenario in there ready

PR #1204, opened 2017-02-11
Bundle Update on 2017-02-12

deppbot:
**Updated RubyGems:**

- Updated: [active_utils](http://github.com/shopify/active_utils), [3.2.3...3.2.5](https://github.com/shopify/active_utils/compare/v3.2.3...v3.2.5) ([CHANGELOG](https://github.com/Shopify/active_utils/blob/master/CHANGELOG.md))
- Updated: [activemerchant](http://github.com/activemerchant/active_merchant), [1.61.0...1.63.0](https://github.com/activemerchant/active_merchant/compare/v1.61.0...v1.63.0) ([CHANGELOG](https://github.com/activemerchant/active_merchant/blob/master/CHANGELOG))
- Updated: [autoprefixer-rails](https://github.com/ai/autoprefixer-rails), [6.6.1...6.7.2](https://github.com/ai/autoprefixer-rails/compare/6.6.1...6.7.2) ([CHANGELOG](https://github.com/ai/autoprefixer-rails/blob/master/CHANGELOG.md))
- Updated: [bonsai-elasticsearch-rails](https://github.com/omc/bonsai-elasticsearch-rails), [0.0.4...0.2.0](https://github.com/omc/bonsai-elasticsearch-rails/compare/v0.0.4...v0.2.0)
- Updated: [bootstrap_form](http://github.com/bootstrap-ruby/rails-bootstrap-forms), [2.5.3...2.6.0](https://github.com/bootstrap-ruby/rails-bootstrap-forms/compare/v2.5.3...v2.6.0) ([CHANGELOG](https://github.com/bootstrap-ruby/rails-bootstrap-forms/blob/master/CHANGELOG.md))
- Updated: [cancancan](https://github.com/CanCanCommunity/cancancan), [1.15.0...1.16.0](https://github.com/CanCanCommunity/cancancan/compare/v1.15.0...v1.16.0) ([CHANGELOG](https://github.com/CanCanCommunity/cancancan/blob/develop/CHANGELOG.rdoc))
- Updated: [capybara](http://github.com/jnicklas/capybara), [2.10.1...2.12.0](https://github.com/jnicklas/capybara/compare/2.10.1...2.12.0) ([CHANGELOG](https://github.com/jnicklas/capybara/blob/master/History.md))
- Updated: [childprocess](http://github.com/jarib/childprocess), [0.5.9...0.6.1](https://github.com/jarib/childprocess/compare/v0.5.9...v0.6.1)
[...]

PR #1209, opened 2017-02-13
Bundle update 2012 02 13

CloCkWeRX:
Was  #1204 

This pins us to a version of hashie < 3.5, whiich we should revisit when 3.5.3+ is out; and leaflet-rails 0.7.0ish, which breaks selenium tests.

Still stuff to upgrade, but later


CloCkWeRX:
Weird the forum specs broke...

PR #1212, opened 2017-02-13
DRY up our controller

Br3nda:


Br3nda:
i just worked out why this project has .html.haml instead of .haml

So -- don't merge this -- it will work but it will make errors in the console about not finding views.

PR #1213, opened 2017-02-13
use symbols for status codes

maco:
I didn't want to hold up @Br3nda 's PR on a nitpick like this


maco:
On the other specs we're using numbers. I think we'd have to check `status`
instead of `status.code` if we were going to match the symbol.

PR #1214, opened 2017-02-13
User can only sign out their own session

Br3nda:


Br3nda:
actualy.... we don't need to reload it, cancancan has already.

PR #1217, opened 2017-02-15
Fix up to match rubocop's MultilineMethodCallBraceLayout recommendation

Br3nda:


maco:
@pozorvlak I think you and I said we don't like how long lines get when you
do #1 and so prefer #2?

Br3nda:
It is.... curious. I can rubocop -a to autofix it, and rubocop has swapped the location all over the place.
I'm not sure why at all

Br3nda:
As long as we're consistent, it's fine. There might be another cop that's going to pull it all to # 1 or # 2 soon as we get it off the rubocop todo list

pozorvlak:
Yeah, I think it's half-applying style 1 if the first argument is on the same line as the opening paren (and thus moving the closing paren up), and half-applying style 2 if the first argument is on a new line (and thus moving the closing paren down). Applying a "multiline method indentation" cop will consistently apply one or the other style. In which case, it probably makes more sense to apply both cops at once.

My preference is for style 2, but I think consistency is more important than which style we adopt.

Br3nda:
My preference is nearly always whatever the rubocop default says, do I don't need to remember special rules for any project as I type.

Br3nda:
I'll turn on more cops and push here.

Br3nda:
seems more consistent now.
the only odd one out is authentications.. which is chained methods calls like this

```
      @authentication = current_member.authentications
        .create_with(
          name: name,
          token: auth['credentials']['token'],
          secret: auth['credentials']['secret']
        )
        .find_or_create_by(
          provider: auth['provider'],
          uid: auth['uid'],
          name: name
        )
```

Br3nda:
Seems more consistent now. the only odd one out is authentications, which looks like this
```
      @authentication = current_member.authentications
        .create_with(
          name: name,
          token: auth['credentials']['token'],
          secret: auth['credentials']['secret']
        )
        .find_or_create_by(
          provider: auth['provider'],
          uid: auth['uid'],
          name: name
        )
```

PR #1222, opened 2017-02-16
Small clean ups on forums controller

Br3nda:


Br3nda:
specs are passing, but codeclimate is finding more duplicate code each time i change something.

Br3nda:
Codeclimate is claiming this causes duplicate code between posts and plant parts.
neither files are changed in here -- so nope. Codeclimate you're dreaming.

pozorvlak:
I think what's happening there is that those duplications have been removed in the `dev` branch, and CodeClimate is comparing this to `dev` rather than to the most recent common ancestor. So if I update the branch with the latest `dev` changes, CodeClimate should be happy again. And in fact it's just gone green as I was typing this :-)

app/controllers/forums_controller.rb
```diff
@@ -1,86 +1,51 @@
...
-      format.json { render json: @forum }
-    end
+    respond_with(@forum)
   end
 
   # GET /forums/1/edit
-  def edit
-    @forum = Forum.find(params[:id])
-  end
+  def edit; end
```
pozorvlak: Do we need to define this at all?

PR #1225, opened 2017-02-16
DRY comments controller

Br3nda:
I also added more tests for when trying to delete or edit a comment you didn't write.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,17 +1,15 @@
 class CommentsController < ApplicationController
   before_action :authenticate_member!, except: [:index, :show]
   load_and_authorize_resource
+  respond_to :html, :json
+  respond_to :rss, only: :index
 
   # GET /comments
   # GET /comments.json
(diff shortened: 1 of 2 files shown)


app/controllers/comments_controller.rb
```diff
@@ -40,54 +35,30 @@ def edit
...
-      format.json { head :no_content }
-    end
+    respond_with(@post)
   end
 
   private
 
   def comment_params
-    params.require(:comment).permit(:author_id, :body, :post_id)
+    params.require(:comment).permit(:body, :post_id)
```
Br3nda: note: you can't set the author_id ever.
the old code was allowing the params, and the deleting it (see line 62 of old half of diff)

Br3nda:
Code climate is complaining that the photos controller is similar to the scientific names controller
neither files were changed in this PR. :-/

app/controllers/comments_controller.rb
pozorvlak: D:

PR #1227, opened 2017-02-18
Shrinking the giant crop model code

Br3nda:
I've pulled the csv import out to CsvImporter class
It's only called from db:seeds and a rake task.

### DIFF
--- a/app/models/crop.rb
+++ b/app/models/crop.rb
@@ -211,68 +211,6 @@ class Crop < ActiveRecord::Base
     interesting_crops
   end
 
-  # Crop.create_from_csv(row)
-  # used by db/seeds.rb and rake growstuff:import_crops
-  # CSV fields:
-  # - name (required)
-  # - en_wikipedia_url (required)
-  # - parent (name, optional)
-  # - scientific name (optional, can be picked up from parent if it has one)
-
-  def self.create_from_csv(row)
-    name, en_wikipedia_url, parent, scientific_names, alternate_names = row
-
-    cropbot = Member.find_by(login_name: 'cropbot')
-    raise "cropbot account not found: run rake db:seed" unless cropbot
-
-    crop = Crop.find_or_create_by(name: name)
-    crop.update_attributes(
-      en_wikipedia_url: en_wikipedia_url,
-      creator_id: cropbot.id
-    )
-
-    if parent
-      parent = Crop.find_by(name: parent)
-      if parent
-        crop.update_attributes(parent_id: parent.id)
-      else
-        logger.warn("Warning: parent crop #{parent} not found")
-      end
-    end
-
-    crop.add_scientific_names_from_csv(scientific_names)
-    crop.add_alternate_names_from_csv(alternate_names)
-  end
-
-  def add_scientific_names_from_csv(scientific_names)
-    names_to_add = []
-    if !scientific_names.blank? # i.e. we actually passed something in, which isn't a given
-      names_to_add = scientific_names.split(/,\s*/)
-    elsif parent && !parent.scientific_names.empty? # pick up from parent
-      names_to_add = parent.scientific_names.map(&:name)
-    else
-      logger.warn("Warning: no scientific name (not even on parent crop) for #{self}")
-    end
-
-    cropbot = Member.find_by(login_name: 'cropbot')
-
-    return if names_to_add.empty?
-    raise "cropbot account not found: run rake db:seed" unless cropbot
-
-    add_names_to_list(names_to_add, 'scientific')
-  end
-
-  def add_alternate_names_from_csv(alternate_names)
-    # i.e. we actually passed something in, which isn't a given
-    return if alternate_names.blank?
-
-    cropbot = Member.find_by!(login_name: 'cropbot')
-    names_to_add = alternate_names.split(/,\s*/)
-    add_names_to_list(names_to_add, 'alternate')
-  rescue
-    raise "cropbot account not found: run rake db:seed" unless cropbot
-  end
-
   def rejection_explanation
     return rejection_notes if reason_for_rejection == "other"
     reason_for_rejection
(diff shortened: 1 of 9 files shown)


maco:
Thanks, Shiny. I took a stab at this before and got lost amid all the selfs.

Br3nda:
Cool. I'll update the specs soon.

Br3nda:
@maco I got post too.... And ended up rewriting the  bit I didn't understand.

db/seeds.rb
```diff
@@ -31,7 +31,7 @@ def load_crops
   Dir.glob("#{source_path}/crops*.csv").each do |crop_file|
     puts "Loading crops from #{crop_file}..."
```
pozorvlak: I'd be inclined to move this whole for-loop into `CsvImporter.import_file`, but no biggie.

spec/models/crop_spec.rb
```diff
@@ -419,164 +419,106 @@
...
-        tomato = FactoryGirl.create(:tomato, parent: parent)
         expect(tomato.parent).to eq parent
         expect(tomato.parent.default_scientific_name).to eq "Parentis cropis"
-
-        tomato.add_scientific_names_from_csv('')
         expect(tomato.default_scientific_name).to eq "Parentis cropis"
       end
 
-      it "doesn't add a duplicate scientific name" do
-        tomato = FactoryGirl.create(:tomato)
```
pozorvlak: It looks like we're no longer testing that adding a (scientific|alternate) name is idempotent; this makes me nervous. How about a couple of tests along the following lines?

```ruby
it "doesn't add a duplicate scientific name" do
  row = ["parent", "http://en.wikipedia.org/wiki/Parent", "", "Foo bar, Foo bar"]
  tomato = CsvImporter.new.import_crop(row)
  expect(tomato.scientific_names.size).to eq 1
end

it "doesn't add a duplicate scientific name from parent" do
  parent = CsvImporter.new.import_crop(
    ["parent", "http://en.wikipedia.org/wiki/Parent", "", "Parentis cropis"])
  tomato = CsvImporter.new.import_crop(
    ["Tomato", "http://en.wikipedia.org/wiki/Parent", "parent". "Parentis cropis"])
  expect(tomato.scientific_names.size).to eq 1
end
```

Br3nda:
ooo.. there are more rubocops turned on in dev - so now this failing.

will sort, and add specs soon.

spec/models/crop_spec.rb
Br3nda: added

pozorvlak:
Looks great apart from the new CodeClimate issue. I guess you can factor out a common method and pass in the model class as an argument?

Br3nda:
There is now similar code in adding a ScientificName and AlternateName.

The old code had made that down to 3 re-useable methods. However that was more code than the original - and i think part of what made the old code confusing.

Can we keep this duplicate code because, for now, it's clearer to read.

app/models/crop.rb
```diff
@@ -325,29 +263,6 @@ def self.case_insensitive_name(name)
...
-  def add_names_to_list(names_to_add, list_name)
-    names_to_add.each do |n|
-      if name_already_exists(list_name, n)
-        logger.warn("Warning: skipping duplicate #{list_name} name #{n} for #{self}")
-      else
-        create_crop_in_list(list_name, n)
-      end
-    end
-  end
-
```
Br3nda: below here iss the old way, that reduced the duplicate code.

app/models/crop.rb
```diff
@@ -325,29 +263,6 @@ def self.case_insensitive_name(name)
 
   private
```
Br3nda: below here iss the old way, that reduced the duplicate code.

spec/models/crop_spec.rb
```diff
@@ -419,164 +419,135 @@
...
+        expect(tomato.alternate_names[3].name).to eq "Quux"
+      end
+
+      it "doesn't add a duplicate alternate name" do
+        row = ["tomato", "http://en.wikipedia.org/wiki/tomato", "", "", "Foo"]
+        tomato = CsvImporter.new.import_crop(row)
+        row = ["tomoto", "http://en.wikipedia.org/wiki/tomoto", "", "", "Foo"]
+        tomoto = CsvImporter.new.import_crop(row)
+        expect(tomato.alternate_names.size).to eq 1
+        expect(tomoto.alternate_names.size).to eq 0
```
pozorvlak: Sorry, everyone, I screwed up by approving this. We require that Crop A can have AlternateName N at most once; we don't require that Crop A and Crop B can't share an AlternateName. The test is failing because it's not testing for the desired behaviour. I'll submit a PR fixing the test.

pozorvlak: Wait, no, we actually assume in the code that every AlternateName belongs to exactly one Crop. So what's happening here is that the second crop grabs the AlternateName from its former owner. We probably don't want to allow that.

pozorvlak: And we also seem to insist that each ScientificName belongs to exactly one Crop? That doesn't sound right.

pozorvlak:
```
2.3.3 :027 > ScientificName.where(name: 'Cucurbita maxima').count
   (0.3ms)  SELECT COUNT(*) FROM "scientific_names" WHERE "scientific_names"."name" = $1  [["name", "Cucurbita maxima"]]
 => 4
```

Riiiiiight. So before we were creating duplicate `AlternateName` and `ScientificName` objects for each (crop, name) pair. So it was possible for two crops to have the same AlternateName _as a string_, but they were represented in the DB by different objects. So this PR actually changes the importer's behaviour, by disallowing two crops from having string-equal scientific/alternate names.

Br3nda:
DB schema only allows an alt name or sci name to belong to one crop. So.... We want duplicates? This seems messy.

maco:
Broccoli, cauliflower, kale, Brussels sprouts, and all varieties of those
all have the same scientific name. Yes, we have to have duplicates.

And yep, there are multiple different species that have the same common
name. It *is* messy, but only because real life is.

pozorvlak:
Alternatively, ScientificName has_and_belongs_to many Crops.

Br3nda:
Are alt names used for anything more than looks up when creating plantings and harvests?

Br3nda:
I've been thinking of having a preferred name too.  It is very weird to see all these American (I presume) names on the crops in my garden.

maco:
Search in general uses them, for when someone wants to read about a crop
but not necessarily plant it. I don't think our Markdown extension is aware
of them, but if it is then the duplication means we'd need to either link
that to a disambiguation page (a la wikipedia) or have a dropdown for the
poster to choose from.

Br3nda:
I've re-instated the old duplicate-making functionality here #1241

PR #1228, opened 2017-02-18
Release 20

cesy:
Mostly cleanup, but let's keep going with frequent releases.


maco:
There are security fixes in here, so fast is good.

Br3nda:
Note: I've changed nearly every view in this, and lots of controllers, so anyone who can run this and test lots if appreciated. The forums seem to have the least specs, so testing there is useful.

Br3nda:
i hit the button

PR #1230, opened 2017-02-20
DRY the gardens controller

Br3nda:


app/controllers/application_controller.rb
```diff
@@ -77,4 +77,8 @@ def configure_permitted_parameters
       )
     end
   end
+
+  def expire_homepage
+    expire_fragment("homepage_stats")
+  end
```
Br3nda: this can be called by members and plantings controller, when i get to DRYing those

PR #1237, opened 2017-02-23
Bundle Update on 2017-02-23

deppbot:
**Updated RubyGems:**

- Updated: actionmailer, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/actionmailer/CHANGELOG.md))
- Updated: actionpack, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/actionpack/CHANGELOG.md))
- Updated: actionview, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/actionview/CHANGELOG.md))
- Updated: [active_utils](http://github.com/shopify/active_utils), [3.2.3...3.2.5](https://github.com/shopify/active_utils/compare/v3.2.3...v3.2.5) ([CHANGELOG](https://github.com/Shopify/active_utils/blob/master/CHANGELOG.md))
- Updated: activejob, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/activejob/CHANGELOG.md))
- Updated: activemodel, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/activemodel/CHANGELOG.md))
- Updated: activerecord, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/activerecord/CHANGELOG.md))
- Updated: activesupport, 4.2.7.1...4.2.8 ([CHANGELOG](https://github.com/rails/rails/blob/v4.2.8/activesupport/CHANGELOG.md))
- Updated: [autoprefixer-rails](https://github.com/ai/autoprefixer-rails), [6.6.1...6.7.5](https://github.com/ai/autoprefixer-rails/compare/6.6.1...6.7.5) ([CHANGELOG](https://github.com/ai/autoprefixer-rails/blob/master/CHANGELOG.md))
- Updated: [bonsai-elasticsearch-rails](https://github.com/omc/bonsai-elasticsearch-rails), [0.0.4...0.2.0](https://github.com/omc/bonsai-elasticsearch-rails/compare/v0.0.4...v0.2.0)
[...]

PR #1238, opened 2017-02-23
Upgrade rails, other gems

CloCkWeRX:
Another version of #1237 - chipping away again at stale deps.


CloCkWeRX:
Not sure if 

1) Crop csv loading alternate names doesn't add a duplicate alternate name
     Failure/Error: expect(tomato.alternate_names.size).to eq 1
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/models/crop_spec.rb:505:in `block (4 levels) in <top (required)>'

is flakey or legit regression.

CloCkWeRX:
Maybe more alarming is the fact that exited with a 0/'tests passed'!

CloCkWeRX:
Uh, hold off on merging this; its passing incorrectly (#1240). I'm not sure if the test was failing at the start.

pozorvlak:
@mention-bot tags in a lot of people who are no longer active on the project: we should configure it not to, or revoke their commit access.

cesy:
Let's remove commit access for the people who haven't been active in a long time.

pozorvlak:
So that would be @tygriffin and @leto at least. @jcaudle recently mentioned
on IRC that he might have time to get involved again.

PR #1240, opened 2017-02-23
Ensure failed build fails travis

CloCkWeRX:
19d65587beca220a09aabbc7cf29dfbe99b985ed may have similar regression (travis executed and checked each return code, but wouldn't any longer in that commit)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,7 @@ script:
     if [ "$STATIC_CHECKS" = 'true' ]; then
       bundle exec script/check_static
     else
-      bundle exec rake db:migrate --trace
-      bundle exec rspec --tag $RSPEC_TAG spec/
-      bundle exec rake jasmine:ci
+      bundle exec rake db:migrate --trace && bundle exec rspec --tag $RSPEC_TAG spec/ && bundle exec rake jasmine:ci
     fi
 services:
   - elasticsearch
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,7 @@ before_install:
 before_script:
   - >
     if [ "$STATIC_CHECKS" != 'true' ]; then
-      bundle exec rake db:create db:migrate db:test:prepare
-      bundle exec rake assets:precompile
+      bundle exec rake db:create db:migrate db:test:prepare && bundle exec rake assets:precompile
     fi
 script:
   - >


CloCkWeRX:
@pozorvlak maybe a better approach for this would be to return to the one-line-per-executable in the yaml; but prefix all commands with a relevant test.

IE instead of 
```
    if [ "$STATIC_CHECKS" = 'true' ]; then
      bundle exec script/check_static
    else
      bundle exec rake db:migrate --trace && bundle exec rspec --tag $RSPEC_TAG spec/ && bundle exec rake jasmine:ci
    fi
```
go for
```

 - if [ "$STATIC_CHECKS" = 'true' ]; then; bundle exec script/check_static; fi
 - if [ "$STATIC_CHECKS" != 'true' ]; then; bundle exec rake db:migrate --trace; fi
 - if [ "$STATIC_CHECKS" != 'true' ]; then; bundle exec rspec --tag $RSPEC_TAG spec/; fi
 - if [ "$STATIC_CHECKS" != 'true' ]; then; bundle exec rake jasmine:ci; fi

```

A little ugly but maybe a lot cleaner than &&'s all over the shop.

CloCkWeRX:
@Br3nda can you see if you can bisect to pin down the CSV count regression locally, with this one applied? I had a hard time spotting it on travis builds

pozorvlak:
@CloCkWeRX would that help? It looks to me like Travis is running `bundle exec rspec spec/`, throwing its return value away, and only looking at the return value of the last line to be run, which is `bundle exec rake jasmine:ci`. In which case, we want to run all the tests on one line (or maintain an explicit did-previous-tests-pass variable).

Br3nda:
We could achieve this using overcommit.
it would run all linters configured in overcommit, and all the devs with a checkout can run those same commands easily.

Br3nda:
https://github.com/brigade/overcommit

this is the config i use
```
PreCommit:
  ALL:
    on_warn: fail  # Treat all warnings as failures
    quiet: false
  RuboCop:
    enabled: true
    command: ['bundle', 'exec', 'rubocop', '-D']
  TrailingWhitespace:
    enabled: true
    exclude:
      - '**/db/structure.sql'  # Ignore trailing whitespace in generated files
  CoffeeLint:
    enabled: true
  HardTabs:
    enabled: true
  CommitPlease:
    enabled: true
  AuthorEmail:
    enabled: true
  AuthorName:
    enabled: true
  YamlLint:
    enabled: true
  CssLint:
    enabled: true
  HamlLint:
    enabled: true
  JsonSyntax:
    enabled: true
  BundleOutdated:
    enabled: true
    on_warn: warn
  RailsSchemaUpToDate:
    enabled: true

PostCheckout:
  ALL:
    quiet: false
  BundleInstall:
    enabled: true
  RailsSchemaUpToDate:
    enabled: true

PostMerge:
  BundleInstall:
    enabled: true
  RailsSchemaUpToDate:
    enabled: true

PrePush:
  ALL:
    on_warn: fail  # Treat all warnings as failures
    quiet: false
  Brakeman:
    enabled: true
  RSpec:
    enabled: true
    command: ['bundle', 'exec', 'rspec']
```

pozorvlak:
OooooOOoOOOoooo. That looks fantastic!

CloCkWeRX:
> it looks to me like Travis is running bundle exec rspec spec/, throwing its return value away, and only looking at the return value of the last line to be run

Sort of - though its bash/the shell environment making that call.

Either `a && b && c` would chain the commands together (and terminate immediately on first failure); or having the commands listed in the .travis.yml as separate array items (so travis executes each and checks the exit code)

If we wanted to emulate the old travis behaviour in the conditional we added; either want to put them in as seperate array items letting travis look at the exit code; or check/store `$?` after each command.

Something like:
```
if foo; then
else
   thing
   did_thing_exit_ok = $?
   thing2
   did_thing2_exit_ok = $?

   test did_thing2_exit_ok == 0 && test did_thing_exit_ok == 0 
fi
```

Br3nda:
I've got overcommit running in travis, if you want to integrate that:
https://github.com/Br3nda/growstuff/commit/08c51dd84f2c97f7cd20a01341becca892debba1

Br3nda:
To make it stop on error I added `set -e`.  I had to set it back to +e later or other parts of the script Travis makes for itself will fail.

Br3nda:
```
ript:
  - >
    set -e;
    if [ "${STATIC_CHECKS}" = "true" ]; then
      bundle exec ./script/check_static
    else
      bundle exec rake db:migrate --trace
      bundle exec rspec --tag $RSPEC_TAG spec/
      bundle exec rake jasmine:ci
    fi;
    set +e
```

CloCkWeRX:
@Br3nda I'm happy with any approach that ensures the build fails; did you want to throw in a PR for the +e/-e behaviour; its cleaner than this.

Br3nda:
I've got overcommit working in #1247

PR #1241, opened 2017-02-24
Fix up csv importer, should make duplicates

Br3nda:
Now matches the old functionality.


CloCkWeRX:
> spec/models/crop_spec.rb:507:7: W: Lint/BlockAlignment: end at 507, 6 is not aligned with it "Adds a duplicate alternate name for second crop" do at 500, 8.

Merge away when that goes green

PR #1260, opened 2017-03-02
Typo fix; install git-guilt (needed for post-commit hook)

pozorvlak:


Br3nda:
Oops. Git guilt is interesting but not needed. Happy to merge.

PR #1263, opened 2017-03-04
Bundle audit needs updating in test runs

Br3nda:
This tells bundle-audit to download the latest security advisories


CloCkWeRX:
We might be able to get away with deppbot doing pull requests that include security issues - potentially sidestepping bundler-audit failing a PR because an unrelated vulnerability was found. Worth thinking about if we see it constantly rejecting otherwise good PR's

Br3nda:
btw: the bundle audit is only a warning.

PR #1264, opened 2017-03-04
Added missing semicolons to js files

Br3nda:


cesy:
Taking out the warning on unmodified lines makes sense, but why are we taking out the warning on fail?

Br3nda:
The default is for a fail to be a fail. 
We have fail set to only warn. 

Now that jshint passes, let's set fail to be a fail again (default)

PR #1265, opened 2017-03-04
Keep on shrinking that crop model file

Br3nda:


Br3nda:
Something is really really broken on this PR. I need to set up elastic search on my dev to really test this

PR #1272, opened 2017-03-09
Lock haml_lint to 0.20.0 and update other gems

Br3nda:


Gemfile
```diff
@@ -121,7 +121,7 @@ group :development, :test do
   gem 'factory_girl_rails'              # for creating test data
   gem 'haml-i18n-extractor'
   gem 'haml-rails'                      # HTML templating language
-  gem 'haml_lint'                       # Checks haml files for goodness
+  gem 'haml_lint', '~> 0.20.0'          # Checks haml files for goodness
```
CloCkWeRX: Worth adding a link to the why (I'll quickly go fix the original PR perhaps?)

CloCkWeRX: Oh, and don't be afraid to check out the bundle-update-xx-yy branch from this repo either, to pile changes atop what deppbot does - I've found that makes life a bit easier!

Issue #474, opened 2014-12-12
Remove photo from planting


Skud:
I should be able to remove a photo from a specific planting without deleting it from the system altogether.  (However, if that planting represents the last use of the photo, then the photo should be deleted -- I think this happens already via the photo model.)

PR #1279, opened 2017-03-12
View harvests for a planting

Br3nda:
The routes already existed for this, but wasn't filled in.
Noticed when starting work on #474


CloCkWeRX:
Test was just flakey I think, re-running

PR #1282, opened 2017-03-12
Gardens order

Br3nda:
This page is linked to from the front page: 
http://www.growstuff.org/gardens

and shows all gardens

it is ordered by name - and the first page is gardens that haven't been planting in for 4 years.

This PR changes that to order by :updated_at, resulting in a page full of active gardens.


CloCkWeRX:
Should look at others (harvest, plantings, etc) to default to the latest if they aren't already

PR #1287, opened 2017-03-16
Readme update for new contributors

Br3nda:


Br3nda:
@arun1595 as you're a new contributor - do you have more questions we could answer in the README?

Issue #1250, opened 2017-02-28
Web workers are exceeding memory quota


pozorvlak:
From the Heroku logs:

```
2017-02-28T10:47:53.136222+00:00 heroku[web.1]: Process running mem=564M(110.0%)
2017-02-28T10:47:53.136306+00:00 heroku[web.1]: Error R14 (Memory quota exceeded)
```

This seems to happen about once every 20 seconds: I can't see any pattern to what comes before.

Br3nda:
Can you see something in the access logs? Is it some (spammer bot) hitting with large posts of data?

Br3nda:
... and is it happening on staging?

Br3nda:
2017-04-17T03:03:24.756308+00:00 heroku[web.1]: Error R14 (Memory quota exceeded)

Still happening.
continuing to chase this

Br3nda:
#1315 will help this, once merged and released

PR #1292, opened 2017-03-24
Query optimising

Br3nda:
Reduces the number of queries on the home page, by:

* Moving interesting to a scope, so it's performed in the database, instead of pulling all records and running ruby
* Adding .includes() for several things we need at display time.
* Adding the bullet gem, which displays in-efficient queries in development

Hoping this helps #1250


app/models/planting.rb
```diff
@@ -8,19 +8,20 @@ class Planting < ActiveRecord::Base
...
+  scope :has_photos, -> { joins(:photos).where("photos.id IS NOT NULL") }
+
   delegate :name,
     :en_wikipedia_url,
     :default_scientific_name,
     :plantings_count,
     to: :crop,
     prefix: true
 
-  default_scope { order("created_at desc") }
```
Br3nda: this was a duplicate (in the file twice)

app/models/crop.rb
```diff
@@ -33,12 +33,15 @@ class Crop < ActiveRecord::Base
...
     # ok on sqlite and psql, but not on mysql
-    where(approval_status: "approved").reorder('random()')
+    approved.reorder('random()')
   }
   scope :pending_approval, -> { where(approval_status: "pending") }
   scope :approved, -> { where(approval_status: "approved") }
   scope :rejected, -> { where(approval_status: "rejected") }
 
+  scope :interesting, -> { approved.has_photos }
+  scope :has_photos, -> { includes(:photos).where.not(photos: { id: nil }) }
```
pozorvlak: Could this go into PhotoCapable?

app/models/member.rb
```diff
@@ -36,6 +36,15 @@ class Member < ActiveRecord::Base
...
   scope :wants_newsletter, -> { where(newsletter: true) }
 
+  scope :interesting, lambda {
+    confirmed
+      .located
+      .recently_signed_in
+      .has_plantings
+  }
+
+  scope :has_plantings, -> { joins(:plantings).where("plantings.id IS NOT NULL") }
```
pozorvlak: I slightly prefer the `where.not(plantings: { id: nil }) }` style, but no biggie.

Br3nda: oh yes, i didn't know i could do that! looks better

app/models/crop.rb
Br3nda: Good idea - i've made a task to do that @ #1294

app/models/member.rb
pozorvlak: You did it in a couple of other places in this PR :-)

Issue #619, opened 2015-01-11
Add inflections for irregular plural crop names


Skud:
Some crop names have irregular pluralisation, eg. kale -> kale.  We should add inflections for these to to make things nicer when displaying crop pages.  Please let us know any crops that need inflections added, in the comments below.

So far we have:
- broccoli
- kale
- squash (pluralises to squashes but "squash" seems more common)

If we ever get around to doing this in the database somehow (see eg. #612 which also suggests this) then at least these inflections will give us a starting point.

FYI to developers: you add these inflections in config/initializers/inflections.rb and you can see them in action on the relevant crop page (look at the sidebar for example).

Skud:
FYI: it would be great if someone could go through http://growstuff.org/crops/hierarchy (or any other handy list of crops you might have) and make a list of all the ones which pluralise irregularly.  The most common irregularity is that they simply don't pluralise (eg. kale -> kale) so look for those first and you'll get most of them.

pozorvlak:
We've added a few common ones in #624, but we're leaving this issue open so people can suggest more.

jinky32:
Hi,
I started to have a go at this but wasn't really confident with Ruby changes shown in #624 so I kind of went off on a bit of a tangent :) and created a SKOS representation of some parts of the hierarcy (I'm not really sure how useful it is so I didn't go too far!).

You can see it as Gist at https://gist.github.com/jinky32/370d30dc00f3ce14eb11 or the file is hosted at http://www.stuartbrown.name/veg.skos.xml.  It's quite fun to put the SKOS into http://labs.sparna.fr/skos-play/upload and either print a Hierarchical Tree (second page) or use some of the interactive visualisations at the very bottom of the age (visualise).

I added a couple of the alternative labels defined in a google doc on talk.growstuff as skos:altLabel and also added a couple of owl:sameAs links to wikidata URIs.

As I say not sure if it's useful but it kept me entertained for a little while :)

hbrodsk1:
I took some time to add more irregulars to the list as well updated the feature specs to make sure all of the previous ones were included in the tests as well as the new ones I added

PR #1296, opened 2017-03-30
Update irregular crop pluralizations issue #619

hbrodsk1:
Added to the list of irregular crop pluralizations and added tests for the ones that were added as well as ones that were on the list already but not accounted for in tests.


Br3nda:
Thanks for the contribution @hbrodsk1

PR #1303, opened 2017-04-03
Disable eslint

Br3nda:


Br3nda:
Let's add to ignores instead

Br3nda:
Haven't had time to add ignores, do let's not keep the build broken. Let's disable eslint.

PR #1306, opened 2017-04-10
Make sure interesting members load once

### LINKED ISSUE #1305: Interesting members are not unique
image

This is a bug i introduced when i moved this to a scope. In my dev dataset there aren't duplicates

it needs to be a left outer join -- but it isn't immediately obvious to me how to convince active record to do that AND recognise which model it's to create objects as.

CloCkWeRX:
Fixes #1305 I think (just checking travis...)


app/models/member.rb
```diff
@@ -43,7 +43,7 @@ class Member < ActiveRecord::Base
       .has_plantings
   }
 
-  scope :has_plantings, -> { joins(:plantings).where("plantings.id IS NOT NULL") }
+  scope :has_plantings, -> { joins(:plantings).group(:owner_id) }
```
Br3nda: :sparkles:

Issue #863, opened 2016-04-10
Know the rainfall for my garden


CloCkWeRX:
Farmlogs does this, we could do it on a smaller scale: each day hit a weather API and record temp/rainfall/etc.

This lets me compare climate to yield when planting next year

CloCkWeRX:
https://www.mapbox.com/blog/weather-along-route/ points out http://openweathermap.org/api which might be of interest

Br3nda:
Maybe wunderground integration would work here?

CloCkWeRX:
I've done a proof of concept for this; here's the rake task output:
```
clockwerx@clockwerx-laptop:~/growstuff$ bundle exec rake planting:determine_current_weather
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Weather for Adelaide, SA is 17 degree(s) celcius and Clouds
Weather for Adelaide, SA is 17 degree(s) celcius and Clouds
Weather for Watson is 12.34 degree(s) celcius and Clouds
Weather for O'Connor is 12.35 degree(s) celcius and Clouds
Weather for Downer is 12.36 degree(s) celcius and Clouds
```

Plantings have PlantingWeatherLogs; which have a weather_data JSON structure. That ties us to postgres; but I think its worth it.

Sample data from the API call looks like:
```
  {"coord"=>{"lon"=>138.6, "lat"=>-34.92},
   "weather"=>
    [{"id"=>803,
      "main"=>"Clouds",
      "description"=>"broken clouds",
      "icon"=>"04d"}],
   "base"=>"stations",
   "main"=>
    {"temp"=>17,
     "pressure"=>1024,
     "humidity"=>51,
     "temp_min"=>17,
     "temp_max"=>17},
   "visibility"=>10000,
   "wind"=>{"speed"=>5.1, "deg"=>190},
   "clouds"=>{"all"=>75},
   "dt"=>1491809400,
   "sys"=>
    {"type"=>1,
     "id"=>8204,
     "message"=>0.0103,
     "country"=>"AU",
     "sunrise"=>1491771922,
     "sunset"=>1491812870},
   "id"=>7839644,
   "name"=>"Adelaide",
   "cod"=>200}}
```

leogaggl:
Hi Daniel,

I am looking at Growstuff (again) for a gardening-related project for myself and my partner, and I noticed this issue as well as the associated PR. My work involves a lot of sensor data, and one of these is weather sensors. 

[...]

PR #1307, opened 2017-04-10
Log weather periodically via OpenWeatherAPI - proof of concept/WIP

CloCkWeRX:
#863
For simplicity, I've just left a default/hardcoded API key in so its easy to play with - that'll be gone by the final version

TODO:
 * Check the ODBL license is appropriate to integrate with our DB.
 * Check if its really ODBL or CCBYSA (http://openweathermap.org/price is unclear)
 * Staging, production, etc accounts
 * Frontend display of data
 * API limits.
 * On deployment; configuring https://devcenter.heroku.com/articles/scheduler
 * On deployment, ensure we are on postgres 9.4 or higher (https://devcenter.heroku.com/articles/heroku-postgresql#version-support-and-legacy-infrastructure)


CloCkWeRX:
Super annoying, I have no idea how likes_controller has changes on it.

app/models/planting_weather_log.rb
```diff
@@ -0,0 +1,50 @@
...
+#     "message":0.0097,
+#     "country":"GB",
+#     "sunrise":1491801287,
+#     "sunset":1491850179
+#   },
+#   "id":2643743,
+#   "name":"London",
+#   "cod":200
+# }
+class PlantingWeatherLog < ActiveRecord::Base
```
Br3nda: Shouldn't this be per garden? Or even per location? The basil and the tomatoes are next to each other so have the same weather afaik

Br3nda:
Btw I have and old fork of grow stuff with moisture , light, temp,and fertiliser readngs per planting (actually per sensor, which is associated with many plantings).

CloCkWeRX:
> schema.rb changing
That's my local DB having a bunch of different feature branches on it; and me (tiredly) doing a rubocop --auto-correct && git commit -a; sigh.

Prior to merge, I'll go clean it up to be only relevant bits.

app/models/planting_weather_log.rb
CloCkWeRX: I went for planting so you can answer things like "Why did my leeks grow better last year than this year? Oh, 27 days of 40 degree temperature in a row on last year's planting" more easily 

IE its simpler to do ```SELECT AVG(temp) FROM planting_weather_log WHERE planting_id = ?``` and only have data recorded for the duration of the planting; than it is to do

```SELECT AVG(temp) FROM garden_weather_log JOIN gardens JOIN plantings WHERE garden_weather_log.created_at BETWEEN planting.start and planting.end GROUP BY planting_id```
 x lots.


Plantings also have a start and end date usually (planting.active); while a Garden is forever; and at least one per user; even if they new do anything with the site. 
I suspect we'd run out of the free tier of API usage if we did that for everyone's gardens.

For dealing with plantings in the same garden/location/etc; we can probably cache the API requests at a city/geocoded location level.

CloCkWeRX:
Will pick this up again in a while

PR #1309, opened 2017-04-13
Fix 1305

### LINKED ISSUE #1305: Interesting members are not unique

Br3nda:
this is PR #1306 with the rubocop fixed
fixes #1305


Br3nda:
i've also upgraded haml_lint, as per https://github.com/brigade/overcommit/issues/486

PR #1310, opened 2017-04-13
Adding counter caches

Br3nda:
This reduces the query count on many pages by about 50.


Br3nda:
part of #1205

Br3nda:
part of #1250

PR #1316, opened 2017-04-16
Garden thumbnail display cleanup

Br3nda:
Puts the `<li>` inside a `<ul>`
adds a class to the `<ul>`
turns off the bullets  in `ul.plantings`

Moves the number `2` from being in two far apart files, to instead being on subsequent lines in the same file. This is better connascence. (It's only called in this one place)

Added `.includes(:crop)` to reduce query count

Updated specs

before:
image

after:
image

### DIFF
--- a/app/helpers/gardens_helper.rb
+++ b/app/helpers/gardens_helper.rb
@@ -21,13 +21,14 @@ module GardensHelper
     if plantings.blank?
       "None"
     else
-      output = ""
-      plantings.first(2).each do |planting|
+      output = '<ul class="plantings">'
+      plantings.each do |planting|
         output += "<li>"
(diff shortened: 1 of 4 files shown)


app/helpers/gardens_helper.rb
```diff
@@ -21,13 +21,14 @@ def display_garden_plantings(plantings)
     if plantings.blank?
       "None"
     else
-      output = ""
-      plantings.first(2).each do |planting|
```
CloCkWeRX: Will that render funny if there are 3+ plantings now?

CloCkWeRX: Nevermind, I see now

PR #1318, opened 2017-04-17
Photo ui

Br3nda:
- Editing a photo's title
- Don't overwrite photo's title if we re-import from flickr later
- New functionality, we can now remove a link from an item (harvest, garden, seed, plantings). so we don't need to delete the whole photo anymore if we link to the wrong item.
- Places the photo at the top, as that is what the user is seeking on this page.
- Makes the photo size responsive, so it fits in phone screens
- Fixed bug that was pushing the menu off screen on phones

### DIFF
--- a/app/models/photo.rb
+++ b/app/models/photo.rb
@@ -34,13 +34,23 @@ class Photo < ActiveRecord::Base
     licenses = flickr.photos.licenses.getInfo
     license = licenses.find { |l| l.id == info.license }
     {
-      title: info.title || "Untitled",
+      title: calculate_title(info),
       license_name: license.name,
       license_url: license.url,
       thumbnail_url: FlickRaw.url_q(info),
       fullsize_url: FlickRaw.url_z(info),
       link_url: FlickRaw.url_photopage(info)
     }
+  end
+
+  def calculate_title(info)
+    if id && title # already has a title saved
+      title
+    elsif info.title # use title from flickr
+      info.title
+    else
+      'untitled'
+    end
   end
 
   def set_flickr_metadata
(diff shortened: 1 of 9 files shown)


Br3nda:
before
image


after
image

Br3nda:
before
image

after
image

Br3nda:
and when not signed in, it will look like this
image

image

app/models/photo.rb
```diff
@@ -10,6 +10,10 @@ class Photo < ActiveRecord::Base
 
   default_scope { order("created_at desc") }
 
+  def associations?
+    plantings.any? || harvests.any? || gardens.any? || seeds.any?
```
Br3nda: should probably add counter caches for these

Br3nda:
fixes #474

app/views/photos/_photo_association_delete.haml
```diff
@@ -0,0 +1,4 @@
+- if can? :edit, photo
+  = link_to photo_associations_path(photo_id: photo.id, type: type, id: thing.id),
+    method: 'delete', class: 'btn btn-default btn-xs' do
+    %span.glyphicon.glyphicon-remove{ title: "Remove link" }
```
CloCkWeRX: Need i18n?

app/views/photos/show.html.haml
```diff
@@ -8,7 +8,19 @@
...
 .row
-  .col-md-6
+  .col-md-8
+    %p= image_tag(@photo.fullsize_url, alt: @photo.title, class: 'img img-responsive')
+
+  .col-md-4
+    %p
+      - if can? :destroy, @photo
+        = link_to @photo, method: :delete,
+          data: { confirm: 'Are you sure?' }, class: 'btn btn-default btn-xs' do
```
CloCkWeRX: Ditto i18n

CloCkWeRX:
Will let you decide if you want to internationalise bits or just do that later.

PR #1320, opened 2017-04-17
Route for flickr auth callback -- Fixes  #1106

### LINKED ISSUE #1106: cannot connect flickr account

Br3nda:
This puts a route in, where omniauth-flickr seems to want it, since the release 14 gem updates that broke it.

I don't have facebook, so someone else checking the facebook auth still works is a good idea.

### DIFF
--- a/app/views/photos/new.html.haml
+++ b/app/views/photos/new.html.haml
@@ -36,5 +36,5 @@
 - else
   .alert
     You must
-    = link_to "connect your account to Flickr", '/auth/flickr'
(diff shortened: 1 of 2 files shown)


cesy:
Looks good, let's make sure we test facebook on staging.

CloCkWeRX:
Facebook signin seems to work OK still on prod

maco:
Yay!

PR #1324, opened 2017-04-22
Plantings#index button move-around

Br3nda:
* The button are too big to fit beside the text, so moved into a horizontal bar.
* Used glyphicons, so save space
* Made the Planting attributes font size normal sized, because it's the most important text on a planting page, and shrinking it makes it hard to read.
* Removed the harvests list. It doesn't need to be here.

Before:
image

After: 
image


app/assets/stylesheets/overrides.sass
```diff
@@ -118,14 +116,13 @@ p.stats
...
-      font-size: 85%
       width: 100%
 
       dt
         text-align: left
-        width: 80px
+        width: 120px
       dd
-        padding-left: 80px
+        padding-left: 120px
```
Br3nda: this can be wider now, and take up the room that used to have buttons in it.

Br3nda:
before
image

after
image

CloCkWeRX:
Nice.

PR #1325, opened 2017-04-22
Plantings form - photos/images selectors

Br3nda:
I'd like to make the plantings form a whole lot less daunting. This is a small step there. 
By using the names+photos of the gardens, the user can work out which garden they want to create the planting in much faster. It's one click, and it's visual now.

css needs some work, and the specs needs updating.

To make a planting, all that's needed is the crop and the garden. The form could make that clearer, to make entering data much faster for users.


Br3nda:
image

Br3nda:
Perhaps they can plant multiple crops in the garden at once...

cesy:
Travis is failing and it's out of date - can you check, please?

cesy:
I like the idea of this change, though.

Br3nda:
@cesy I was waiting on feedback before fixing tests
I can make the CSS look better and update specs then...

Br3nda:
Other idea if a "quick plant" form, from the Gardens#show page -- where you only need to select the crop and it creates a planting in that garden on that day (and take the suniness etc from the other plantings)

PR #1329, opened 2017-04-25
Moved planting maturity to planting model. simplified controller

Br3nda:


app/models/planting.rb
```diff
@@ -137,4 +124,31 @@ def percentage_grown(current_date = Date.current)
...
+    (finished_at - planted_at).to_i
+  end
+
+  def self.mean_days_until_maturity(plantings)
+    ## Given a set of finished plantings, calculate the average/mean time from start to finish
+    differences = plantings.collect(&:start_to_finish_diff)
+
+    return if differences.compact.empty?
+
+    differences.compact.sum / differences.compact.size unless differences.compact.empty?
```
Br3nda: i've left this as average (mean).
but i think it should change to median - so one user's 1600 day old tomatoes don't runs the stats for everyone else.

Br3nda: i think also it should try to work out the days_until _maturity from the user's own plantings first -- and if there in't enough data, try people in nearby climates, and then finally resort to using the entire dataset.

People in Singapore will have different results to me in Whanganui-a-tara

Br3nda:
've left this as average (mean).
but i think it should change to median - so one user's 1600 day old tomatoes don't ruin the stats for everyone else.

i reckon we should try to work out the days_until _maturity from the user's own plantings first -- and if there isn't enough data, try plantings in nearby climates, and then finally resort to using the entire dataset.

People in Singapore will have different results to me in Whanganui-a-tara

app/models/planting.rb
```diff
@@ -8,6 +8,8 @@ class Planting < ActiveRecord::Base
   belongs_to :crop, counter_cache: true
   has_many :harvests, -> { order(harvested_at: :desc) }, dependent: :destroy
 
+  before_save :calc_and_set_days_before_maturity
```
CloCkWeRX: Not a huge fan of before_saves, could this be better done with a planting updater pattern? IE any controller type action that would update maturity uses that; while other simple saves just use .save (rake tasks for example)

CloCkWeRX:
Sigh, ... 'action pattern'. Its fine now; and for this intended use case, have been bitten with unexpected side affects before though from extra queries/external behaviours glued onto a .save

app/models/planting.rb
Br3nda: So, in the controller's `update()` and `create()`, we'd call `@planting.calc_and_set_days_before_maturity()` before save?

CloCkWeRX: Yeah; that sort of thing. 

It's probably not a big deal right now, but if we did something like "... and when a planting is probably matured; issue an email prompt to harvest it"; for example; it wouldn't make as much sense in a before_save.

Br3nda:
@clockwerx I agree with your concerns. I'll get the call moved back to the controller

PR #1337, opened 2017-05-15
Moving has_photos scope to PhotoCapable

### LINKED ISSUE #1294: move has_photos scope to PhotoCapable

Br3nda:
This moves some repeating code to photocapable.

fixes #1294


Br3nda:
I've removed the default scope from Harvest. It causes an ORDER clause to be added in queries even when not selecting from that table. E.g. if we 're looking for PlantParts that have a Harvest

app/models/harvest.rb
```diff
@@ -9,7 +9,6 @@ class Harvest < ActiveRecord::Base
   belongs_to :plant_part
   belongs_to :planting
 
-  default_scope { order('created_at DESC') }
```
CloCkWeRX: "Harvest ordering lists most recent harvests first" seems upset by this; may need corresponding changes to various controllers, specs, views

Br3nda:
The order ends up in crop's `has_many :plant_parts, through: :harvests`
Yeah, this needs more time than i could give in my lunch break.

app/models/harvest.rb
Br3nda: The default scopes i've seen in other projects are e.g. `where(deleted: false)` etc. 

Growstuff has a log of query calls in the views too - where I've usually seen them in the controller with order and pagination controlled there.

Br3nda: I've added this default scope back in, and added a `reorder()` to plant_parts.harvests. this does what you'd expect unscoped to do.

Br3nda: Related: https://github.com/rails/rails/issues/5701
Unscoped doesn't take away the default scope's order by, but reorder did the trick.

PR #1340, opened 2017-05-20
Display seeds for trade, on homepage, as thumbnails.

### LINKED ISSUE #1334: home page tables don't adapt well to small screens
image

image

Note: that's a firefox screen shot. On Chrome on android the hamburger menu is push right off the screen when the tables are wider than the screen, and the user has to scroll to find it.

Br3nda:
Display seeds for trade, on homepage, as thumbnails.
Note: there is an existing seeds/_thumbnails template, but it's not really a thumbnail.

Fixed #1334  for seeds

### DIFF
--- a/app/models/seed.rb
+++ b/app/models/seed.rb
@@ -31,7 +31,8 @@ class Seed < ActiveRecord::Base
     allow_nil: true
 
   scope :tradable, -> { where("tradable_to != 'nowhere'") }
-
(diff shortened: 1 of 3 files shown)


Br3nda:
image

The photo comes from the crops

Br3nda:
I'd really like a review here. I have some view refactors i'd like to propose after this.

PR #743, opened 2015-05-22
WIP: Issue 468 member's account deletion

cesy:
Fixes issue #468

Not ready to merge yet.

Please refer to previous pull request (PR #662) for discussion.

PR #1341, opened 2017-05-21
Issue 1044 account deletion

### LINKED ISSUE #468: Account deletion
There is currently no option to delete a member's account.  As a member, I should be able to do this via my settings page.

We need to make sure that all related things are deleted (the member's plantings, gardens, photos, etc) but also make sure that:
- if the member is a crop wrangler, don't delete crops they've added -- just reassign them to cropbot or something
- if the member is or was an admin, ditto.
- we need to figure out what to do with deleted posts -- maybe replace them with "DELETED POST" so you can see that something was there, but not the details of what it was?  This is what Dreamwidth does.  If you don't do this, then any replies look very odd indeed.

Br3nda:
Uses acts_as_paranoid to mark members as deleted.

~~If a query does `Planting.all` etc, it will get all plantings, but some will have no owner.
To only see plantings with owners, do `Plantings.joins(:owner).all`~~
The default scopes only allows see gardens+plantings+Harvests+seeds with a not-deleted owner. 

includes code from #1340 because it needs the interesting seeds scope fixed.
this includes some code from #743 
fixes #468

- [x] Feature spec of pressing the delete account button
- [x] Check password matches before allowing deleting
- [x] Soft delete user's data too. (Gardens, Plantings, Harvests, Seeds).


app/controllers/members_controller.rb
```diff
@@ -54,6 +54,11 @@ def show
     end
   end
 
+  def destroy
```
Br3nda: TODO!! Needs to check the password supplied is correct before letting them destroy their account.

cesy:
Thank you so much for picking this up.

db/migrate/20170413221549_counter_caches.rb
```diff
@@ -4,7 +4,7 @@ def change
     add_column :members, :harvests_count, :integer
     add_column :members, :seeds_count, :integer
 
-    Member.find_each do |member|
+    Member.unscoped.find_each do |member|
```
Br3nda: I had to modify this old migration. The models now want the deleted_at column to exist, and it doesn't yet at the point in the migration.

Br3nda:
I need some decisions in whether it's okay to expose a deleted users gardens and plantings and photos and seeds. This data will appear to have no owner. I'm not sure this is what a user expects when they delete their account. Should I soft delete their other data too?

pozorvlak:
Yes, I think soft-deleting their other data is the Right Thing.

Br3nda:
Interestingly, the DELETE member action is already there -- no controller changes needed. I couldn't immediately work out how to override that - probably need to read up on devise more - hence this PR is still a work in progress.

Br3nda:
Added specs from @cesy 's  61f93df6ff5ef9bf74606a90a262f6db41092f09
Now checks password was provided before allowing delete
Changed order.member's belongs_to to include deleted
Included deleted members on the order search

Ready for review.

app/controllers/plantings_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
     @crop = Crop.find_by(slug: params[:crop]) if params[:crop]
     @show_all = params[:all] == '1'
 
-    @plantings = plantings.paginate(page: params[:page])
+    @plantings = plantings
```
Br3nda: oh, i saw this was already paginated, so i removed the duplicate while here.

Br3nda:
I'd really love a review here - so we can get the release out...
anyone?

cesy:
I'll have a look when I can, but don't have the brain for a full review at the moment, so may need help from others.

spec/features/members/deletion_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+      login_as(member)
+      FactoryGirl.create(:comment, author: member, post: othermemberpost)
+      FactoryGirl.create(:comment, author: other_member, post: memberpost, body: "Fun comment-y thing")
+      # deletion breaks if no wranglers exist
+      FactoryGirl.create(:cropbot)
+      # deletion breaks if ex_member doesn't exist
+      FactoryGirl.create(:member, login_name: "ex_member")
+    end
+
+    # scenario "has option to delete on member profile page" do
```
cesy: Does this one still need fixing?

spec/features/members/deletion_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+
+      scenario "replaces posts with deletion note" do
+        visit post_path(memberpost)
+        expect(page.status_code).to eq(404)
+      end
+
+      scenario "replaces comments on others' posts with deletion note, leaving post intact" do
+        visit post_path(othermemberpost)
+        expect(page).not_to have_content member.login_name.to_s
+        # TODO
```
cesy: Still outstanding?

spec/features/members/deletion_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+        # expect(page).to have_content other_member.login_name.to_s
+        # expect(page).to have_content "This comment was removed as the author deleted their account."
+      end
+
+      scenario "leaves a record of orders and payments intact" do
```
cesy: Thank you for covering this.

cesy:
Thank you so much for pushing this forward.

Br3nda:
Pending specs added -- and more model specs.
added a before_destroy hook to unsubscribe from newsletter

cesy:
I'd like someone to run a manual test on staging before we put this onto production.

PR #1343, opened 2017-05-24
Release 26

Br3nda:


app/controllers/admin_controller.rb
```diff
@@ -1,16 +1,12 @@
...
-    end
   end
 
   def newsletter
     authorize! :manage, :all
     @members = Member.confirmed.wants_newsletter.all
-    respond_to do |format|
-      format.html # index.html.haml
-    end
+    respond_with @member
```
Br3nda: Looks like a bug we missed. Better fix (and add regression tests) or newletter will break

Br3nda: fixed in #1352

Br3nda:
Need #1341 to pass review, then we can release this.

cesy:
Why is this dependent on that? It would be good to get it live either way.

Br3nda:
I have deleted my account on staging.
My photos are gone -- garden, plantings, harvest. The forum post and comments

While still logged in (before deletion) i discovered i can't comment on old forum posts. That's a separate problem. Seems forum commenting is broken on really old data.

Br3nda:
If you start a forum thread, the whole thread will 404
http://www.staging.growstuff.org/posts/brenda-20170611-what

cesy:
Okay, let's figure out forums separately, then, but looks like this is good to go.

PR #1345, opened 2017-05-27
Some more DRY controllers

Br3nda:


app/controllers/crops_controller.rb
```diff
@@ -4,37 +4,23 @@ class CropsController < ApplicationController
...
-    end
+    @crops = crops
+    @has_requested_pending = current_member.requested_crops.pending_approval.size if current_member
+    @filename = filename
+    respond_with(@crops)
   end
 
   def requested
-    @requested = Crop.pending_approval.where(requester: current_member).paginate(page: params[:page])
+    @requested = Crop.pending_approval
```
Br3nda: this repeats a query in `index()`. will dry this

app/controllers/crops_controller.rb
```diff
@@ -52,29 +38,20 @@ def wrangle
...
-    respond_to do |format|
-      format.html
-    end
   end
 
   # GET /crops/search
   def search
     @term = params[:term]
-    @matches = Crop.search(@term)
+    @matches = Crop.approved.search(@term)
```
Br3nda: Note: this is a functionality change... Maybe we wanted crops that are not approved to be findable?

app/controllers/crops_controller.rb
```diff
@@ -127,34 +89,25 @@ def create
...
-          create_name('alternate', value)
-        end
-        params[:sci_name].each do |index, value|
-          create_name('scientific', value)
-        end
-        unless current_member.role? :crop_wrangler
-          Role.crop_wranglers.each do |w|
-            Notifier.new_crop_request(w, @crop).deliver_now!
-          end
+    if @crop.save
```
Br3nda: I reckon this should be in a transaction -- so you get the whole crop data saved, never half

app/controllers/crops_controller.rb
```diff
@@ -164,55 +117,39 @@ def update
...
-        format.html { render action: "edit" }
-        format.json { render json: @crop.errors, status: :unprocessable_entity }
+    if @crop.update(crop_params)
+      recreate_names('alt_name', 'alternate')
+      recreate_names('sci_name', 'scientific')
+
+      if previous_status == "pending"
+        requester = @crop.requester
+        new_status = @crop.approval_status
+        Notifier.crop_request_approved(requester, @crop).deliver_now! if new_status == "approved"
```
Br3nda: This can DRY more

app/controllers/crops_controller.rb
```diff
@@ -231,4 +168,28 @@ def crop_params
                                                       :_destroy,
                                                       :id])
   end
+
+  def filename
+    "Growstuff-Crops-#{Time.zone.now.to_s(:number)}.csv"
+  end
+
+  def crop_json_fields
```
Br3nda: There's gotta be a better place for this

app/controllers/scientific_names_controller.rb
```diff
@@ -25,23 +26,22 @@ def new
...
 
-    flash[:notice] = 'Scientific name was successfully created.' if @scientific_name.save
+    @scientific_name.save
     respond_with(@scientific_name.crop)
   end
 
   # PUT /scientific_names/1
   # PUT /scientific_names/1.json
   def update
-    flash[:notice] = 'Scientific name was successfully updated.' if @scientific_name.update(scientific_name_params)
```
Br3nda: the flash comes from the responders gem now. so we don't have to repeat ourselves AND it'll be translated already.

app/models/crop.rb
```diff
@@ -47,8 +47,7 @@ class Crop < ActiveRecord::Base
     format: {
       with: %r{\Ahttps?:\/\/en\.wikipedia\.org\/wiki\/[[:alnum:]%_\.()-]+\z},
       message: 'is not a valid English Wikipedia URL'
-    },
-    if: :approved?
+    }, presence: true
```
Br3nda: the form wasn't enforcing that you need a wikipedia url.. I remember it used to.
the markup on the form says it's required, so let's make it required.

spec/features/crops/crop_wranglers_spec.rb
```diff
@@ -48,7 +48,7 @@
       fill_in 'en_wikipedia_url', with: "http://en.wikipedia.org/wiki/Maize"
       fill_in 'sci_name[1]', with: "planticus maximus"
       click_on 'Save'
-      expect(page).to have_content 'Crop was successfully created'
+      expect(page).to have_content 'crop was successfully created.'
```
Br3nda: the flash notices have changed a little because they're from the responders gem now.

Br3nda:
15 issues fixed in code climate. I think this is enough for this PR.

app/controllers/crops_controller.rb
```diff
@@ -4,40 +4,22 @@ class CropsController < ApplicationController
...
-
-    respond_to do |format|
-      format.html
-      format.json { render json: @crops }
-      format.rss do
-        @crops = Crop.recent.includes(:scientific_names, :creator)
-        render rss: @crops
-      end
-      format.csv do
-        @filename = "Growstuff-Crops-#{Time.zone.now.to_s(:number)}.csv"
```
CloCkWeRX: This seems like it wouldn't be handled with the responder, has it moved somewhere else?

app/controllers/crops_controller.rb
CloCkWeRX: I tend to use rabl in other projects; perhaps a short term would be to push it to an ERB JSON template:

`format.json render :partial => "users/show.json"`

app/controllers/crops_controller.rb
Br3nda: the filename? It's in line 14 now. telling rails to allow csv is handled in line 7

PR #1346, opened 2017-05-27
Fixed url in crop rss feed

Br3nda:


CloCkWeRX:
Does this need test coverage at all? Happy if you just wanted to add a spec direct to dev for it.

PR #1357, opened 2017-06-08
Bundle update 2017-06-08

cesy:
This is the same thing deppbot would do anyway. I suggest we get the release out the way first and put this in the next one.


CloCkWeRX:
Everything was in the rails version bump PR anyhow

PR #1362, opened 2017-06-15
Upgrade and unpin some gems

cesy:
This updates a load of gems, getting us closer to the Rails upgrade.


Gemfile
```diff
@@ -45,13 +45,13 @@ gem 'sidekiq'
 gem 'bluecloth'
 
 # Pagination
-gem 'will_paginate', '~> 3.0'
+gem 'will_paginate'
 
 # user signup/login/etc
-gem 'devise', '>= 4.0.0'
```
CloCkWeRX: I think >= pins should be retained; so that other gems can't downgrade us if they aren't compatible (we get a bundler error instead)

Issue #1356, opened 2017-06-05
Garden diary


Br3nda:
I use a paper garden diary to remind me of tasks, and suggest plantings. 
It would be great to see my plantings / progress in a diary form on growstuff.
Would this feature be useful to anyone else?

Br3nda:
Some scribblings in a moleskine.

![image](https://cloud.githubusercontent.com/assets/12266/26807334/00c32566-4aaa-11e7-9a93-d676a4fe1dd2.png)
![image](https://cloud.githubusercontent.com/assets/12266/26807339/0ba0e022-4aaa-11e7-946b-21737847ddd4.png)

Br3nda:
![image](https://user-images.githubusercontent.com/12266/26866787-dea7dde2-4bb6-11e7-9fdb-dad0fbb150c3.png)

pozorvlak:
Cool! I expect we'd want some kind of "export as ICS/RSS/etc" feature, so
people could include their gardening diaries in Google Calendar or
whatever.​

Br3nda:
I have a branch for this -- but it lead into interesting thoughts.

I think perhaps we should consider merging plantings and harvests into one table. and then adding a type -- We can then display all events from an efficient query, and also start adding other types like "pest detected" or "storm damaged", or "transplanted" as further event types.

Br3nda:
something like this http://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html

CloCkWeRX:
Had similar ideas for tasks like

- weeding reminders based on common weeds
- compost schedules (turn it)
- watering reminders
- other daily tasks

Given cron/scheduled tasks cost $; very keen on ical of events, with reminders that I can one click add to google calendar or similar.

Br3nda:
the heroku scheduler seems to be free.. ?

CloCkWeRX:
Maybe done with

- Activities (the UX could use refinement), but has the ical version of this.
- Explicit transplanting.

PR #1381, opened 2017-08-03
Publish ical for plantings/plantings by owner index

CloCkWeRX:
A small part of #1356 

image

Needs its own responder to be done for Harvests, and we probably want a user centric controller/set of routes to do combined feeds; so I've left that for later.


CloCkWeRX:
Sigh, have to rewrite it with respond_with in mind.

PR #1391, opened 2017-09-05
Bundle update and change to platform-api instead of heroku

Br3nda:
I got this upon bundle install

```
Post-install message from heroku-api:

‼️  The heroku-api gem will not work.
‼️  You must instead use the platform-api gem.
‼️  The heroku-api gem communicated with the Legacy API which has been disabled.
‼️  https://devcenter.heroku.com/changelog-items/118
```

So here's a bundle update, and a heroku gem change.


Br3nda:
BTW: i re-enabled codeclimate for this project -- it was possible linked to @maco 's acocunt before (it used to show her photo on the status).. but stopped a while ago.

Br3nda:
updating from erubi to erubis gem broke stuff - pinned better errors version so we don't upgrade yet.

planting.percentage_grown was failing - returning 0 -- so changed to use floats so the percentage is calculated correctly

pozorvlak:
Looks good to me, assuming tests pass. I was going to ask why you removed git-guilt, but looking at the npm description it doesn't look like it does anything we need.

Br3nda:
There's a repeating deprecation warning about a language code changing.

Br3nda:
Quite a few elasticsearch failures

     Elasticsearch::Transport::Transport::Errors::BadRequest:

       [400] {"error":{"root_cause":[{"type":"parsing_exception","reason":"Unknown key for a START_OBJECT in [filter].","line":1,"col":152}],"type":"parsing_exception","reason":"Unknown key for a START_OBJECT in [filter].","line":1,"col":152},"status":400}

     # ./app/models/crop.rb:229:in `search'

     # ./spec/models/crop_spec.rb:407:in `block (3 levels) in <top (required)>'

  8) Crop search doesn't find non-matches

     Failure/Error: response.records.to_a

     

     Elasticsearch::Transport::Transport::Errors::BadRequest:

       [400] {"error":{"root_cause":[{"type":"parsing_exception","reason":"Unknown key for a START_OBJECT in [filter].","line":1,"col":150}],"type":"parsing_exception","reason":"Unknown key for a START_OBJECT in [filter].","line":1,"col":150},"status":400}

     # ./app/models/crop.rb:229:in `search'

     # ./spec/models/crop_spec.rb:399:in `block (3 levels) in <top (required)>'

  9) Crop search doesn't find 'pending' crop

     Failure/Error: response.records.to_a

     

     Elasticsearch::Transport::Transport::Errors::BadRequest:

       [400] {"error":{"root_cause":[{"type":"parsing_exception","reason":"Unknown key for a START_OBJECT in [filter].","line":1,"col":152}],"type":"parsing_exception","reason":"Unknown key for a START_OBJECT in [filter].","line":1,"col":152},"status":400}

Br3nda:
Production is running elastic search 1.7.5.
I'm running 1.7.6

But Travis is running 5.5.0

The right answer is probably to upgrade everything to the newer elastic search.

Br3nda:
staging is on 2.4.0..

That's not great. We should probably bump production up to 2.4.0

Br3nda:
SO MANY YAKS SHAVED.

I've changed the Travis config, to not use the default Travis version of elasticsearch (5.5.0)
Instead, I added sudo and installed and started elasticsearch 2.4.0

This passes

We do need to upgrade further, but the elastic search we're using on Heroku seems to have 2.4.0 as their newest available.

Also, production instance is on elasticsearch 1.7.5 and really needs upgrading to match staging and Travis.

PR #1400, opened 2017-10-17
Migrate to new heroku platform api - maintenance mode trigger

Br3nda:
Attempting to fix the heroku deploys - this is tricky to test so i haven't.


pozorvlak:
1400 pull requests - wow.

PR #1404, opened 2017-10-20
Sort finished plantings by finished_at

### LINKED ISSUE #1383: Previously planted not sorted
Plantings previously planted in this garden are unsorted. 
I think they should be most recent finished first 
screenshot_2017-08-06-13-17-07

Br3nda:
fixes #1383

### DIFF
--- a/app/views/gardens/show.html.haml
+++ b/app/views/gardens/show.html.haml
@@ -65,7 +65,7 @@
     %h3 Previously planted in this garden
     .row
       - if @finished_plantings.size.positive?
-        - @finished_plantings.each do |planting|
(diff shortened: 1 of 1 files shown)


pozorvlak:
I was going to say "I think this logic belongs in the controller or in a scope, not in the view", but then I looked and it *is* in the controller! It was added in 4726b57c, which appears to have been merged as part of Release 26, and should definitely be live on production now. I am confused.

Also, and relatedly, I think this needs a regression test.

Br3nda:
oh yes. i didn't even check the controller.
Let's close this one

PR #1407, opened 2017-10-20
Dev is almost always on port 3000

Br3nda:
The links on letter opener currently have port 8080. I don't know what this project has those - the default port for those running rails in dev is port 3000.
Let's fix up to match what I think most people are doing.


pozorvlak:
Unicorn listens on port 8080 by default - is anyone using that in development? I was for a while but now mostly use `rails s`. I think we can configure Unicorn to listen on port 3000 in development, but can't work out how right now. Docs are [here](https://bogomips.org/unicorn/).

pozorvlak:
```
Failures:
  1) Notifications GET /notifications works! (now write some real specs)
     Failure/Error: response.status.should be(302)
     
       expected #<Integer:605> => 302
            got #<Integer:401> => 200
     
       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/requests/notifications_spec.rb:9:in `block (3 levels) in <top (required)>'
```

Whaaaaaaaaa?

Br3nda:
Restated the Travis job and it passes

PR #1412, opened 2017-10-22
Basic read only api  - members, gardens, plantings, crops, photos, harvests, seeds

### LINKED ISSUE #582: Implement v1 crop API

### LINKED ISSUE #583: Implement v1 planting, harvest, and seed API endpoints
Based on discussions with developers, implement improved v1 planting/harvest/seed API endpoints, remedying any known problems with v0.

Br3nda:
I've marked all resources as immutable, so this is a read-only API adhering to rest and JSON API spec.

The gems available for authorization when modifying data are all based on pundit, so we should either roll our own or switch to Pundit. (i'd prefer we switched to Pundit because it gives more control and is easier to configure/understand).

fixes #583 
fixes #582
fixes #583
fixes #384

- [x] members specs
- [x] gardens specs
- [x] Plantings specs
- [x] crops specs
- [x] photos specs
- [x] harvests specs
- [x] seeds specs


Br3nda:
`Your bundle is locked to factory_girl (4.8.2), but that version could not be`....

something odd is going on - maybe a version got pulled.

Br3nda:
Looks like that's the last version
```
4.8.2 (October 20, 2017)
  Rename factory_girl_rails to factory_bot_rails
```
https://github.com/thoughtbot/factory_bot_rails/blob/master/NEWS

Good rename! i'll log an issue for us to migrate - in the mean time i've put the Gemfile.lock back to 4.8.0

Br3nda:
I have begun writing from request specs for these.

pozorvlak:
It would be nice to DRY this up further if possible. Is there any way the resources could be derived from the models, or from annotations on the models?

Br3nda:
The idea is that you don't automatically put everything on the model into the resources:
"Any of a resource’s attributes that are accessible must be explicitly declared. "
http://jsonapi-resources.com/v0.9/guide/resources.html#Attributes

This becomes useful when you get to v2 of the API, where relationships have changed, but the v1 API remains the same.

I don't think we can move this to annotations in the model files without straying into an unsupported way of using JSON API gem and making the model files complex.

It seems good to me that the attributes exposed are explicitly mentioned, not magically appearing, so save us making a terrible security mistake one day.

pozorvlak:
OK, makes sense. Can you add some tests to cover the other resource types?

Br3nda:
Sure thing - i'll make a bunch more request specs.

Br3nda:
Couple more request spec files to go

PR #1416, opened 2017-10-23
All crop photos

### LINKED ISSUE #514: Crop photos should include harvest photos
Crops currently show photos of plantings (eg. at the top of the crop detail page, in the crop thumbnail, etc). We should judiciously display a mixture of harvest and planting photos as appropriate. 

For instance, in the crop thumbnail, harvest photos are probably more distinctive.

On the crop detail page, we should show a mixture of both.

Br3nda:
crops#show has only the 3 most recent photos, even though we have much more
This adds a link to crop/photos, so the rest of the photos can be found

also fixes #514


app/views/crops/_photos.html.haml
```diff
@@ -1,5 +1,15 @@
+%h2 Photos of #{crop.name} harvests
+.row
+  - unless crop.harvest_photos.empty?
```
pozorvlak: Should we have some "nothing to show here" text if `harvest_photos` is empty? I guess it's consistent with what's there already, so if we're going to change it we should do so separately.

PR #1418, opened 2017-10-23
Randomise the crops on the homepage

### LINKED ISSUE #1331: home page crops never change

Br3nda:
fixes #1331


app/models/crop.rb
```diff
@@ -106,6 +106,10 @@ class Crop < ActiveRecord::Base
     end
   end
 
+  def harvest_photos
```
pozorvlak: I think this method belongs in #1416?

Br3nda:
Thanks for the reviews @pozorvlak . I've fixed up this PR to be a one liner.

PR #1436, opened 2017-11-06
Cropname on photos new - fixes #1433

### LINKED ISSUE #1433: Photos#new doesn't say which item you're adding the photo to
For those with short memory spans, it's easy to forget what you were adding a photo to.

logangingerich:
Adds a subheader to Photos#New that says "Choose Photo for ____" with the name of the item that you are adding to. 
screen shot 2017-11-06 at 7 47 11 am

fixes  #1433


logangingerich:
For some reason the failing tests here weren’t showing up for me locally, I’ll probably need to make some adjustments..

Br3nda:
Can it be more than the crop name? You're adding the photo to a planting or harvest or garden.

There is the item type and the item id already in the #new method, can you look up the item and then use the `.to_s` to display it?

app/controllers/photos_controller.rb
```diff
@@ -19,8 +19,9 @@ def index
   def new
     @type = params[:type]
     @id = params[:id]
-
     @photo = Photo.new
+    planting = Planting.find @id
```
Br3nda: It might not be a planting. It might be a garden

There's some code later in the same file, that turns the item type and id into the model itself
`
    item_class = Growstuff::Constants::PhotoModels.get_item(params[:type])
    item = item_class.find_by!(id: params[:id], owner_id: current_member.id)
`
That'll look up the actual model, whatever type it is.

logangingerich:
Ah, good to know. I was thinking photos were just on planting’s, so that’s where I was going wrong. I’ll try to get this updated as soon as possible

app/controllers/photos_controller.rb
```diff
@@ -21,6 +21,8 @@ def new
     @id = params[:id]
 
     @photo = Photo.new
+    item_class = Growstuff::Constants::PhotoModels.get_item(params[:type])
```
Br3nda: this is good -- do you think it can go into a method and the code be called from 2 places, instead of it repeating?

logangingerich: Definitely! I will refactor it soon.

pozorvlak:
It looks like the headless-browser tests are failing on Travis: have you [installed PhantomJS](https://github.com/teampoltergeist/poltergeist#installing-phantomjs)? We're using version 2.1.1.

We should probably document that as an optional dependency :-/

logangingerich:
I haven't yet install PhantomJS but I will definitely do that when I start working on the refactoring mentioned above!

Br3nda:
```
  2) PhotosController GET new assigns a planting id
     Failure/Error: @item = item_class.find_by!(id: params[:id], owner_id: current_member.id)
     
     ActiveRecord::RecordNotFound:
       Couldn't find Planting
     #
``` 
 The specs are probably not passing in a real id for the type. That's what is causing the page to break.

logangingerich:
Just wanted to let you know I'm still working on this. I've done some refactoring but am trying to figure out how to update the specs here to accurately test things with my changes (as @Br3nda says, they're currently not passing in a real id for the type)

pozorvlak:
@logangingerich no problem, thanks for letting us know :-)

logangingerich:
I'm not sure why I'm still failing the test: `rspec ./spec/features/signout_spec.rb:16 # signout after signout, redirect to signin page if page needs authentication` Seems completely unrelated to the changes I made but I'll keep working on it.

pozorvlak:
@logangingerich thanks for chipping away at this, and I'm sorry it's being frustrating!

Br3nda:
i'm getting the same error on your branch
```
Finished in 4 minutes 57.1 seconds (files took 7.26 seconds to load)
1505 examples, 1 failure, 2 pending
Failed examples:
rspec ./spec/features/signout_spec.rb:16 # signout after signout, redirect to signin page if page needs authentication
```

and i'm not sure why, but we'll need to work it out before we can merge

Br3nda:
debugging that
```
   30:       def self.get_item(type)
   31:         byebug
=> 32:         class_name = ALL.select do |model|
   33:           model[:type] == type
   34:         end[0][:class]
   35:         class_name.constantize
   36:       end

```
the variable named `type` passed in is `nil`

Br3nda:
this new change throws an exception when there's no id+type passes for photos#new

the existing signout spec visits visit "/photos/new", which works with the current code but doesn't make sense (you can't create a photo without an associated model, so users shouldn't be able to just navigate to photos/new).

logangingerich:
Thanks for working on this! I was a bit lost on how to fix it to pass the spec.

Br3nda:
one more spec.

Br3nda:
i've cleaned up some mess in the controller and the specs -- so it's passing now.  I added a feature spec.

there's some parts of the haml that @pozorvlak has commented on that still can be made simpler and less confusing.

Issue #1099, opened 2016-11-30
Deployment happens twice per merge


pozorvlak:
Both the ElasticSearch and non-ElasticSearch builds deploy on success; we should only deploy once (probably from the ElasticSearch build)?

Br3nda:
Are you heroku pipelines to deploy?

pozorvlak:
No, we're using [Travis CI's deployment hook](https://docs.travis-ci.com/user/deployment/heroku/).

Br3nda:
Okay. Maybe pipelines will help. I've used in @WhareHauora

PR #1439, opened 2017-11-09
Release 32

Br3nda:
For Friday 10 November


pozorvlak:
The Travis "static checks" build failed on the dev branch, with a Heroku error (from memory, something about only being able to run one Free-size dyno). I guess I should fix #1099, huh...

PR #1444, opened 2017-11-11
Release 33

Br3nda:
Planned for Friday 17/11.


Br3nda:
When this is merged, use "rebase and merge". Then we don't get the extra commit in master.

PR #1445, opened 2017-11-12
planting owner needs to match garden owner

### LINKED ISSUE #503: Refactor: "let" in rspec tests

### LINKED ISSUE #1342: Planting has different owner to its garden
The way factoryGirl creates models in our spec means many are running with member1 owning the Garden and member2 owning the planting in that garden.

This is an invalid state for plantings to be in. -- and we may be missing strange data interactions that are possible because of this

I propose we add validations to not allow a planting to have a different owner to its garden.
I'd like to go even further and have a planting delegate it's owner to it's Garden.

We never have plantings that aren't in a garden.

Br3nda:
it doesn't make a lot of sense for a planting to be in someone else's garden.
and it doesn't make sense to harvest from someone else's planting.

the UI doesn't let you do that, so the spec shouldn't be running in this way either.

I've introduced more of the `let` syntax too

fixes #1342 
fixes #503


spec/models/planting_spec.rb
```diff
@@ -148,12 +151,6 @@ def one_hundred_day_old_planting
     planting.owner.should be_an_instance_of Member
   end
 
-  it "owner isn't necessarily the garden owner" do
```
Br3nda: This test hints that plantings in someone else's garden were an allowed feature... not sure why.

cesy:
The only thing to consider here is community gardens, where the garden is owned by a community group and individual plantings may be owned by members of that group - it's something we always hoped to do, but haven't built yet. So let's bear that in mind to make this future-proof if possible, but your logic and description is right for now.

Br3nda:
We'd need a bigger permissions overhaul. I think a member to gardens permissions system. That would be a cool feature. Is there an existing story for that?

CloCkWeRX:
Raised a very large feature via #1451 for community gardens

Br3nda:
Can we merge this? It makes the plantings in our specs have  the same owner as their gardens finally, making them valid tests

PR #1449, opened 2017-11-13
Don't show lifespan of perennials

### LINKED ISSUE #1440: Median lifespan of perennial plants needs to include those that not finished
Right now the strawberries lifespans is showing "15 days" because it only includes finished plants in the calculation, and the finished strawberry plants are all failed crops.
http://www.growstuff.org/crops/strawberry

Median lifespan of perennial plants needs to include those that not finished?
Alternatively we don't display the lifespan of perennials.

Br3nda:
fixes #1440


Br3nda:
```
Errors on modified lines:
/home/travis/build/Growstuff/growstuff/app/models/crop.rb:1:1: C: Metrics/ClassLength: Class has too many lines. [181/179]
```

PR #1453, opened 2017-11-15
Fixes #1450 Adds a more prominent add photo button to garden and planting show views

### LINKED ISSUE #1450: Move photo buttons to be beside other actions
I did an informal user test, and the thing s new user struggled with most was finding the "add photo" button action.
I think it should be moved beside all the other actions for an object, like edit and delete.

logangingerich:
In the planting and garden show views, I added an extra 'Add photo' button next to the other actions. I didn't get rid of the original add photo buttons as I thought it would be easier for users just to have another option. I'm definitely open to styling or button placement changes.

fixes #1450

### DIFF
--- a/app/views/gardens/_actions.html.haml
+++ b/app/views/gardens/_actions.html.haml
@@ -21,3 +21,8 @@
               class: 'btn btn-default', id: 'delete_garden_link' do
       %span.glyphicon.glyphicon-trash{ title: "Delete" }
       Delete
+  - if can?(:edit, @garden) && can?(:create, Photo)
+    = link_to new_photo_path(type: "garden", id: @garden.id),
+    class: 'btn btn-primary' do
+      %span.glyphicon.glyphicon-camera{ title: "Add Photo" }
+      Add Photo
--- a/app/views/gardens/_actions.html.haml
+++ b/app/views/gardens/_actions.html.haml
@@ -23,6 +23,6 @@
       Delete
   - if can?(:edit, @garden) && can?(:create, Photo)
     = link_to new_photo_path(type: "garden", id: @garden.id),
-    class: 'btn btn-primary' do
+              class: 'btn btn-primary' do
       %span.glyphicon.glyphicon-camera{ title: "Add Photo" }
       Add Photo
--- a/app/views/gardens/show.html.haml
+++ b/app/views/gardens/show.html.haml
@@ -12,6 +12,12 @@
 .row
   .col-md-9
     %p.btn-group= render 'gardens/actions', garden: @garden
+    - if can?(:edit, @garden) && can?(:create, Photo)
+      %p
+        = link_to new_photo_path(type: "garden", id: @garden.id),
+        class: 'btn btn-primary' do
+          %span.glyphicon.glyphicon-camera{ title: "Add Photo" }
+          Add Photo
 
     - unless @garden.active
       .alert.alert-warning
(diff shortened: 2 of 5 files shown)


app/views/gardens/show.html.haml
```diff
@@ -12,6 +12,12 @@
 .row
   .col-md-9
     %p.btn-group= render 'gardens/actions', garden: @garden
+    - if can?(:edit, @garden) && can?(:create, Photo)
```
Br3nda: if this was in gardens/actions instead, would it appear in lots of logical places?

logangingerich: Good call, I'll work on it!

app/views/plantings/show.html.haml
```diff
@@ -67,6 +67,9 @@
     %p= render 'plantings/progress', planting: @planting, show_explanation: true
 
     = render 'plantings/actions', planting: @planting
+    - if !@planting.photos.empty? || (can?(:edit, @planting) && can?(:create, Photo))
```
Br3nda: I think this says, you can only have an "add photo" button if there already some photos.
The `if !@planting.photos.empty?`.

app/views/gardens/_actions.html.haml
```diff
@@ -21,3 +21,8 @@
               class: 'btn btn-default', id: 'delete_garden_link' do
       %span.glyphicon.glyphicon-trash{ title: "Delete" }
       Delete
+  - if can?(:edit, @garden) && can?(:create, Photo)
+    = link_to new_photo_path(type: "garden", id: @garden.id),
+    class: 'btn btn-primary' do
```
pozorvlak: The `class` here should be indented level with `new_photo_path`. I'm not actually sure this will work - HAML is whitespace-sensitive. Could you add a unit test for this behaviour? Ignore any "don't add new view specs" comments you find, we've been meaning to delete those.

app/views/plantings/show.html.haml
pozorvlak: I think this is harmless, but unnecessary - it says "if the planting has photos OR you can add a photo to the planting...", and then on the next line we check "if you can add a photo to the planting...". So we have the following truth-table:

 - no photos, can't add photo: no button
 - no photos, can add photo: show button
 - has photos, can't add photo: no button
 - has photos, can add photo: show button

So line 70 is doing nothing. Delete line 70, then this is good.

app/views/plantings/show.html.haml
```diff
@@ -67,6 +67,9 @@
...
+      - if can?(:create, Photo) && can?(:edit, @planting)
+        = link_to "Add photo", new_photo_path(type: "planting", id: @planting.id), class: 'btn btn-primary btn-xs'
```
pozorvlak: I see we're already testing for the presence of an "Add photo" button on this page (in `spec/views/plantings/show.html.haml`): @Br3nda, do you think we need a check for this additional button?

Br3nda: It's a bit confusing to read like that. Can it be any clearer?

logangingerich:
Thanks for the feedback! I should have time early tomorrow to fix things up.

Br3nda:
i've cleaned up some mess in the controller and the specs -- so it's passing now.  I added a feature spec.

there's some parts of the haml that @pozorvlak has commented on that still can be made simpler and less confusing.

Br3nda:
Awesome! Thank u

PR #1454, opened 2017-11-17
De-deprecate controller and view specs

### LINKED ISSUE #1132: Permit controller or view tests under specific criteria
#1127

I'd be +1 permitting controller tests for:
 - API/structured responses (xml, json, etc)
 - Security testing (ok vs 403 responses)

View tests at the page level would still be better as features (When I view a page, I see an X, click a Y)

View tests at the javascript component level or helper/widget level (Graph widget X, given this data, does Y, Z)

Outcome of this would be to add it to the coding style wiki and amend the various deprecation notices

Good talk on it via https://www.youtube.com/watch?v=MkPHntWZAPc - specifically using features for 'end to end' style tests on the absolute critical path, vs the risk of flakey behaviour (which has caused us much fun with PRs).

This concept ("The test pyramid") is useful for assessing what is better tested where
image

If we end up with lots of tests, the above talk introduces a companion concept, the "PUMA" test:
http://www.bbc.co.uk/blogs/internet/entries/ba0c030e-d031-4ab6-8ba6-3afe41807b55
image

The less a test meets that criteria, the more worth pushing it down the testing pyramid

pozorvlak:
We deprecated controller and view specs on the grounds that they were brittle, and were a poorer measure of user experience than feature specs. However, feature specs have their own problems: they're much slower to run, and flakier (see #901). We also ran into a few cases where feature specs erroneously passed because they were checking for the presence of a string that occurred in the error page!

Hence, we're cautiously un-deprecating controller and view specs.

Fixes #1132

### DIFF
--- a/spec/controllers/account_types_controller_spec.rb
+++ b/spec/controllers/account_types_controller_spec.rb
@@ -1,15 +1,3 @@
-## DEPRECATION NOTICE: Do not add new tests to this file!
-##
-## View and controller tests are deprecated in the Growstuff project.
-## We no longer write new view and controller tests, but instead write
-## feature tests (in spec/features) using Capybara (https://github.com/jnicklas/capybara).
-## These test the full stack, behaving as a browser, and require less complicated setup
-## to run. Please feel free to delete old view/controller tests as they are reimplemented
-## in feature tests.
-##
-## If you submit a pull request containing new view or controller tests, it will not be
-## merged.
-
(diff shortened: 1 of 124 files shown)


Br3nda:
```
/home/travis/build/Growstuff/growstuff/spec/views/gardens/show.html.haml_spec.rb:49:1: C: Style/EmptyLines: Extra blank line detected
```

PR #1460, opened 2017-11-19
Fix gardens actions, and add specs

Br3nda:
A bug slipped into all garden display pages - and there were no tests that caught it.
This adds feature specs for gardens#index, which would have detected the un-initialised
variable in haml

### DIFF
--- a/app/views/gardens/_actions.html.haml
+++ b/app/views/gardens/_actions.html.haml
@@ -21,8 +21,8 @@
               class: 'btn btn-default', id: 'delete_garden_link' do
       %span.glyphicon.glyphicon-trash{ title: "Delete" }
       Delete
-  - if can?(:edit, @garden) && can?(:create, Photo)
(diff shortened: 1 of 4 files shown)


Br3nda:
I've also added the missing "add photos" button to plantings/actions (with glypicons) so it appears on all sets of buttons for plantings. Removed it from plantings#show

Looks like this on thumbnails
image

and there on plantings#show
image

app/views/gardens/_actions.html.haml
```diff
@@ -21,8 +21,8 @@
               class: 'btn btn-default', id: 'delete_garden_link' do
       %span.glyphicon.glyphicon-trash{ title: "Delete" }
       Delete
-  - if can?(:edit, @garden) && can?(:create, Photo)
```
Br3nda: here's the variable that doesn't exist.

PR #1471, opened 2017-11-26
Remove default ordering scopes

### LINKED ISSUE #485: Get rid of ordering in default scopes
Turns out it's a bad idea to use default scopes for ordering: http://pragdave.blogs.pragprog.com/pragdave/2012/03/be-careful-using-default_scope-and-order.html. It also plays very badly with joins.  We should get rid of them, then create explicit ordered scopes (eg. Post.recent, Harvest.recent) and use those where needed.

### LINKED ISSUE #1383: Previously planted not sorted

Br3nda:
put the ordering in the controllers

fixes #485 
fixes #1383

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -5,7 +5,7 @@ class AccountsController < ApplicationController
 
   # GET /accounts
   def index
-    @accounts = Account.all
+    @accounts = Account.all.order(created_at: :desc)
     respond_with(@accounts)
   end
 
--- a/app/models/harvest.rb
+++ b/app/models/harvest.rb
@@ -127,6 +127,8 @@ class Harvest < ActiveRecord::Base
     photos.order(created_at: :desc).first || crop.default_photo
   end
 
+  private
+
   def crop_must_match_planting
     return if planting.blank? # only check if we are linked to a planting
     errors.add(:planting, "must be the same crop") unless crop == planting.crop
--- a/app/models/harvest.rb
+++ b/app/models/harvest.rb
@@ -124,7 +124,7 @@ class Harvest < ActiveRecord::Base
   end
 
   def default_photo
-    photos.first || crop.default_photo
+    photos.order(created_at: :desc).first || crop.default_photo
   end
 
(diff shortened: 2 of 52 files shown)


app/models/harvest.rb
```diff
@@ -39,7 +39,7 @@ class Harvest < ActiveRecord::Base
 
   ##
   ## Scopes
-  default_scope { joins(:owner) }
+  default_scope { joins(:owner) } # Ensures owner exists
```
Br3nda: this probably should move to a mixin (or whatever those are called in rails)

Br3nda:
I've started cleanup in controller specs to make up for removing some model specs.

app/views/plant_parts/index.html.haml
```diff
@@ -9,7 +9,8 @@
     - if plant_part.crops.empty?
       No crops are harvested for this plant part (yet).
     - else
-      != plant_part.crops.map { |c| link_to(c, c) }.join(", ")
+      - plant_part.crops.limit(100).each do |crop|
```
CloCkWeRX: Maybe worth sticking up into the controller the "decide how many things to show" parts? Not that we've done it here, but if we were doing pagination for this list it'd make sense there.

Br3nda: yeah, i agree. Will move

Br3nda: This page doesn't seem to have any links to it.... Maybe it serves no useful purpose?

app/controllers/accounts_controller.rb
```diff
@@ -5,7 +5,7 @@ class AccountsController < ApplicationController
 
   # GET /accounts
   def index
-    @accounts = Account.all
+    @accounts = Account.all.order(created_at: :desc)
```
Br3nda: this is curiously missing pagination

CloCkWeRX: Raised #1475

PR #1485, opened 2017-12-09
Show photos of a planting's harvests

Br3nda:
Moves the existing "thumbnail" of a harvest to "card" -- and makes a new partial called "thumbnail" that is just the thumbnail of the harvest

then, displays those thumbnails on the planting#show is it associated with.

image
image


app/views/plantings/show.html.haml
```diff
@@ -63,10 +64,11 @@
           %dt Last Harvest:
           %dd #{@planting.days_to_last_harvest} days after planting
 
-    %p= render 'plantings/harvests', planting: @planting
-    %p= render 'plantings/progress', planting: @planting, show_explanation: true
 
-    = render 'plantings/actions', planting: @planting
```
Br3nda: actions moved to the top

PR #1510, opened 2018-01-03
Use actions partial on plantings

Br3nda:


app/views/plantings/_card.html.haml
```diff
@@ -37,22 +37,4 @@
           %dd= planting.finish_predicted_at if planting.finish_predicted_at.present?
 
     %p= render 'plantings/progress', planting: planting, show_explanation: true
-
-    = link_to 'Details', planting_path(planting),
-      class: 'btn btn-default btn-xs'
-
-    - if can?(:edit, planting) && can?(:create, Harvest)
```
pozorvlak: The `can?(:create, Harvest)` check is missing from _actions.html.haml - is it unnecessary, or should it be ported across?

app/views/plantings/_card.html.haml
```diff
@@ -37,22 +37,4 @@
           %dd= planting.finish_predicted_at if planting.finish_predicted_at.present?
 
     %p= render 'plantings/progress', planting: planting, show_explanation: true
-
-    = link_to 'Details', planting_path(planting),
```
pozorvlak: This isn't present in _actions.html.haml - should it be?

pozorvlak:
@Br3nda: are the answers to my questions "no, this should not be present in _actions.html.haml" and "yes, it is unnecessary"? If so, let's merge it.

Br3nda:
You caught two things i missed. I'll fix them up before we merge this.

Br3nda:
Supersceed by #1550

PR #1518, opened 2018-01-09
Auto corrected by following FactoryBot/DynamicAttributeDefinedStatically

Br3nda:
Auto corrected by following FactoryBot/DynamicAttributeDefinedStatically

Click [here](https://awesomecode.io/projects/55/ruby_configs/14401) to configure it on awesomecode.io


spec/factories/harvests.rb
```diff
@@ -6,7 +6,7 @@
     plant_part
     planting nil
     owner { planting.present? ? planting.owner : FactoryBot.create(:member) }
-    harvested_at Time.zone.local(2015, 9, 17)
+    harvested_at { Time.zone.local(2015, 9, 17) }
```
CloCkWeRX: Thats... not really dynamic!

Br3nda: true! I think there's some tests that should be fixed to use TimeCop

PR #1527, opened 2018-01-13
Charts specs

Br3nda:


Br3nda:
The tests are green here.... So the gchart poltergeist errors are mysterious.

PR #1530, opened 2018-01-13
Auto corrected by following RSpec/ExampleWording

Br3nda:
Auto corrected by following RSpec/ExampleWording

Click [here](https://awesomecode.io/projects/55/ruby_configs/14389) to configure it on awesomecode.io


CloCkWeRX:
50-50 on this - I like 'should' style assertions with an approps context. Not everything fits but a lot does.

PR #1540, opened 2018-01-21
Charts controllers

Br3nda:
Charts are open data (read-only too) and don't need to complicate the existing controllers.
I've fixed the permissions errors on the crops chart data by moving to another controller
did the same for gardens
Moving the charts code out to another file also reduces the complexity of the gardens and crops controller.


pozorvlak:
I can only see merge commits on this PR, no actual code changes. What am I missing?

Br3nda:
that's odd... i wonder if it got caught in another PR

Br3nda:
The relevant commits are on the dev branch already

PR #1550, opened 2018-01-21
Ancestry of food.  Seeds -> Planting -> Seeds -> Planting

### LINKED ISSUE #535: Connect planting to seeds
When I plant something, and say that I planted it from seed, I'd like to be able to specify which seed (from my stash) I used.  See also #501 which is the equivalent story for plantings/harvests.

### LINKED ISSUE #1431: Actions for seeds. We only have delete.
Seeds records currently live forever. If the seeds have been planted, traded, lost or otherwise no longer available as seeds, the only action on growstuff is to delete them.

What actions should we have? What data is useful to assemble on seeds and what can we use it for?

Br3nda:
closes #1431 (Actions on seeds, other than just delete)
closes #535 (Connection plantings to seeds to plantings to seeds etc)

- [x] db structure
- [x] specs for relationships
- [x] Save seeds with their parent planting
- [x] Save plantings with their parent seed
- [x] Don't show finished/sown seeds as available to trade
- [x] Update seeds#index to differentiate between seeds you have, and seeds you used to have
- [x] UX for marking seeds all sown
- [x] feature specs for seed all sown
- [x] UX to harvest seeds from plantings
- [x] UX to show seeds a planting has produced
- [x] Modify existing seed->plant action to record the relationship


Br3nda:
i'm thinking seeds should have a type, which we can get from
```
  PLANTED_FROM_VALUES = [
    'seed', 'seedling', 'cutting', 'root division', 'runner',
    'bulb', 'root/tuber', 'bare root plant', 'advanced plant',
    'graft', 'layering'
  ].freeze
```

Br3nda:
Plantings#show
image

Br3nda:
Seeds#show

image

Br3nda:
The UX displayed is all to the left and it bugs me. There might be a better way.

pozorvlak:
Am I right in thinking "whakapapa a kai" can't be easily translated into English, but means something like "the genealogy of food"? Google Translate renders it as "food list", but that doesn't seem right.

Br3nda:
Ha. I didn't realise I'd not typed English. Yeah, it kinda translates to genealogy of food. Knowing where ancestors came from, not just the names but where and what they went through. The same with food.

Br3nda:
i don't know what Hound wants.... i can't see any comment that aren't outdated.

Br3nda:
Rebased

Br3nda:
Active planting:
image

Same planting, marked finished. Button vanish.
image

Br3nda:
So, this also includes an _actions partial for gardens, harvests, and plantings. There's harvests on the home page -- and there's tests for the home page too.
Only tradable && unfinished seeds appear on the home page.

Finishable is now a concern.

Br3nda:
Leaving this overnight incase @cesy wants to review again.

PR #1554, opened 2018-01-22
Add date_taken to photos

### LINKED ISSUE #1421: Order photos by date taken
currently we order by the local database :created_at, but both flickr and the jpeg have a date taken field. When the user imports several photos of a plant, we can order by the date the photo was taken, and then it appears to grow from photo to photo.

image

mftaff:
closes #1421


Br3nda:
Looks good. I think you forgot to check in `schema.rb`

mftaff:
Whoops..
Just added it now

app/controllers/plantings_controller.rb
```diff
@@ -31,7 +31,7 @@ def show
     @planting = Planting.includes(:owner, :crop, :garden, :photos)
       .friendly
       .find(params[:id])
-    @photos = @planting.photos.order(created_at: :desc).includes(:owner).paginate(page: params[:page])
+    @photos = @planting.photos.order(date_taken: :asc).includes(:owner).paginate(page: params[:page])
```
Br3nda: Can we keep it as descending order? Newest photos first.

mftaff: for sure! 
changing now.

app/controllers/plantings_controller.rb
```diff
@@ -31,7 +31,7 @@ def show
     @planting = Planting.includes(:owner, :crop, :garden, :photos)
       .friendly
       .find(params[:id])
-    @photos = @planting.photos.order(created_at: :desc).includes(:owner).paginate(page: params[:page])
+    @photos = @planting.photos.order(date_taken: :desc).includes(:owner).paginate(page: params[:page])
```
CloCkWeRX: Do we want to migrate old data (fall back on created at if null? Fetch from flickr again?)

Br3nda: What's the performance hit of that?


ALSO, i just realised we'd want to index that column too - so the query can order it faster

CloCkWeRX: Updating existing records in a migration to fall back to created_at or re-lookup from flickr - one off, no future perf considerations

Using COALESCE... its fairly fast, particularly if we added the index on the coalesce key itself.

Br3nda:
@mftaff Thank you for your work here. I rebased the dev branch and it's made a mess of your PR. I'll fix this up for you, probably later today.

Br3nda:
Replaced with #1568

PR #1563, opened 2018-01-28
Release 40 

Br3nda:


Br3nda:
I think the git conflict is because I squashed the paperclip patch.

I think if I debased dev against the master branch, it'll come right. I'll need to turn off the dev bra check restrictions to do this.

Br3nda:
I've rebased dev against `master`, and it's looking good again. 
The inflight PRs will need to be rebased too.

Br3nda:
I've tested and it's looking good on staging. Is anyone available to approve?

PR #1568, opened 2018-02-04
Add date_taken to photos

Br3nda:
This branch contains the work by @mftaff on #1554, with the newly rebased dev branch.
This PR should be much cleaner to merge


Br3nda:
I'll add re-fetching from flickr into this PR

Br3nda:
So - How do we want to handle photos that no longer exist on flickr? the jpegs won't render anyways so i'm thinking we delete them.
But i'll need to do it in a way that doesn't delete *all* photos just because flickr queries are failing. a "404 photo not found" is pretty similar to a "404 flickr changed their API method name" one day.

Br3nda:
For now, we just `say()` that the update failed. The photos without good flickr records will live on.

PR #1569, opened 2018-02-05
Removing shop, products, accounts, orders

### LINKED ISSUE #1552: Is it time to remove paid account functionality?
I don't think it works. I can't actually check out an order.
Could we do better with patreon or other subscription, not managed without our code.

Br3nda:
Pushing branch now so CI can run.

closes #1552


Br3nda:
I think this will make the rails 5 upgrades easier. Much less code to fix up.

CloCkWeRX:
Can we safely remove the activemerchant gem too?

Br3nda:
Oh I thought I did. Forgot to push

Br3nda:
pushed -- also removed config

Br3nda:
isn't it beautiful, how little code is left?

Issue #1581, opened 2018-02-10
Planting#show broken on prod 


Br3nda:
Works fine on staging.

Br3nda:
Possible the migration hasn't run

Br3nda:
i'm running it manually.

Br3nda:
it's crashing out when it hits a flickr account that is no longer authorized.

Br3nda:
fixed on prod now

PR #1582, opened 2018-02-10
Don't update flickr photos in migration

Br3nda:
fix for #1581

We can run the update of photos.date_taken manually


Br3nda:
Need to fix production, so I'm gonna a merge this.

PR #1591, opened 2018-03-14
Use photos helpers to find photos

Br3nda:
Moved photo image selection all into one file. Removes a lot of repeating code and then let me tidy up photo selection.
The photo for seeds was different on the #show and #index and home page. This makes it always the same. 
Also tidied up crop photo selection to look on planting, harvests, and then seeds. 
Added unit tests of the helpers.

### DIFF
--- a/app/models/crop.rb
+++ b/app/models/crop.rb
@@ -120,12 +120,7 @@ class Crop < ActiveRecord::Base
   # later we can choose a default photo based on different criteria,
   # eg. popularity
   def default_photo
-    # most recent photo
-    return photos.order(created_at: :desc).first if photos.any?
-
-    # Crop has no photos? Look for the most recent harvest with a photo.
-    harvest_with_photo = Harvest.where(crop_id: id).joins(:photos).order('harvests.id DESC').limit(1).first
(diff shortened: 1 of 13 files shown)


Br3nda:
Anyone wanna review a pull request?

app/models/crop.rb
```diff
@@ -120,12 +120,7 @@ def default_scientific_name
   # later we can choose a default photo based on different criteria,
   # eg. popularity
   def default_photo
-    # most recent photo
-    return photos.order(created_at: :desc).first if photos.any?
-
-    # Crop has no photos? Look for the most recent harvest with a photo.
-    harvest_with_photo = Harvest.where(crop_id: id).joins(:photos).order('harvests.id DESC').limit(1).first
-    harvest_with_photo.photos.first if harvest_with_photo
+    first_photo(:plantings) || first_photo(:harvests) || first_photo(:seeds)
```
cesy: Should this fall back to placeholder on the end, or is it okay not to?

Br3nda: It will fall back on placeholder. That's done in the helpers.

PR #1601, opened 2018-04-01
Feature/fluid layout

Br3nda:
Expanding the layout to fill the screen. 

image

Screenshots in each size:  (Most noticeable on the larger screens)

also fixes 

## Before

image

image

image

## after

image

image

image


Br3nda:
Also fixes bug causing a random-ish number of crops to appear onthe homepage.
Changed from `.reorder('RANDOM()')` to instead use shuffle.

PR #1604, opened 2018-04-01
Fix mentionbot config

Br3nda:


pozorvlak:
@Br3nda, would you mind splitting out the mentionbot config from the Ownable changes? I think it should be enough to rebase this branch onto #1605.

Br3nda:
That's where it went! I lost those changes and retyped it.
I'll fix up and push here again soon.

Br3nda:
@pozorvlak ready for review again

Br3nda:
I've learned it's not a config problem. The bot is broken.

PR #1607, opened 2018-04-01
Don't allow duplicate plant parts names

Br3nda:


pozorvlak:
Causes [a bunch of failures](https://travis-ci.org/Growstuff/growstuff/jobs/361200288#L2220) in `spec/features/harvests` :-(

pozorvlak:
Still failing :-( Looks like you need to explicitly ask `Faker` for a unique value: https://github.com/stympy/faker#ensuring-unique-values

pozorvlak:
Lots of ` Faker::UniqueGenerator::RetryLimitExceeded` errors - I guess we need a call to `Faker::UniqueGenerator.clear` somewhere in our test setup/teardown code - maybe a `config.before` block in `spec/spec_helper.rb` or `spec/rails_helper.rb`? Sorry, it's been a while since I've had to think about RSpec configuration.

Br3nda:
It's ... odd.. it ought to be clearing out plant parts as it goes anyways.
I think it's something else. It passes on my laptop but not on travis. I will keep poking it to find out why.

Br3nda:
`Faker::UniqueGenerator::RetryLimitExceeded:`

Looks like we'll need to add our own uniqueness.

Issue #869, opened 2016-05-16
Inconsistency: Show post doesn't show reply very obviously


CloCkWeRX:
Show post doesn't permit reply
- Visit posts index page (http://staging.growstuff.org/posts), reply to post
- Visit actual post (http://staging.growstuff.org/posts/clockwerx-20160422-does-this-timestamp-correctly-on-the-front-page-or-stay-as-1-minute-forever#comments) - no reply controls available.

maco:
The "Comment" button on the post page does what the "Reply" button on the index page does. However, I guess you do have to [scroll past a bajillion comments](http://growstuff.org/posts/skud-20130319-requests-for-new-crops) to reach that. Is that what this issue is trying to address?

CloCkWeRX:
Ah hah, missed it because it was below the fold/differently named. +1 to adding a comment/reply link straight under the post

cesy:
Okay, so we should add a second link so that it's obvious for people in both places.

PR #1621, opened 2018-04-19
Show second comment button when many comments on a post

jenkr55:
Fixes https://github.com/Growstuff/growstuff/issues/869

Added second Comment button and made it look like the other buttons in this area.

screen shot 2018-04-18 at 7 39 04 pm
screen shot 2018-04-18 at 6 01 11 pm

Made this happen when there are more than 10 comments but open to suggestions!


pozorvlak:
It's an issue we hoped we'd banished :-/ Do you remember which tests were
failing?

jenkr55:
Sounds good! I pushed some changes and here's what it looks like now
screen shot 2018-04-19 at 12 09 39 pm
screen shot 2018-04-19 at 12 10 02 pm

pozorvlak:
The Poltergeist failure is merely annoying, but the flaky controller test is worrying. Thanks!

PR #1664, opened 2018-07-03
Auto corrected by following quotes

Br3nda:
Auto corrected by following quotes

Click [here](https://awesomecode.io/projects/55/javascript_configs/753) to configure it on awesomecode.io


Br3nda:
added bootstrap-accessibility.min.js  to ignore list

PR #1666, opened 2018-07-04
Auto corrected by following semi

Br3nda:
Auto corrected by following semi

Click [here](https://awesomecode.io/projects/55/javascript_configs/754) to configure it on awesomecode.io


Br3nda:
I added this to ignores

PR #1694, opened 2018-09-08
Upgrade/elasticsearch

Br3nda:
Staging is currently on elastic search 6 (you can't install elastic search 2 for heroku anymore)
our code dependencies are for elastic search 2.

This is why the elastic search functionality is broken on staging. 

This upgrades us to elasticsearch 6. We'll need to upgrade the heroku resource on production too when we push this live.


Br3nda:
still gettings errors:
```[!!!] Index does not exist (Elasticsearch::Transport::Transport::Errors::NotFound)```

Br3nda:
Tumeke!

it works

Issue #1631, opened 2018-04-22
Crop#show displays child crop varieties that have not been approved 


Br3nda:
Only approved crops should display.

Br3nda:
Eg. Potato shows unapproved varieties right now .
Potato varieties

Br3nda:
http://www.growstuff.org/crops/potato

PR #1741, opened 2018-11-26
hides crop varieties that have not been approved

brandonbaker40:
Hides the crop varieties that have not been approved. Issue #1631

- prevents the list item from displaying in the crop-hierarchy if the crop has not been approved
- adds the 'hide' class if the crop has not been approved

/cc @Br3nda


CloCkWeRX:
LGTM :)

PR #1759, opened 2018-12-30
[WIP] Fix/extra loads

### LINKED ISSUE #1758: unnecesary calls to retrieve data from the database
the method `load_and_authorize_resource` will do something like this:
```
def show
  @project = Project.find(params[:id])
  authorize! :show, @project
end
```
see https://github.com/ryanb/cancan/wiki/authorizing-controller-actions

That makes a lot of our code unnecessary. Cancancan is already loading the data we need, and then we go and do it again, making a second call to the database in many places.

Br3nda:
Closes #1758 

The cancancan `load_and_authorize_resource` will load the expected resource for controller method. Yet in many of our methods, we load again. 

Generally, the authorise will have already blocked the user from continuing if they were doing something they aren't allowed to do. So we are still secure.

However, the act of pulling the same record we already have loaded is unnecessary. It slows us down and makes our code less dry 

So I've started finding these extra loads and removing. To assure myself that the controller still works I've started adding more controller specs. In particular, I'm adding tests for scenarios when a user is trying to do something they're not allowed to do, such as editing or deleting a different users growstuff data


Br3nda:
will need to be reviewed after rails5 branch merges

Br3nda:
i'll reopen after rails5 merges

app/controllers/posts_controller.rb
```diff
@@ -4,54 +4,40 @@ class PostsController < ApplicationController
...
+    @posts = @posts.where author: @author if @author.present?
+    @posts = @posts.order(created_at: :desc).includes(:author, comments: :author).paginate(page: params[:page])
     respond_with(@posts)
   end
 
-  # GET /posts/1
-  # GET /posts/1.json
-  # GET /posts/1.rss
   def show
-    @post = Post.includes(:author, comments: :author).find(params[:id])
```
CloCkWeRX: This is likely more optimised than the load_and_authorise_resource version, as it would generate a left join instead of multiple DB calls

PR #1768, opened 2019-01-01
Upgrade heroku stack

Br3nda:
Based on @eoinkelly's work in https://github.com/ODNZSL/nzsl-online/pull/480

The version of the heroku stack is old on both staging and prod. Prod uses cedar-14 and staging uses heroku-16

by moving this into heroku's `app.json` file, we will be able to both upgrade it from a PR, and also keep staging and prod in sync.

This also adds enough info for someone else to quickly deploy their own version of growstuff.org on heroku if they want to.


cesy:
Do we want to do this at the same time as getting the Rails upgrade live, or finish that first for easier investigation of issues?

Br3nda:
I'm happy to do both at once. Prod and staging are out of sync, so if like to test on cedar-18 on staging, and know we will be deploying the same to prod.

PR #1778, opened 2019-01-05
Fix modified schema deploy error

Br3nda:
Travis isn't deploying to staging - instead, it has an error about the schema file changing.

The error Travis-CI during deploy is:
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)
	modified:   db/schema.rb
```

I've re-run migrate, and the schema version format has changed in Rails 5.2. Here's the modified file needed.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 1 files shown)


Br3nda:
I merged this, so we can fix staging deploys. It seems trivial enough

PR #1781, opened 2019-01-07
WIP - pushing the commits one by one, to see which fails on deployment to heroku

Br3nda:


Br3nda:
running a git bisect instead.

PR #1796, opened 2019-01-18
Fixing the app.json for Bonsai

Br3nda:
Heroku can't find an app called "bonsai-elasticsearch"

Also, we should match the production version


cesy:
Looks like this is the sparkpost code?

Br3nda:
Yes, i merged in commits from anotjer pr so i could test this using heroku's or review apps. Without the sparkpost chnages, it fails to build because sparkpost is no linger available for purchase on the marketplace

PR #1804, opened 2019-01-27
Tidy up display on gardens#index

Br3nda:
The new menus on gardens#index aren't lined up correcly.

before (current production)
image

after (this pr)
image

Th buttons should line up on the top right, at any size screen now.


Br3nda:
Further tidy up the buttons
image

PR #1818, opened 2019-02-07
Fixes checking for github bots in contributors

Br3nda:
Fixes all the travis-ci fails for dependabot PRs.


Br3nda:
Good idea. I've changed it to read in Ruby.

PR #1858, opened 2019-03-11
Remove staging file and re-instate memcachier

Br3nda:
The GrowStuff codebase has a separate staging.rb and production.rb

this means our staging isn't quite the same as production. We could have a wildly different set up on staging and therefore not catch issues there and only break once it hits production

Instead, let's have an identical set up on staging and production.
This is usually done by having Heroku treat staging the same as production. Any bad config in production.rb would be noticed as soon as we test on staging.
Any difference between the two can be in ENV variables.

There already was different cache settings between the 2 instances - this change would ensure we have the same memcache set up on both staging and production.


config/environments/production.rb
```diff
@@ -102,8 +111,8 @@
 
   # this config variable cannot be put in application.yml as it is needed
   # by the assets pipeline, which doesn't have access to ENV.
-  config.mapbox_map_id = 'growstuff.i3n2c4ie'
-  config.mapbox_access_token = 'pk.eyJ1IjoiZ3Jvd3N0dWZmIiwiYSI6IkdxMkx4alUifQ.n0igaBsw97s14zMa0lwKCA'
+  config.mapbox_map_id = ENV['MAPBOX_ID']
+  config.mapbox_access_token = ENV['MAPBOX_ACCESS_TOKEN']
```
CloCkWeRX: Missing actionmailer configuration?
```config.host = 'staging.growstuff.org'
config.action_mailer.default_url_options = { host: 'staging.growstuff.org' }```
?

CloCkWeRX:
+1 to the concept; though I reckon anything we pull out to an ENV should get documentation in a README or DEPLOYMENT / wiki; incase whatever we have configured has to be setup from scratch again.

config/environments/production.rb
Br3nda: Added

Br3nda:
Rebased

config/environments/production.rb
Br3nda: Line 92 `config.action_mailer.default_url_options = { host: ENV['MAIL_SENDER_HOST'] }`

Br3nda:
rebased

PR #1865, opened 2019-03-13
[HOTFIX] Patch Rails

Br3nda:


Br3nda:
This is security fix, so going live

PR #1883, opened 2019-03-24
Production release

Br3nda:
* Gardens now have GardenTypes


Br3nda:
Sneakily adding https://github.com/Growstuff/growstuff/pull/1883/commits/ad90dd6dfaf402002da2a95135dc02d967d68a2e

This is blocking new users from confirming their account (unless they think to add www in front of the url in confirmation emails)

PR #1918, opened 2019-04-13
Release 51

Br3nda:
* fact cards on plantings#show
* restore memcache
* replace elastic search with searchkick


Br3nda:
Let's release this by next Wednesday

Br3nda:
Note for release: The searchkick reindex on staging resulted in errors:
`"The bulk request must be terminated by a newline" `

I fixed it by doing this form the console
```ruby
Crop.all.each_slice(5) do |batch|
  Crop.searchkick_index.import(batch)
end
```

I tried it with a slice/batch of 200 and got the same errors
I suspect there's a batch limit in batch indexing on bonsai.

Br3nda:
Fix for reindexing in  #1927

PR #1936, opened 2019-04-22
Timeline

Br3nda:
Tinkering with timelines. feedback welcome.
image
image
image


milesgould:
That looks great!

Br3nda:
I'm trying to work out what people most want from recording this data. I think a page showing you what is going to be ready next would be more useful than a profile page for a planting. I'm gonna test a few more things.

CloCkWeRX:
I like the UI too, +1 to this.

PR #1937, opened 2019-04-22
Add percy - visual difference detection

Br3nda:
Adding a "visual differences" tool, to make it easier to review changes to look/layout


Br3nda:
Click through to see the differences introduced in this code change
https://percy.io/Growstuff/growstuff/builds/1769026

Note: The random crops on the homepage are being detected as a difference.

PR #2006, opened 2019-06-27
Release 55

Br3nda:


Br3nda:
`ActionView::Template::Error (nil can't be converted to a Time value):`
On the member profile pages. 
Probably for records wirh planted at nil

Br3nda:
Needs #2007, #2008

PR #2057, opened 2019-07-08
[WIP] Start tracking seed trades

### LINKED ISSUE #605: Track seed trades

### LINKED ISSUE #946: Seeds available for trade shows me seeds I cannot trade for
http://growstuff.org/ -> "Seeds available for trade" lists a number of trades, all of which are 'local trade only', but are on another continent (Australia vs America).

For a logged in member with a location set, we should prefer showing seed trades (local) near their location (we already do this on the places page, though there appear to be a few bugs there as well).

IE:
http://growstuff.org/places/Adelaide shows 3x crops/seeds available for trade (but not the actual seed-trades themselves - we should maybe make the same UI as the frontpage appear here)

### LINKED ISSUE #1906: Record trades
As a member of growstuff who gives seeds to someone else, i want to record the trade, and invite the recipient to join growstuff, with these seeds already added to their stash.

Br3nda:
closes #605
closes #1906
closes #946


Br3nda:
image

Br3nda:
Ah damn. I meant to merge a different PR

Br3nda:
I'll remove these commits

PR #2065, opened 2019-07-09
Fixing mixed content and moving links to use https

Br3nda:


Br3nda:
The map/leaflet code sure repeats alot. I think it needs a refactor/clean up.

PR #2070, opened 2019-07-10
fix display of badges (colours)

Br3nda:
we were overriding the colours


Br3nda:
image
image


image
image


image

image

Br3nda:
the colours are back:
image

PR #2077, opened 2019-07-11
Fixes markdown on staging that links to production

### LINKED ISSUE #2072: growstuff crop markdown always links to production
if you use this syntax on staging:
[tomato](crop)

you get a link to 
http://growstuff.org/crops/tomato

this means people testing staging can accidentally stray onto production and not realise they're editing production data

- [ ] fix links to be to the same site
- [ ] if full URL is still needed, then use https

Br3nda:
fixes #2072. by not using full urls in growstuff markdown links to crops
Also moves to reading host from env, so we can have different ones on staging and prod

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -100,7 +100,7 @@ Rails.application.configure do
   }
   ActionMailer::Base.delivery_method = :smtp
 
-  config.host = 'growstuff.org'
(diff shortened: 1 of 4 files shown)


Br3nda:
Yes. They were already set when we checked

PR #2078, opened 2019-07-12
use most liked photo as default photos

### LINKED ISSUE #540: Show photos based on "likes"
I want to be able to "like" a photo (i.e. click on a little icon/link to indicate that I think it's a good pic). These "likes" should be recorded and used to choose which photos to display on the crops page.  This solves the problem of often having pictures of indistinguishable seedlings as the default crop photo.

When member A "likes" member B's photo, member B should get a notification.

See also: list of "reasons" on #541.

Br3nda:
Closes #540


Br3nda:
I think we'd ban those users.

Which we need to build too. There some spam on the forums right now that need a flag and ban.

pozorvlak:
> I think we'd ban those users.

Yeah, that too :-)

PR #2085, opened 2019-07-13
Renamed photographings to photo associations

Br3nda:
now it matches the name of its controller, and it is clearer to understand. This feels more railsy


CloCkWeRX:
Does this need a table rename migration too?

db/migrate/20190130090437_add_crop_to_photographings.rb
```diff
@@ -7,4 +7,9 @@ def change
       p.set_crop && p.save!
     end
   end
+  class Photographing < ApplicationRecord
```
CloCkWeRX: While this works, I tend to avoid classes in schema migrations/management to avoid this; and push data migrations to rake tasks or rely on raw sql.

PR #2139, opened 2019-08-19
Production release 

Br3nda:
* new threaded inbox
* sent messages 
* untrash/undelete messages.
* migration from paranoia to discard


Br3nda:
Tested. Works nicely. Let's release this today

PR #2200, opened 2019-10-05
📝 Inline editing and display tidy ups

### LINKED ISSUE #1616: Edit in place for Harvests

Br3nda:
closes #1616 #1618
This is just jquery for now. I think we should add ReactJS or vue and make some components

image
image
image
image

image
image
image
image


Br3nda:
this can be tested at https://robotgarden.herokuapp.com/

PR #2202, opened 2019-10-06
Administer member's roles

### LINKED ISSUE #470: Role administration
As an admin of the site, I should be able to add and remove admins and crop wranglers (and any other roles we come up with).

Br3nda:
closes #470


app/controllers/admin/members_controller.rb
```diff
@@ -14,13 +17,24 @@ def destroy
...
+    end
+
     private
 
     def search_term
       params[:q]
     end
 
-    def auth!
+    def admin?
```
CloCkWeRX: I feel like require_admin! or similar (retaining the !) is a bit more informative (this will explode and spray exceptions everywhere if it fails); while I'd expect a is_admin? type method to return a bool.

Br3nda: i agree. I'll fix this

Issue #480, opened 2014-12-12
Delete many notifications at once


Skud:
instead of having to delete every notification individually, it would be nice if there were checkboxes and you could delete all checked items.  There should also be a "select all" option.

bestest-mensch:
@Br3nda Is this still something that needs to be worked on? If so I'd like to pick up this issue.

Br3nda:
please go ahead - they're messages now as part of mailboxer gem.

bestest-mensch:
@Br3nda What do you think about something like this?

Web:
![Screen Shot 2019-10-05 at 10 29 02 AM](https://user-images.githubusercontent.com/16200338/66257150-085f1500-e75b-11e9-991e-fe5e4580b2de.png)

Mobile:
![Screen Shot 2019-10-05 at 10 28 47 AM](https://user-images.githubusercontent.com/16200338/66257151-0ac16f00-e75b-11e9-940a-0d4ac32438ee.png)

Br3nda:
Awesome. Yes

PR #2203, opened 2019-10-06
add ability to delete multiple conversations at once

bestest-mensch:
Adds the ability to delete multiple conversations/messages at once #480. I noticed a couple other small issues I could address on the conversations page. Wasnt sure if we wanted to keep a strict scope on this or not.

Please let me know if I missed any steps in the contribution workflow. I had issues getting this spec in particular to pass, but am unsure how I would have impacted it.
spec/features/gardens/gardens_spec.rb:67

### DIFF
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -33,6 +33,13 @@ class ConversationsController < ApplicationController
     redirect_to conversations_path(box: params[:box])
   end
 
+  def destroy_multiple
+    Mailboxer::Conversation.where(id: params[:conversation_ids]).each do |conversation|
+      conversation.move_to_trash(current_member)
+    end
+    redirect_to conversations_path(box: params[:box])
+  end
+
   private
 
   def mailbox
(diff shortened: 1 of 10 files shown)


spec/features/conversations/index_spec.rb
```diff
@@ -20,8 +20,8 @@
 
     describe 'deleting' do
       before do
-        # delete button
-        click_link class: 'delete'
+        check 'conversation_ids[]'
+        click_button 'Delete Selected'
```
Br3nda: needs an `expect` to actually test something here.

app/controllers/conversations_controller.rb
```diff
@@ -33,6 +33,13 @@ def destroy
     redirect_to conversations_path(box: params[:box])
   end
 
+  def destroy_multiple
```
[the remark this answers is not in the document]
bestest-mensch: I added coverage for this and modified the action.

spec/features/conversations/index_spec.rb
bestest-mensch: Thanks. There were already 3 expectations that occurred after this before block. I added a couple additional specs for coverage that was missing.

PR #2221, opened 2019-10-19
📧 Polish planting reminders stylings and data

Br3nda:
image
image


app/mailers/notifier.rb
```diff
@@ -25,15 +25,34 @@ def notify(notification)
...
+          @late << planting
+        elsif planting.harvest_time?
+          @harvesting << planting
+        else
+          @others << planting
+        end
+      end
+    end
+
+    @subject = "Your #{Date.today.strftime('%B %Y')} #{@sitename} progress report"
```
Br3nda: i've left this as "Date.today", because Time.zone.today would use the server's timezone, which doesn't make much sense.

PR #2275, opened 2019-11-15
Add marker for leaflet maps, and use asset pipeline to serve file

Br3nda:
Map icons are broken on prod, have been for a long time. but they work on dev!

image

This changes the leafet js to use the rails asset pipeline to find the marker file. So the file can be found on production too.

I've had to set the marker, so i used a spade. because why not.

image

I've also fixed the missing `var` in the javascript, and fixed the scss/css to pass the linter tests.


Br3nda:
i've fixed the members#show map too

Br3nda:
and on the crops#show it'll use the crop's icon - unless the crop doesn't have an icon then it'll use the sprouts

image
image
image

Br3nda:
image
image

Br3nda:
image

Br3nda:
added photos
image

Br3nda:
image

PR #2286, opened 2019-11-23
Predict month of harvest on perennials [EDIT: show nearby harvesting months for all crops]

### LINKED ISSUE #1441: Predict harvest months for perennial plantings
Since release 32 we predict the first harvest of a planting. If it is perennial that is currently the last prediction we make. 

If a planting is more than a year old, we can predict future plantings. For example, this tree has fruit every April http://www.growstuff.org/plantings/shiny-the-deck-feijoa

This should be specific to a planting (for now) as all out plantings are in different climates.

Br3nda:
If the planting has harvested before, show those months as harvesting months

if the planting has no recorded harvests (e.g. if was only just planted) then display months based on the same crop growing at the same/nearby latitude.

image

closes #1441

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
 after_script:
   - >
     if [ "${COVERAGE}" = "true" ]; then
-      gem install codeclimate
+      gem install codeclimate-test-reporter
       codeclimate-test-reporter
     fi
   - >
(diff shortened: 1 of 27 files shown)


Br3nda:
image

Br3nda:
added the neighbouring plants to plantings#show, because they'll help us debug if we get wierd results later
image

Br3nda:
image

Br3nda:
image

PR #2288, opened 2019-11-23
Production release

Br3nda:
Let's get these caching improvements live


Br3nda:
tested on staging. huge improvement

PR #2303, opened 2019-11-30
Reduce number of db queries

Br3nda:
reduced home page queries from 91 to 12


Br3nda:
Most of the queries are in the default_photo method. It's going to need more re-factoring, and probably making the code less DRY.

Issue #1640, opened 2018-05-10
We're frequently exceeding our RAM quota on production.


pozorvlak:
![growstuff_ram](https://user-images.githubusercontent.com/225991/39879830-c8c221c4-5473-11e8-9d45-b0a1e84d6bd1.PNG)

Here's [Heroku's guide to optimizing memory usage for Ruby](https://devcenter.heroku.com/articles/ruby-memory-use).

Br3nda:
Do we know which page?

Br3nda:
Might be  #1638

pozorvlak:
They could definitely be related, but I think the causality might go the other way - exceeding our RAM quota means we're hitting swap, which will slow page-loads way down. Optimising queries may very well help, but adding caching is unlikely to reduce our RAM usage :-)

Br3nda:
i didn't think heroku gave you swap - in my experience it kills the app when you exceed the allocated resources you've paid for.

pmackay:
I notice the whole app runs on one dyno - would using a worker be possible? Or would that create a budget problem?

Br3nda:
i think the individual dyno would still exceed it's RAM quota.

There is a new relic - there may be clues on there on which pages/code is being memory hungry

apdarr:
@Br3nda That's right, on the [Common Runtime](https://devcenter.heroku.com/articles/dynos#common-runtime-networking), where the app runs, dynos can use swap when exceeding memory usage. Hobby dynos run on multitenant machines, which can sometimes mean performance greater than expected if the host is underutilized by competing Heroku apps. 

 @pmackay If the app has or begins to emit [H12 timeout errors](https://devcenter.heroku.com/articles/error-codes#h12-request-timeout) on Heroku, then it's likely time to think about adding a background process if budget allows. Using something like [rack-timeout](https://github.com/heroku/rack-timeout) can also provide an interim reprieve. [The Scout](https://elements.heroku.com/addons/scout) add-on is free and is specifically designed for identifying memory issues and N+1 queries in Rails apps. That combined with New Relic should provide a decent start. 

There are a few other approaches to tackle memory issues, but I thought I'd start with the above info!

pozorvlak:
@apdarr Thanks!

Br3nda:
The biggest memory user so far is the comfy Mexican sofa (on staging). Let's check again in a week

Br3nda:
[...]

PR #2320, opened 2019-12-09
Reading #index lookup from ElasticSearch instead of the database

### LINKED ISSUE #2315: Keep crops photo in elastic search.
The biggest cause of db queries is the logic to find a photo on a card 

Crops are indexed in elastic search. Let's store the photo url there too, and then display the photo without querying the postgresql database.

### LINKED ISSUE #2326: Rewrite default_photo code to not make lots of db queries.
Part of #1640

### LINKED ISSUE #2331: rspec: [searchkick] WARNING: Records in search index do not exist in database: 61, 98, 100, 103, 107, 110, 113, 114, 115, 116, 117
See https://travis-ci.org/Growstuff/growstuff/jobs/625189047?utm_medium=notification&utm_source=github_status

Br3nda:
Indexes Harvests, Seeds, Plantings, Photos into Elastic search and then uses those and the existing Crops data in ES to display #index pages and the home page.

Also includes child records (e.g. how many likes a photo is indexed as `likes_count` ) in the elastic search indexes, so we can display those without querying the database. 

Modified the likes code, to not do n+1 dB queries in the lookup in the view.

Part of fixing #1640

Closes  #2315
Closes #2326
fixes #2331 

## Notes for reviewers:
This PR gets records from elasticsearch and doesn't turn them into activerecord models. This means we can do a whole page render, including child records, without hitting the slower PostgreSQL database.  
To achieve this, we're using duck typing - the data we get from the searchkick elastic search wrapper is pretty close to what an activemodel object would look like, but not exactly. We can't do `link_to planting`, so instead, we do `link_to planting_path(slug: planting.slug)` etc.

We can't have an attribute with a `?` e.g. `active?` and `tradable?`. These are not supported as valid field names in elasticsearch, so those methods have had their `?` removed. This is the change I'm least happy with because it's a nice rail/ruby feature but it's also very unique to ruby, hence why other tools don't support it.


app/models/concerns/crop_search.rb
```diff
@@ -21,6 +18,7 @@ def search_data
         scientific_names: scientific_names.pluck(:name),
         # boost the crops that are planted the most
         plantings_count:  plantings_count,
+        harvests_count:   harvests.size,
```
Br3nda: needs a counter cache for this

app/models/concerns/search_harvests.rb
```diff
@@ -29,18 +29,18 @@ def search_data
...
+        planting_slug:    planting&.slug,
 
-        #photo
+        # photo
         has_photos:       photos.size.positive?,
         thumbnail_url:    default_photo&.thumbnail_url || crop.default_photo&.thumbnail_url,
 
         # counts
-        harvests_count:   harvests_count,
+        # harvests_count:   harvests_count,
```
CloCkWeRX: Dead? If so, removable

Br3nda:
Only 2 tests left to fix
```
 1) home page when anonymous show tradeable seed

     Failure/Error: expect(subject).to have_link href: seed_path(tradable_seed)

       expected to find link nil with href "/seeds/bleokydc-magic-bean" but there were no matches

     Shared Example Group: "shows seeds" called from ./spec/features/home/home_spec.rb:89

     # ./spec/features/home/home_spec.rb:40:in `block (3 levels) in <top (required)>'

     # ./spec/spec_helper.rb:59:in `block (3 levels) in <top (required)>'

     # ./spec/spec_helper.rb:57:in `block (2 levels) in <top (required)>'

  2) home page when signed in show tradeable seed

     Failure/Error: expect(subject).to have_link href: seed_path(tradable_seed)

       expected to find link nil with href "/seeds/eytsgkjv-magic-bean" but there were no matches

     Shared Example Group: "shows seeds" called from ./spec/features/home/home_spec.rb:98

     # ./spec/features/home/home_spec.rb:40:in `block (3 levels) in <top (required)>'

     # ./spec/spec_helper.rb:59:in `block (3 levels) in <top (required)>'

     # ./spec/spec_helper.rb:57:in `block (2 levels) in <top (required)>'


```

app/models/garden.rb
```diff
@@ -77,6 +77,8 @@ def mark_inactive_garden_plantings_as_finished
     end
   end
 
+  def reindex; end
```
Br3nda:
```suggestion
  def reindex(refresh: false); end
```

app/views/plantings/_harvests.html.haml
```diff
@@ -10,11 +10,11 @@
       = link_to harvests_path(return: 'planting', harvest: {crop_id: @planting.crop_id, planting_id: @planting.id, plant_part_id: plant_part.id}), method: :post, class: 'dropdown-item' do
         = plant_part.name
 
-- if planting.harvests.empty?
+- if @planting.harvests.empty?
```
Br3nda:
```suggestion
- if planting.harvests.empty?
```

Br3nda:
We've had several rubocop upgrades recently, so there's also many changes to satisfy the new linter checks that comes with the newer rubocop

app/controllers/photos_controller.rb
```diff
@@ -13,18 +13,22 @@ def show
...
+      where = { planting_id: @planting.id }
     end
-    @photos = @photos.order(created_at: :desc)
-      .includes(:owner)
-      .paginate(page: params[:page])
+
+    @photos = Photo.search(
+      load:     false,
+      boost_by: [:created_at],
+      where:    where,
```
simicic: I was wondering, why is the search done by `id` if we get the `slug` for searching all of these entities - (Crop, Planting, Member, etc)? 
One call to the DB could be saved if  ES would store slug instead of id (or both).
What am I missing?

PR #2329, opened 2019-12-15
Reducing memory profile of the web app

Br3nda:
Remove newrelic - profiler says it's using a lot of memory, which is normally fine but we're over the 500MByte cap on heroku

Part of #1640


Br3nda:
Removed bootstrap datepicker. the built in datepicker in a modern browser is just as good now.

CloCkWeRX:
> The job exceeded the maximum time limit for jobs, and has been terminated.
That's a bit worrying, hopefully flakey.

Br3nda:
I'm not going to be sure we've reduced it yet. Only paid heroku tiers have metrics, Ave staging isn't paid

Br3nda:
I'm gonna split this. Removing the datepicker is a bigger task, because the quick finish button does some nifty stuff.

Br3nda:
i'll reopen this when it's ready

PR #2344, opened 2019-12-20
show days of progress

Br3nda:
i need the days, because some data on crops show predicted finish in days. So here's an addition of the days grown versus days expected, so i can tell if the potatoes are ready for christmas


Br3nda:
image

PR #2350, opened 2019-12-26
Routes using slugs (instead of ids)

Br3nda:
this seems to be intended by the models having slugs (added about 7 years ago) but the routes weren't using them

copied from #2320

### DIFF
--- a/app/controllers/data_controller.rb
+++ b/app/controllers/data_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class DataController < ApplicationController
+  abstract
+  before_action :authenticate_member!, except: %i(index show)
+  load_and_authorize_resource id_param: :slug
+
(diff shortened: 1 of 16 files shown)


app/controllers/data_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class DataController < ApplicationController
```
Br3nda: @CloCkWeRX I'd like your opinion on this. It's holding the repeating code from the garden, harvest, seed, and plantings controllers. Putting an end to typos in one controller making iit inconsistent.
It could be a mixin instead - but an inherited controller methods seems more railsy. not sure tho.

CloCkWeRX: I think it makes sense. If we had two competing concerns 'controller with slugs' and 'controller with xyz'; that'd be the point to refactor to a mixin.

PR #2367, opened 2020-01-06
likes not needing to look up db record

Br3nda:
from #2320

### DIFF
--- a/app/views/photos/_likes.html.haml
+++ b/app/views/photos/_likes.html.haml
@@ -1,14 +1,13 @@
 %span.likes
-  - if member_signed_in?
-    - if can?(:new, Like) && !photo.liked_by?(current_member)
-      = link_to likes_path(photo_id: photo.id, format: :json),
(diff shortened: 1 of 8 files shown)


app/views/photos/_likes.html.haml
```diff
@@ -1,14 +1,13 @@
...
-      = link_to likes_path(photo_id: photo.id, format: :json),
-                method: :post, remote: true, class: 'photo-like like-btn' do
-        = render 'likes/count', likeable: photo
+  - if member_signed_in? && can?(:new, Like)
+    - if !photo.liked_by_members_names.include?(current_member.login_name)
+      = link_to likes_path(type: 'Photo', id: photo.id, format: :json),
+                method: :post, remote: true, class: 'photo-like like-btn ' do
+        = render 'likes/count', likeable: photo, liked: false
     - else
-      - like = photo.likes.find_by(member: current_member)
```
Br3nda: this is the extra lookup on every display - that's what we're removing the need for, in this PR

Issue #2369, opened 2020-01-06
Improve gardens index page for new (all) members


simicic:
As a new member seeking inspiration, `https://www.growstuff.org/gardens` is not really inspiring, gardens links, inactive checkbox and jump to cover half of the page (not to mention how it looks on mobile)

**Some of the easy fixes could be**:  
1. Navigation for "My gardens", "Everyone's gardens" and "Add a garden" does not need to be a card element - it can simply be links next to each other underneath the title.
2. Inactive checkbox can follow and be next to the rest of the filters.
3. "Jump to" can be removed? 3 lines of jump to names like `2019` and `8x2 bed` is less valuable than inspiring garden photos and list of plantings.
4.  For mobile keep garden cards as is, on web garden cards can be modified into: column for photo, owner and actions and other column next to it for plantings (limit number of plantings to keep the size unified and add a line for perennials)

I would love to share Growstuff with my friends but currently style looks quite outdated and not engaging. I think when you register (or explore) a new website you do seek inspiration or a tool that can help you improve how you do things. Being inspired by people around can be a good start to engage new users - therefor - Gardens#index is an important page. 

Let me know what you think!

CloCkWeRX:
#513 would help a bit with this, especially since flickr has all sorts of arbitrary limits on it these days; coupled with a sort by recent plantings as a default link.

We should probably also automatically mark things inactive with no plantings in > 2 years or something like that; to get the actually active ones.

simicic:
@CloCkWeRX thanks for the response, #513 will not solve the issue with half a page being irrelevant content - but it I do agree it will help in engaging users. However, not sure how you pay for hosting, but having Flickr as a host for the images is actually quite nice. (I say this without knowing the plan for the platform in the future)

Marking gardens as inactive is also useful and can be a separate issue. 
Still, do you think I could pick some of the quick fixes as suggested? (I also joined  gitter - we can discuss it further there)

Br3nda:
agreed. that page could be much better.

Br3nda:
I agree the jumpto should be removed, unless you're viewing your own gardens.

I want the whole page redone, really. It is useful as an overview of your own gardens. It's a mess when viewing everyone's gardens. And it has a lot of display mess like the finish button being huge.

Br3nda:
[...]

PR #2371, opened 2020-01-07
GH-2369 update gardens index style

simicic:
### Description
Gardens index page needs some love and this is an attempt of it. 
Following changes were made: 
- page navigation is no longer stretched cross page - I followed existing designs done in plantings and crops index
- jump to is shown only when viewing own gardens
- name of the garden does not have green background
- no perennial plantings is shown if none of the items in category exist
- plantings are shown next to the image instead of below

### Testing
- I tested it by clicking around - looks good
- I also checked how it looks on mobile - and there is this issue of not being stretched cross screen, but after visiting for example plantings index page and going back to gardens - the content is shown correctly - not sure why that happens, let me know if you know what the deal is there...

Here are some examples with my limited data...
Before: 
Screenshot from 2020-01-07 11-15-46

After: 
Screenshot from 2020-01-07 11-20-31

Screenshot from 2020-01-07 11-30-29

### DIFF
--- a/app/controllers/gardens_controller.rb
+++ b/app/controllers/gardens_controller.rb
@@ -4,7 +4,7 @@ class GardensController < DataController
   def index
     @owner = Member.find_by(slug: params[:member_slug])
     @show_all = params[:all] == '1'
-    @my_garden = params[:member_slug].present? ? true : false
+    @show_jump_to = params[:member_slug].present? ? true : false
 
     @gardens = @gardens.includes(:owner)
     @gardens = @gardens.active unless @show_all
--- a/app/controllers/gardens_controller.rb
+++ b/app/controllers/gardens_controller.rb
@@ -5,6 +5,7 @@ class GardensController < DataController
     @owner = Member.find_by(slug: params[:member_slug])
     @show_all = params[:all] == '1'
 
+    @gardens = @gardens.includes(:owner)
     @gardens = @gardens.active unless @show_all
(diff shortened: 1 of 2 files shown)


app/controllers/gardens_controller.rb
```diff
@@ -4,7 +4,9 @@ class GardensController < DataController
   def index
     @owner = Member.find_by(slug: params[:member_slug])
     @show_all = params[:all] == '1'
+    @my_garden = params[:member_slug].present? ? true : false
```
Br3nda: this variable name is confusing. how about 
```suggestion
    @show_member = params[:member_slug].present? ? true : false
```

Br3nda:
if you put "closes #2369" in your pull request description, then github automagically closes that issue when this is merged

Br3nda:
@simicic I've sent you an invite to the growstuff org on github. Once you've accepted you will have the powers to merge this PR yourself. 
Addressing the comment is optional, as i've hit approve anyways .

app/controllers/gardens_controller.rb
simicic: I agree! Suggested term can be also read in 2 different ways - to show a member or that is a show action for member. How about `show_jump_to` - because that is why it is ultimately used for.

simicic:
@Br3nda I accepted the invite i see more stuff about the organization, but I still do not have the write access - but that is ok! ;)

PR #2425, opened 2020-02-14
Fix travis

Br3nda:
Travis-ci build is broken with errors about postgresql not running.

I went to the dev branch and re-ran a buld that was fine last week, and now it's also broken with no porgresql. Some thing has changed somewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 language: ruby
 dist: bionic
 services:
-  - postgresql
+  # - postgresql
   - xvfb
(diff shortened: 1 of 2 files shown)


Br3nda:
it was  installing PostgreSQL 9.4 from apt  --
```yaml
  apt:
    packages:
      # Postgresql version used here should match production
      - postgresql-9.4
      - postgresql-client-9.4
```

 I've changed it to use:
```yaml
addons:
   postgresql: "9.4"
```

PR #2453, opened 2020-03-02
BUGFIX: member slug needed on photos, to make link

Br3nda:
links to photo owner is a broken link whenever the login_name doesn't match the slug.

e.g. when the user has upper case in their login name

### DIFF
--- a/app/models/concerns/search_photos.rb
+++ b/app/models/concerns/search_photos.rb
@@ -25,8 +25,8 @@ module SearchPhotos
         liked_by_members_names: liked_by_members_names,
         # owner
         owner_id:               owner_id,
-        owner_login_name:       owner.login_name,
-        owner_slug:             owner.slug,
+        owner_login_name:       owner_login_name,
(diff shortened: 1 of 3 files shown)


app/models/concerns/search_photos.rb
```diff
@@ -26,6 +26,7 @@ def search_data
         # owner
         owner_id:               owner_id,
         owner_login_name:       owner.login_name,
+        owner_slug:             owner.slug,
```
Br3nda:
```suggestion
        owner_slug:             owner_slug,
```

app/models/concerns/search_photos.rb
```diff
@@ -26,6 +26,7 @@ def search_data
         # owner
         owner_id:               owner_id,
         owner_login_name:       owner.login_name,
```
Br3nda:
```suggestion
        owner_login_name:       owner_login_name,
```

PR #2509, opened 2020-04-13
Auto corrected by following Format Css Code

Br3nda:
Auto corrected by following Format Css Code

Click [here](https://awesomecode.io//repos/Growstuff/growstuff/format_configs/css) to configure it on awesomecode.io

### DIFF
--- a/app/assets/stylesheets/_crops.scss
+++ b/app/assets/stylesheets/_crops.scss
@@ -7,7 +7,7 @@
     bottom: 0;
     color: $white;
     margin: 0;
-    opacity: .8;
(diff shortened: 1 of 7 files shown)


maco:
@Br3nda it looks like Awesome and Code Climate disagree about leading zeros. I can't get into the Awesome config page you linked to tweak it, though.

Br3nda:
yeah, the awesode code seems to wake up once a month and send us PRs, because we only get a few a month on the free tier.
I should fix these config. Awesome code doesn't read the scss lint settings which is frustrating and time consuming.

PR #2555, opened 2020-06-02
Gem upgrades, including Rails 5.2.4.3

Br3nda:
Handful of upgrades. Pinning our rails version in Gemfile has meant we missed these important updates


Br3nda:
Needed to add manifest.js as per https://github.com/rails/sprockets-rails/issues/444

Br3nda:
Failing tests now need the crop.parent added back to the api

PR #2572, opened 2020-06-17
Production release

Br3nda:
using the mainline branch


Br3nda:
Releasing this now, as it has security patch for geocoder

PR #2608, opened 2020-08-16
Feature/parallel tests

Br3nda:


Br3nda:
Older sequential run  
Ran for 19 min 26 sec 

Parallel run
 Ran for 21 min 51 sec

*sigh*

it got SLOWER!

Br3nda:
only saved 30 seconds.

nah, this approach isn't making it faster. I think we need to look at caching dependencies, and builds of nokogiril, and a faster way to get elastic search

PR #2671, opened 2020-10-20
Put if statements about each chart block

### LINKED ISSUE #1988: We should avoid showing graph headings with no graph or provide a fallback
https://www.staging.growstuff.org/crops/thai-basil
image

We show the headings, but no graph content; so its a little confusing

wombleton:
Will see if I can quickly add a test. Resolves #1988 

image


cesy:
Welcome to Growstuff! Thanks for looking at this for us. Please can you add yourself to the contributors file as it's your first pull request? The trailing whitespace issue should also be easy to fix. The tests are also failing, which may need a bit more investigation.

wombleton:
> The tests are also failing, which may need a bit more investigation.

Yes, I don't know how to get it to acknowledge that I've mutated `@crop` or else I've done it the wrong way — I think I need Wise Words here.

spec/views/crops/show.html.haml_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe "crops/show" do
```
Br3nda, line 5:
```suggestion
describe "crops/show", :js do

```

Br3nda (approved):

Br3nda:
Thanks @wombleton!

PR #2676, opened 2020-10-26
Run the rspec on github

Br3nda:
travis has become incredibly slow recently - this will run the rspec tests on github so we'll know sooner if they're broken on a pull request


Br3nda:
Travis-ci has announced it is shutting down

PR #2977, opened 2022-03-26
Ruby 2.7.5

CloCkWeRX:
Get us slightly more up to date, then cut over to ruby 3.0.

I'm ignoring the editorconfig complaint about bundler generating with 3 spaces.


Gemfile.lock
```diff
@@ -719,7 +719,7 @@ DEPENDENCIES
   xmlrpc
 
 RUBY VERSION
-   ruby 2.7.2p137
+   ruby 2.7.5p203
```
Br3nda:
```suggestion
  ruby 2.7.5p203
```
not 100% certain this would work - but it seems to be what codeclimate is suggesting

Br3nda (approved):

Gemfile.lock
CloCkWeRX: Yeah, bundler... will keep putting it back; so kind of just ignoring it.

PR #2986, opened 2022-03-26
Swap percy over

CloCkWeRX:
https://docs.percy.io/docs/migrating-to-percy-cli

- [x] Check https://docs.percy.io/docs/capybara#manually


Br3nda:
:+1:

PR #3142, opened 2022-11-06
Bump rubocop

CloCkWeRX:


CloCkWeRX:
Code Climate is a goose, and can't tell that's auto generated.

PR #3158, opened 2022-11-07
Stacked PR: Various rubocop autocorrections

CloCkWeRX:


.rubocop_todo.yml
```diff
@@ -103,11 +103,12 @@ Lint/SymbolConversion:
   Exclude:
     - 'app/controllers/likes_controller.rb'
 
-# Offense count: 7
+# Offense count: 8
```
Br3nda, line 106: This count goes up

CloCkWeRX: True. What happens often is one set of exclusions is exempting a whole file. Then it gets fixed, and three other problems are found.

For now, am focused on auto correcting the safe changes - will chip away - and then regenerating the todo file to reflect current state regularly.

cesy:
If you want to update this onto latest dev, I'm happy to do a final review and get some of it merged.

lib/actions/oauth_signup_action.rb
```diff
@@ -63,13 +63,11 @@ def determine_name(auth)
     name = ''
```
cesy: Is this line still needed if we're taking out the rest of the definition?

cesy:
All the other changes look good to me, so happy to merge apart from that one question.

cesy (approved):

lib/actions/oauth_signup_action.rb
Br3nda: Surprised the linter allows that unused variable. Must be used somewhere? If so, maybe its important to set it?

Br3nda: __
```suggestion
```

Br3nda (approved):

PR #3182, opened 2022-12-09
Upgrade rails to 7.0.4

cesy:
Let's get ourselves more updated


CloCkWeRX:
Ah, nice.
Dependabot *should* be able to do this for us, so curious it's not happy/able to submit things.

PR #3191, opened 2022-12-10
Issue/2419: Limit map to the most recent plantings only

### LINKED ISSUE #2419: Filter community map by users's activity
Currently, community map allows only for search by location. 
When clicking around, I realized most of the user accounts are inactive since ~2013.

I would love to have a quick way to see active members around me so I can follow them in order to be reminded of planting certain plants. 

Does such feature make sense to you? :)
Thanks!

CloCkWeRX:
Fixes #2419 by limiting the display to about 200 plantings.

image

Separately, we should probably write a cron task or similar to end of life any plantings over 5 years old that are abandonware.


app/assets/javascripts/members.js.erb
```diff
@@ -1,5 +1,5 @@
 if (document.getElementById("membermap") !== null) {
-  L.Icon.Default.imagePath = '/assets';
+  L.Icon.Default.imagePath = '/assets'
```
CloCkWeRX:
```suggestion
  L.Icon.Default.imagePath = '/assets';
```

CloCkWeRX:
Will deploy this to growstuff-staging shortly for a bit of testing.

PR #3192, opened 2022-12-10
Issue/1642: Drop facebook auth

### LINKED ISSUE #556: Connect to Facebook

### LINKED ISSUE #1642: Unable to log in with Facebook
Clicking "Sign in with Facebook" on production brings me to Facebook and this error pops up:

screen shot 2018-05-28 at 11 00 58 am

Googling the issue brings up lots of ways to fix it, whoever has access to the Facebook app settings probably needs to make some changes https://stackoverflow.com/questions/37063685/facebook-oauth-the-domain-of-this-url-isnt-included-in-the-apps-domain

CloCkWeRX:
Fixes #1642
Fixes #556 

The original rationale was we'd be usable in small community groups or gardens, and people could share, etc.

While no doubt there are a lot of community groups still on that platform, I think we've seen over the years they as a company aren't worth the tradeoff.

The alternative is to re-add the facebook credentials and jump through their hoops.


cesy (approved):
Yeah, not worth the hassle of jumping through their hoops for what we're doing at the moment.

cesy:
Maybe let's get the rest of the urgent stuff to production first, though, just in case this causes issues for anyone?

PR #3208, opened 2022-12-15
Reupgrade ruby

cesy:
Since it doesn't seem to have been the actual root cause of the deploy issues


Br3nda (approved):
Looks sensible.

Why did the indenting Chang in the gem lock file?

cesy:
Because codeclimate wants more indents than bundler generates, so we have to fix those manually to keep it happy. Once we've fixed enough of codeclimate's other complaints to see what's going on in more detail, we may be able to tweak its settings.

cesy:
And deploy failed again!

PR #3280, opened 2023-03-13
Release 53

CloCkWeRX:


cesy:
Are we happy enough with this to push it to prod, or want to do some proper checks first?

cesy:
@CloCkWeRX are you happy to change this from draft to ready to go, or what tests do we need to do first? Would be great to get all these security improvements onto prod.

CloCkWeRX:
Oops, missed your comment @cesy! Have shifted it to "ready"

.devcontainer/.env
```diff
@@ -0,0 +1,2 @@
+# Ruby version
+VARIANT=3.1.3
```
CloCkWeRX:
```suggestion
VARIANT=3.1.4
```
For the dev container?

cesy (approved):

PR #3327, opened 2023-04-26
Typo fix

cesy:
Minor fix for codespaces - doesn't appear to break anything either way, but good to be tidy.


CloCkWeRX (approved):

CloCkWeRX:
Annoying. https://github.com/cesy/growstuff/commit/db798aa847efc541a6a1e1979b62cc34b1c2caa3 will be me fat fingering the keyboard at some point. Is it worth us adding a yaml-lint to the CI pipeline for these and other config files?

cesy:
Yes, sounds like a good idea.

PR #3329, opened 2023-04-27
Trial a yaml lint

CloCkWeRX:
Protect against 
https://github.com/Growstuff/growstuff/pull/3327/files


CloCkWeRX:
Doesn't do what I hoped.

PR #3330, opened 2023-04-27
Bump to ruby 3.1.4

### LINKED ISSUE #3315: Codespace minor points
When I load up a new codespace now that everything is merged, it now complains if I try rails s until I do rvm install 3.1.3. It then wants me to do bundle install, which also fails with an error about pg. So although it worked in the codespace branch, somehow it now doesn't work in dev branch, or I'm doing something wrong?

CloCkWeRX:
Fix https://github.com/Growstuff/growstuff/issues/3315

Does it work for Heroku? Supposedly https://devcenter.heroku.com/changelog-items/2561 yes

- [ ] Successful staging deploy via heroku using this version


cesy (approved):
Happy now we've solved codeclimate. Long term we may need to find a way to get it to have different criteria for this one line in this file as it's auto-generated with three spaces instead of four.

Issue #3410, opened 2023-08-09
Forums > Surface the year of last post


CloCkWeRX:
https://www.growstuff.org/forums
![image](https://github.com/Growstuff/growstuff/assets/365751/2ae3cc6d-f1de-4f72-b503-67b84c064c35)

Should show '11 Jan 23' or similar instead of just 11 Jan.

Currently:
```
            - if post.updated_at > post.recent_activity
              = post.updated_at.to_date.to_formatted_s(:short)
            - else
              = post.recent_activity.to_date.to_formatted_s(:short)
 ```
Potentially we could introduce a helper than outputs a span with a title attribute, full date and use that everywhere.

PR #3411, opened 2023-08-09
Forums > Dates > Update _summary.html.haml

CloCkWeRX:
Fix issue #3410

https://github.com/Growstuff/growstuff/assets/365751/48b25fc9-9b0a-4858-9578-691aeca0f1fd


cesy:
I like the theory, but I'm not quite confident enough in my memory of syntax to merge it without checking it does what it's intended to do, and I can't currently get the codespace working well enough to test it. If another maintainer spots this and can check it, that would be great, or if you can drop in some screenshots of your testing?

cesy (approved):

PR #3419, opened 2023-08-11
Upgrade to js-routes 2. Put all js routes into a global namespace.

CloCkWeRX:
I don't actually know if we reference any routes in our javascript!


cesy (changes requested):
I prefer removing it, so have approved and merged that PR instead

PR #3420, opened 2023-08-11
Alternative: Remove js routes

CloCkWeRX:
When I look at Growstuff.org, I can get access to `Routes` via the browser console.

But nothing in the code or compiled assests references it!

So; removing as an alternative to https://github.com/Growstuff/growstuff/pull/3419


cesy (approved):
Since nothing currently uses it, we might as well simplify - if we add something that needs it later, we can add the new version or a competitor at that stage.

PR #3454, opened 2023-09-16
Rubocop fixes (Stacked PR)

CloCkWeRX:
https://github.com/Growstuff/growstuff/pull/3453 first

Various misc bits to get us back to where we were before various rubocop upgrades, almost.


cesy:
I'm happy with these changes if you want to update to the latest dev branch and we can get them in. And it looks like Github can even do the merge for us! But codefactor is unhappy - let me see if I can fix that.

cesy:
And now we have to update contributors for the bot or write an exception for that? \*headdesk\*

CloCkWeRX:
Less sure about how actually useful the "members_controller_routing spec" naming pattern is, it was very clear before - members_routing_spec.... routing for operations on members.

Renamed anyway

cesy:
Yeah, I don't think it's a particularly useful change, but it's not worth fighting to make an exception for.

Issue #2067, opened 2019-07-10
Accessibility audit


Br3nda:
And fixing up semantic web pages 

Issues via Lighthouse

* https://github.com/Growstuff/growstuff/pull/3561
* Links do not have a discernible name (seems to be forum posts)
* Background and foreground colors do not have a sufficient contrast ratio. (https://dequeuniversity.com/rules/axe/4.8/color-contrast). (https://github.com/Growstuff/growstuff/pull/3568 + maybe other areas)
* Heading elements are not in a sequentially-descending order

CloCkWeRX:
I turned off accessbilitylint as it was spamming the heck out of pull requests.

CloCkWeRX:
Home page

Results via Chrome > Lighthouse.
![image](https://github.com/Growstuff/growstuff/assets/365751/f3f17c06-3db6-4162-a5db-93ed549bc7d0)

CloCkWeRX:
Will address the alt tags.
Ignoring the color contrast one, as most of the things it is mad about are deliberately .muted

CloCkWeRX:
![image](https://github.com/Growstuff/growstuff/assets/365751/f5d960c6-dc62-4e67-9f86-4573ed3cbc33)

CloCkWeRX:
Alright, for the buttons can go from #5f8e43 to #57803C without much negative impact.

CloCkWeRX:
![image](https://github.com/Growstuff/growstuff/assets/365751/fd9cfc3e-4058-4076-9245-ed1b699d47db)
With the alt tag fixes

CloCkWeRX:
TODO: accessbilitylint was annoying.

axe-core-capybara plugs into the features.

https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-capybara/README.md

PR #3568, opened 2024-01-17
Darken green buttons, headings etc for Accessbility contrast ratios

CloCkWeRX:
#2067

Before:
image

After:
image
(Note a few places missing the change, as this is me in Chrome rails than via the app)


CloCkWeRX:
Just going to land this one, we can back it out if we don't like the overall effect when we look in staging

PR #3573, opened 2024-01-17
Add Accessibility assertions to rspec, replacing accessbility-lint + fixes

### LINKED ISSUE #2067: Accessibility audit

CloCkWeRX:
Fixes #2067 by adding automated auditing in, that we can opt into per major feature or route.

I only put it in a few places, as we had issues to resolve to or whitelist.

Please squash merge this, so that history is fairly clean.

Things changed:
* Join now call to action is visually quite different
* Colors darkened for constrast
* Use of navy blue for text-muted.

Now:
image

Before:
image

I'm not entirely happy with the results, but don't quite have the design skills to pick a different approach.

Rules suppressed:
* Constrast. There are some links which fail the rules, but changing them really flies against the design.
* Required children. Kept failing on the layouts/nav for this section, even though when I read the spec it's correctly structured (tablist has many tabs)
image
* heading-order. Kept dinging us on a h3 followed by a h5. While I addressed a number of these, we don't have things wildly out of sequence.


app/assets/stylesheets/overrides.scss
```diff
@@ -364,15 +366,23 @@ ul.thumbnail-buttons {
   h1 {
     font-size: 400%;
   }
+  .stats a {q
```
CloCkWeRX:
```suggestion
  .stats a {
```

cesy:
Could the background of the Sign up button on the Join now box be made the same colour as the sign in button? I think that would help the contrast and aesthetics there.

cesy:
I like this approach and the changes look fine in principle.

CloCkWeRX:
> Could the background of the Sign up button on the Join now box be made the same colour as the sign in button? I think that would help the contrast and aesthetics there.

The only thing that makes me reluctant is its meant to be a call to action... (which I have now made fairly drab, boo).
At the moment you definitely notice it as the only thing different, in a panel that is a bunch different.

I'll make the change now, but I think we should still have 'sign up' as the biggest visual action on the landing page.

cesy:
Yeah, I see what you mean. Maybe we just make the text bigger on that one?

Issue #1035, opened 2016-08-23
Do we need to run the full test suite both with and without elasticsearch?


CloCkWeRX:
We do

```
env:
  matrix:
    - GROWSTUFF_SITE_NAME="Growstuff (travis)" RAILS_SECRET_TOKEN='xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' GROWSTUFF_ELASTICSEARCH='true'
    - GROWSTUFF_SITE_NAME="Growstuff (travis)" RAILS_SECRET_TOKEN='xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' GROWSTUFF_ELASTICSEARCH='false'

```

and

```
script:
- script/gemfile_check
- bundle exec script/check_contributors_md
- bundle exec rake db:migrate --trace
- bundle exec rspec spec/
```

But only a small amount of tests around searches rely on elasticsearch. The rest are aftercommit hooks:

```
  after_commit { |sn| sn.crop.__elasticsearch__.index_document if sn.crop && ENV['GROWSTUFF_ELASTICSEARCH'] == "true" }
```

Should we:
- Only run elastic search specific tests with GROWSTUFF_ELASTICSEARCH=true
- Refactor away from after commit hooks for _actions_ or otherwise inject the elastic search provider indepent of saving/creating. IE: Creating a Crop, Seed or Planting via a controller from a user action and indexing it makes sense; but does it make sense to index content for rspec level tests? Or does it just introduce the network into test cases?

This could shave a few minutes off builds by being more specific; as well as cut down on the number of times we have to re-run the whole suite.

cesy:
This seems like a good chance to use tags and only run the tests we need to with elasticsearch, yes.

pozorvlak:
I agree with @Cesy: let's only run elasticsearch-specific tests with
elastic search on (and let's make sure there's a test for those post-commit
hooks).

maco:
:+1:

CloCkWeRX:
Lets introduce tags to do this - https://www.relishapp.com/rspec/rspec-core/v/2-4/docs/command-line/tag-option

CloCkWeRX:
So... 502862128fabff1b9fac592f0be4e8069a050107 may have just reduced our flakey test behaviour significantly.

maco:
I think what happened with my branches was master versus dev...maybe

PR #3587, opened 2024-01-21
Faster rspec in CI

### LINKED ISSUE #1549: Investigate ways to speed up rspec run
A fast test suite is run more often

CloCkWeRX:
Fixes #1549
#1035 for further improvement.

1. Strip back installation that isn't needed.
2. Feature specs in their own thing
3. Fail fast
4. Only output failure detail, so its easier to figure out what you broke


CloCkWeRX:
Nevermind

Issue #2082, opened 2019-07-12
notify user when harvest time starts for aplanting


Br3nda:

CloCkWeRX:
iCal feed would be the simplest way to go.

CloCkWeRX:
Previously #1381

CloCkWeRX:
https://jules.google.com/session/16703221337897327633

CloCkWeRX:
https://github.com/Growstuff/growstuff/pull/4604

Issue #1447, opened 2017-11-12
Encourage users to link their existing harvests to plantings and marl plantings finished 


Br3nda:
This will improve the harvest predictions by filling gaps in the data. An alert (like a notification in the top right) leading to a page that prompts for linking any unlinked harvests to their likely planting.
Also, prompting mark a planting as finished if it hasn't harvested for 2 or more years.

Issue #472, opened 2014-12-12
Expected harvest notifications for seed-sown crops


Skud:
This relies on us having "days to harvest" for seeds.  However, if we have that, and someone plants a crop from seed, we should be able to send them a notification telling them when their crop should be ready to harvest, and encouraging them to enter their harvest data into GS.  This could be part of the regular planting reminder email.

CloCkWeRX:
https://github.com/Growstuff/growstuff/pull/4604

PR #3588, opened 2024-01-21
iCalendar feed of plantings

CloCkWeRX:
Partially improve #1447
Partially solve #472
Partially solve #2082

Another one for squash merging

image


CloCkWeRX:
image
Okay, pretty happy with that. Staging doesn't have predictions, but should work out okay

CloCkWeRX:
On mobile:
Screenshot_20240121-183436
Screenshot_20240121-183430

cesy:
Security-wise, it looks like you only provide the link to the owner, but anyone can access it - is that right, since it's not secret but it's most useful to the owner? Is it worth adding a test to make that behaviour clear, that you can access someone else's calendar if you want to help a friend or something, but the page isn't cluttered with a link unless it's your own?

cesy:
Otherwise it seems good to me on a quick skim.

CloCkWeRX:
I'll have a think about it. The intent was to make multiple feeds, tied to a person and also not, like the RSS feeds, JSON feeds, etc long term.

CloCkWeRX:
For now, going to merge as is.

PR #3607, opened 2024-01-27
HAML 6

CloCkWeRX:


CloCkWeRX:
The maintainers of the HAML gem, *facepalm*

Massively refactored the way the gem works but didn't write a migration guide; and their specs are all very abstract.

So; considering:
* Just deploying to staging
* If it works, getting rid of our specs for our filters

Well; closer:

https://growstuff-staging.herokuapp.com/posts/clockwerx-20160422-where-s-my-lettuce
image

At least it doesn't crash

PR #3610, opened 2024-02-03
Fix haml preview

CloCkWeRX:
Attempts to fix markdown showing in the previews of posts, by rendering to HTML first then truncating:
https://growstuff-staging.herokuapp.com/posts

image


CloCkWeRX:
Well, this does fix the preview by forcing the markdown to be rendered.
But it highlighted that our growstuff_markdown utility... isn't actually translating crops when it's interpolated, which is how it's used.

Issue #3611, opened 2024-02-03
Minor regression on crop search + pagination


CloCkWeRX:
```harvests_count=0 has_photos=true id="364" name="Good King Henry" photos_count=0 planters_ids=#<Hashie::Array []> plantings_count=0 scientific_name="Blitum bonus-henricus" scientific_names=#<Hashie::Array ["Blitum bonus-henricus"]> slug="good-king-henry" thumbnail_url="https://api.gbif.org/v1/image/cache/200x/occurrence/4091740753/media/04b3b3d56c0dba97b2795cd4827063b4">, #<Searchkick::HashWrapper _id="482" _index="crops_production_20200824003628370" _score=3.574971 _type="_doc" alternate_names=#<Hashie::Array []> created_at=1421408385 description=nil harvests_count=0 has_photos=false id="482" name="broccoflower" photos_count=0 planters_ids=#<Hashie::Array []> plantings_count=0 scientific_name="Brassica oleracea Botrytis group" scientific_names=#<Hashie::Array ["Brassica oleracea Botrytis group"]> slug="broccoflower" thumbnail_url=nil>, #<Searchkick::HashWrapper _id="355" _index="crops_production_20200824003628370" _score=2.90951 _type="_doc" alternate_names=#<Hashie::Array ["Boabs Bonnet", "Scotty Bons", "Bonney peppers", "Caribbean red peppers"]> created_at=1388659197 description=nil harvests_count=0 has_photos=false id="355" name="cachucha pepper" photos_count=0 planters_ids=#<Hashie::Array []> plantings_count=0 scientific_name="Capsicum chinense" scientific_names=#<Hashie::Array ["Capsicum chinense"]> slug="cachucha-pepper" thumbnail_url=nil>]:Array
2024-02-03T08:41:31.627490+00:00 app[web.1]: 
2024-02-03T08:41:31.627490+00:00 app[web.1]:       return nil unless collection.total_pages > 1
2024-02-03T08:41:31.627491+00:00 app[web.1]:                                   ^^^^^^^^^^^^):
2024-02-03T08:41:31.627491+00:00 app[web.1]: [ebb24b88-a4f7-4b6a-8890-b5ecf6463568]     30:     instead.
2024-02-03T08:41:31.627491+00:00 app[web.1]: [ebb24b88-a4f7-4b6a-8890-b5ecf6463568]     31: 
[...]

PR #3612, opened 2024-02-03
Revert 140fe7a

CloCkWeRX:
#3611

https://www.staging.growstuff.org/crops/search?term=bob

What I still don't understand is this should have test coverage for exactly the scenario that is broken in prod, but it was not failing.
```
describe "crop search" do
  it "search results show the search term in title" do
    visit root_path
    within "form#navbar-search" do
      fill_in "term", with: "tomato"
      click_button "Search"
    end
    expect(page).to have_css "h1", text: "Crops matching \"tomato\""
  end
```


CloCkWeRX:
Extremely annoying, the specs that have broken are.... the JSON versions of what is broken in the HTML version.

Issue #3620, opened 2024-02-04
ActionView::SyntaxErrorInTemplate (Encountered a syntax error while rendering template


CloCkWeRX:
https://growstuff.org/seeds/pazinjanka-soybean-ed4ecd88-3528-468d-a42c-b86eb9f25965

app/views/seeds/show.html.haml:49: syntax error, unexpected '(', expecting ')'

```
2024-02-04T03:10:11.152931+00:00 app[web.1]: [10452e66-befc-4a32-88a8-bf079f664c51] ActionView::SyntaxErrorInTemplate (Encountered a syntax error while rendering template: check - content_for :title, "#{@seed.owner}'s #{@seed.crop} seeds"
2024-02-04T03:10:11.152933+00:00 app[web.1]: - content_for :opengraph do
2024-02-04T03:10:11.152935+00:00 app[web.1]:   = tag("meta", property: "og:image", content: seed_image_path(@seed))
2024-02-04T03:10:11.152935+00:00 app[web.1]:   - if @seed.description
2024-02-04T03:10:11.152936+00:00 app[web.1]:     = tag("meta", property: "og:description", content: og_description(@seed.description))
2024-02-04T03:10:11.152937+00:00 app[web.1]:   = tag("meta", property: "og:image", content: "#{@seed.owner}'s #{@seed.crop} seeds")
2024-02-04T03:10:11.152938+00:00 app[web.1]:   = tag("meta", property: "og:type", content: "website")
2024-02-04T03:10:11.152939+00:00 app[web.1]:   = tag("meta", property: "og:url", content: request.original_url)
2024-02-04T03:10:11.152939+00:00 app[web.1]:   = tag("meta", property: "og:site_name", content: ENV['GROWSTUFF_SITE_NAME'])
2024-02-04T03:10:11.152940+00:00 app[web.1]: 
2024-02-04T03:10:11.152941+00:00 app[web.1]: 
2024-02-04T03:10:11.152941+00:00 app[web.1]: - content_for :breadcrumbs do
2024-02-04T03:10:11.152941+00:00 app[web.1]:   %li.breadcrumb-item= link_to 'Seeds', seeds_path
2024-02-04T03:10:11.152942+00:00 app[web.1]:   %li.breadcrumb-item= link_to @seed.owner, member_seeds_path(@seed.owner)
2024-02-04T03:10:11.152942+00:00 app[web.1]:   %li.breadcrumb-item.active= link_to @seed, @seed
[...]

Issue #3619, opened 2024-02-04
Selenium::WebDriver::Error::TimeoutError:        timeout: Timed out receiving message from renderer: 1.000


CloCkWeRX:
Seemingly only for dev -> mainline, we get a timeout after a second.
```
Failures:

  1) Gardens#index Logged in as member with 10 gardens is accessible
     Failure/Error: expect(page).to be_axe_clean.skipping('color-contrast', 'heading-order', 'aria-required-children').according_to :wcag2a

     Selenium::WebDriver::Error::TimeoutError:
       timeout: Timed out receiving message from renderer: 1.000
         (Session info: chrome-headless-shell=121.0.6167.139)
     Shared Example Group: "is accessible" called from ./spec/features/gardens/index_spec.rb:17
```

Happens across multiple different areas.

See https://github.com/Growstuff/growstuff/pull/3613

PR #3621, opened 2024-02-04
Seeds > Photos - Fix tests not running and fix tests

CloCkWeRX:
#3620

But I don't see the template parsing error.


CloCkWeRX:
Test failures are #3619

PR #3684, opened 2024-03-23
Change naming to be aware of plant parts, even if it sounds oddly formal

### LINKED ISSUE #3682: Improve harvest naming
https://www.growstuff.org/harvests/clockwerx-lettuce
5 lettuces vs what actually harvested, 5 leaves
image

CloCkWeRX:
Fix https://github.com/Growstuff/growstuff/issues/3682


cesy:
Looks like this might have some missing code or something to make it work and tests pass? I think it might need a bit more detail and some more testing before we put it live.

PR #4222, opened 2025-09-07
feat: Add prompts for new activities

CloCkWeRX:
This commit introduces three new prompts to the application to suggest relevant follow-up activities to the user.

- When a user finishes the last planting in a garden, they are prompted to create a new activity for soil cultivation.
- When a user creates a new garden, they are prompted to plan a weeding activity for two weeks in the future.
- When a user marks an activity as finished, they are prompted to repeat the activity in two weeks.

These prompts are displayed as flash messages and include links to pre-filled forms for the new activities.

### DIFF
--- a/app/controllers/gardens_controller.rb
+++ b/app/controllers/gardens_controller.rb
@@ -38,7 +38,10 @@ class GardensController < DataController
 
   def create
     @garden.owner_id = current_member.id
-    flash[:notice] = I18n.t('gardens.created') if @garden.save
+    if @garden.save
+      link = new_activity_path(name: 'Weed the garden bed', garden_id: @garden.id, due_date: 2.weeks.from_now.to_date)
+      flash[:notice] = t('gardens.created_prompt_html', link: link).html_safe
+    end
     respond_with(@garden)
   end
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -125,6 +125,7 @@ en:
     updated: Garden was successfully updated.
     confirm_delete: All plantings associated with this garden will also be deleted. Are you sure?
     confirm_deactivate: All plantings associated with this garden will be marked as finished. Are you sure?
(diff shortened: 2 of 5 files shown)


app/controllers/gardens_controller.rb
```diff
@@ -39,7 +39,10 @@ def edit
 
   def create
     @garden.owner_id = current_member.id
-    flash[:notice] = I18n.t('gardens.created') if @garden.save
+    if @garden.save
+      link = new_activity_path(name: 'Weed the garden bed', garden_id: @garden.id, due_date: 2.weeks.from_now.to_date)
+      flash[:notice] = t('gardens.created_prompt_html', link: link).html_safe
```
CloCkWeRX, line 44: Refactor this to use link_to helpers so html_safe is not required, and repeat for all similar areas.

config/locales/en.yml
```diff
@@ -125,6 +125,7 @@ en:
     updated: Garden was successfully updated.
     confirm_delete: All plantings associated with this garden will also be deleted. Are you sure?
     confirm_deactivate: All plantings associated with this garden will be marked as finished. Are you sure?
+    created_prompt_html: "Garden created. Would you like to <a href=\"%{link}\">plan to weed this garden bed in two weeks</a>?"
```
CloCkWeRX:
```suggestion
    created_prompt_html: "Garden was successfully created. Would you like to <a href=\"%{link}\">plan to weed this garden bed in two weeks</a>?"
```

config/locales/en.yml
```diff
@@ -325,6 +327,7 @@ en:
     string: "%{crop} planting in %{garden} by %{owner}"
     progress:
       progress_0_not_planted_yet: 'Progress: 0% - not planted yet'
+    finished_prompt_html: "Planting finished. Would you like to <a href=\"%{link}\">plan a soil cultivation activity</a>?"
```
CloCkWeRX:
```suggestion
    finished_prompt_html: "Planting was successfully updated. Would you like to <a href=\"%{link}\">plan a soil cultivation activity</a>?"
```

PR #4564, opened 2026-04-26
Memory usage optimization

CloCkWeRX:
This submission implements several memory optimization techniques:
1. **Batch Processing**: Updated `GbifService#import!` to use `find_each` for iterating over `Crop` records, preventing the entire table from being loaded into memory.
2. **SQL-Level Sorting**: Refactored `Post.recently_active` and `Member.nearest_to` to perform sorting at the database level instead of in Ruby. This avoids instantiating large numbers of ActiveRecord objects just for sorting.
3. **Efficient View Rendering**: Replaced `collection_select` with `pluck`-based `select` in several forms (`crops`, `scientific_names`, `alternate_names`). This fetches only the required columns (ID and name) as raw values instead of full ActiveRecord objects.
4. **Query Optimization**: Cleaned up the `index` action in `Admin::MembersController` to remove redundant relations.

---
*PR created automatically by Jules for task [2149092598558110155](https://jules.google.com/task/2149092598558110155) started by @CloCkWeRX*

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,7 +7,6 @@ tmp/*
 node_modules
 public/assets
 .env
-.ruby-version
 .ruby-gemset
 .editorconfig
 .esignore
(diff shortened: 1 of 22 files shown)


CloCkWeRX:
#1640

Issue #1044, opened 2016-09-15
Account deletion for free accounts with no activity


cesy:
See #468 for further discussion. If the person is not a crop wrangler or admin and has not created any non-default garden, not planted anything, then we can delete them with no risk.

cesy:
Feel free to grab bits from my draft at https://github.com/cesy/growstuff/tree/issue468deletion to reuse in this.

CloCkWeRX:
https://github.com/Growstuff/growstuff/pull/4578

PR #4578, opened 2026-04-26
Delete inactive members with no activity in 3 years

CloCkWeRX:
1. Backup taken in heroku.
2. Manually run for now, but could be scheduled later.


CloCkWeRX:
#1640

CloCkWeRX:
#1044

PR #4584, opened 2026-04-27
Update member_slug lookup to 404 when not found

CloCkWeRX:
Updated several controllers (Seeds, Activities, Posts, Plantings, Harvests, and Gardens) to use Member.find_by!(slug: params[:member_slug]) when params[:member_slug] is present. This ensures that an ActiveRecord::RecordNotFound exception is raised (leading to a 404 response) if an invalid slug is provided, while maintaining existing filtering behavior when a valid slug is found.

---
*PR created automatically by Jules for task [17599607196474746471](https://jules.google.com/task/17599607196474746471) started by @CloCkWeRX*

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -7,9 +7,9 @@ class ActivitiesController < DataController
     where = {}
     where['active'] = true unless @show_all
 
-    if params[:member_slug]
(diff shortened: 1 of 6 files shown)


CloCkWeRX:
#1640

PR #4595, opened 2026-04-28
Optimize Harvests with memoization and fragment caching

CloCkWeRX:
Implemented memoization for display and calculation methods in the `Harvest` model and `PredictHarvest` concern to reduce redundant processing. Added fragment caching for harvest-related partials, including a strategy to prevent stale relative time strings by including the current date in the cache key.

---
*PR created automatically by Jules for task [9176733581563564983](https://jules.google.com/task/9176733581563564983) started by @CloCkWeRX*

### DIFF
--- a/app/views/crops/_harvests.html.haml
+++ b/app/views/crops/_harvests.html.haml
@@ -5,7 +5,8 @@
       %p Nobody has harvested this crop yet.
   - unless crop.harvests.empty?
     %ul.list-group.list-group-flush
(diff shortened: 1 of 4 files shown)


app/views/crops/_harvests.html.haml
```diff
@@ -6,13 +6,14 @@
   - unless crop.harvests.empty?
     %ul.list-group.list-group-flush
       - Harvest.where(crop: crop).includes(:owner).order(harvested_at: :desc).limit(3).each do |harvest|
```
CloCkWeRX: @jules cache this query as well, unless the whole partial is already cached

PR #4619, opened 2026-05-03
Replace Sidekiq with Solid Queue

CloCkWeRX:
I have replaced Sidekiq with Solid Queue for background job processing. The implementation includes environment-specific configuration, database schema setup, and integration with Puma via a plugin. I've also ensured that the `Gemfile.lock` remains compatible with multiple platforms and avoids unnecessary regressions.

---
*PR created automatically by Jules for task [9720308009887171417](https://jules.google.com/task/9720308009887171417) started by @CloCkWeRX*


CloCkWeRX:
The good news is this fixed notifications.
The bad news is it's chewing up memory even with the puma plugin :/

PR #4786, opened 2026-09-14
Fix new harvest from planting not prefilling crop and add regression specs

CloCkWeRX:
This change resolves an issue where creating a new harvest from a planting did not prefill the associated crop onto the harvest object or template instance variables. It updates `HarvestsController#new` to handle `planting_slug` and prefill `@harvest.crop` from `@harvest.planting.crop`, and adds unit and request regression tests.

---
*PR created automatically by Jules for task [3655302136529192021](https://jules.google.com/task/3655302136529192021) started by @CloCkWeRX*


CloCkWeRX:
Pending manual regression test locally

Edit: I'll just validate post deploy, can't be more broken than it is now

PR #4795, opened 2026-09-18
release92 (Memory optimisation, misc fixes)

CloCkWeRX:

PR #4772, opened 2026-08-24
Harden Rack::Attack and pagination against aggressive web crawlers

CloCkWeRX:
This change hardens the Rack::Attack configuration against abusive web crawlers:

1. **Daily Page Crawl Rate Limit**: Configured `Allow2Ban` in `config/initializers/rack_attack.rb` to ban IPs that request more than 500 pages in a 24-hour period for 1 week (7 days).
2. **Honeypot Ban Route**: Disallowed `/dont-crawl-me` in `public/robots.txt` and `config/robots.txt`, created the GET route in `config/routes.rb`, and configured `Fail2Ban` in `Rack::Attack` to ban any IP visiting `/dont-crawl-me` for 7 days upon 1 hit.
3. **Out-Of-Bounds Pagination Handling**: Updated `PhotosController#index` to raise `ActiveRecord::RecordNotFound` when `@photos.out_of_bounds?` (e.g. requesting `page=105` when fewer pages exist), returning HTTP 404 Not Found to stop crawlers from endlessly probing invalid page numbers.
4. **Automated Specs**: Added `spec/requests/rack_attack_spec.rb` and updated `spec/controllers/photos_controller_spec.rb` to verify honeypot banning, rate limit banning, and 404 pagination responses.

---
*PR created automatically by Jules for task [2674394950484400196](https://jules.google.com/task/2674394950484400196) started by @CloCkWeRX*

PR #4800, opened 2026-09-19
Stop Rack::Attack depending on memcached, and cut the memcached timeout

Br3nda:
## Why

Production keeps crashing. Heroku kills the web dyno when it uses too much memory.

I read the production logs for `growstuff-prod` from 2026-09-19, 23:24 to 23:35 UTC. Here is what they show:

- The web dyno is a Basic dyno. It has 512 MB of memory.
- The dyno used up to 1 GB of memory. Heroku logged R14 (memory quota exceeded) 19 times.
- Heroku then logged R15 and killed the dyno with SIGKILL. The dyno stayed down for about 3 minutes and 40 seconds.
- One IP address made 753 of the 1,280 requests. 717 of those requests succeeded.
- Memcached timed out again and again.

Rack::Attack should have stopped that IP. It has a limit of 15 requests per minute on some pages. It also bans an IP after 500 page requests in a day. Neither rule stopped the IP.

My best explanation is that Rack::Attack keeps its counters in `Rails.cache`. In production, `Rails.cache` is memcached. When memcached times out, the counters stop working. I have not proven this on production. The first change below makes it independent of memcached either way.

Release v508 (2026-09-18) turned on two Puma workers. That is [#4793](https://github.com/Growstuff/growstuff/pull/4793). Before v508, Puma ran a single process. Two workers on a 512 MB dyno use more memory than the dyno has. This PR does not change that. See "Not in this PR" below.

## What this PR changes

There are three small commits.

1. **Rack::Attack keeps its counters in process memory.** It no longer uses memcached. The store is capped at 8 MB. Each Puma worker has its own counters. This means the limits apply per worker.
2. **Rack::Attack blocks the IPs in a `BLOCKED_IPS` variable.** The variable is a comma separated list. This lets us block a crawler without committing its IP to the repository.
3. **The memcached socket timeout drops from 1.5 to 0.5 seconds.** Dalli retries after a failure. One bad cache call could block a Puma thread for longer than the timeout. Each Puma worker has only 5 threads.

## How I tested

- The Rack::Attack specs pass. I added specs for the new store and for `BLOCKED_IPS`. One new spec breaks `Rails.cache` on purpose. It checks that bans still work.
- All 164 specs in `spec/requests` pass, plus the photos controller spec.
- I booted the production config locally. Rack::Attack uses a `MemoryStore`. The cache store has `socket_timeout: 0.5`.
- I ran the specs with `TZ=UTC`. The two "wait 7 days" specs in `spec/requests/rack_attack_spec.rb` fail without my change when the local timezone has a DST change inside the next 7 days. `travel 7.days` then moves the clock less than 7 real days. CI uses UTC. I did not fix that here.
- I could not test the memcached timeout against real memcachier. Please watch the logs after deploy.

## Not in this PR

These need a decision from the people who run production. I have not changed anything on Heroku.

- **Set `WEB_CONCURRENCY=1` on Heroku.** This gives the dyno the same memory profile it had before v508. Or move to a Standard-2X dyno (1 GB) to keep two workers.
- **Set `BLOCKED_IPS`** to block that crawler. I did not identify who runs the IP. The address is in the production logs (Heroku router logs).
- **Consider `MALLOC_ARENA_MAX=2` or the jemalloc buildpack.** I have not checked whether the Heroku Ruby buildpack already sets `MALLOC_ARENA_MAX`.
- **Consider `rack-timeout`.** The slowest requests took 8 to 17 seconds. I left it out. Stopping a request in a threaded Puma server needs care.
- **Look at the slow pages.** Examples are `/harvests?page=17`, `/places/<address>` and the `.csv` exports.
- **Check the Sidekiq config.** `config/initializers/sidekiq.rb` points at `redis://localhost`. I did not see a Redis add-on on Heroku. I have not investigated this.
- **Rotate credentials.** While I was investigating, a Heroku config command printed secret values into an AI assistant session. They were not sent anywhere else. Rotating the database and AWS keys is a cheap precaution.

## Related

- Refs [#1640](https://github.com/Growstuff/growstuff/issues/1640). This is the open issue about exceeding the RAM quota. This PR does not close it.
- [#4792](https://github.com/Growstuff/growstuff/pull/4792) (places memory) and [#4795](https://github.com/Growstuff/growstuff/pull/4795) (release92) are open memory work. This PR touches none of their files.
- [#4772](https://github.com/Growstuff/growstuff/pull/4772) added the Rack::Attack rules that this PR keeps working.
- [#4793](https://github.com/Growstuff/growstuff/pull/4793) turned on the Puma workers.

## AI assistance

Assisted by Claude Code (Claude Sonnet 5, Anthropic).

Claude read the production logs and the code. It wrote the changes, the specs and this description. Brenda directed the work.

A human has not yet reviewed the code line by line. Please review it as you would any other PR. This PR is a draft until that happens.


CloCkWeRX (approved):

CloCkWeRX:
I suspect the application ended up in a crash loop hitting whatever limit, before merging the recent rack attack changes - app didn't boot up enough before it could write to memcached.

A while later though; this happened:
Screenshot 2026-09-19 174037

So I think the rack attack changes kicked in.
(I'd also deployed straight from dev to trial those changes in release92)

Let's still give this a whirl though

PR #4803, opened 2026-09-20
Add rake tasks to pull crops and a member's data from production

Br3nda:
## Summary
Two rake tasks for filling a local development database with real data from growstuff.org, built to be gentle on a server that has been struggling.

```bash
# approved crops
bundle exec rake crops:pull_from_production

# one member's gardens, plantings and planting photos
MEMBER=login_name bundle exec rake members:pull_from_production
```

Options: `SOURCE_URL`, `DELAY` (seconds between requests, default 2), `MAX_PAGES` (crops), `ACTIVE_ONLY=true` (member: only active gardens and current plantings), `PHOTOS=false`, `LOCAL_EMAIL` and `LOCAL_PASSWORD` (member). Both tasks refuse to run when `Rails.env.production?`.

## How it stays gentle
- One request at a time, with a pause before every request after the first.
- Exponential backoff (or `Retry-After`, capped at 2 minutes) on timeouts, 429s and 5xx responses.
- A hard stop after 3 failed retries instead of hammering. Re-running is always safe, because records are matched and updated rather than duplicated.
- Stops if the server keeps sending the same page, or too many pages, and never follows a `next` link to a different host.
- Only the crops a member's plantings actually use are fetched, so importing a member doesn't need a full crop pull.

## What comes across
- **Members:** created locally if missing (confirmed, with a printed throwaway password unless `LOCAL_PASSWORD` is given).
- **Gardens:** including location, coordinates, description and active flag. Saved with `validate: false`, which skips the geocoding and Wikidata lookups that run after validation, since the values already come from the source.
- **Plantings:** matched by slug, and only into gardens the member owns.
- **Photos:** for plantings. Flickr photos are stored as records pointing at Flickr's own image URLs, as on production, so nothing is downloaded from Flickr.

Not imported: garden photos (the site has no way to list them), garden types (ids may not line up locally), harvests, seeds and activities.

## Things I found about the production API along the way
These are why gardens and plantings use the site's own `.json` endpoints rather than `/api/v1`, and they will matter to anything else that wants to use the API (e.g. React components):
- `/api/v1/crops` pages by offset (`page[limit]`, `page[offset]`) and ignores `page[number]`/`page[size]`. Following `links.next` is the only reliable way to page.
- The related-resource endpoints (`/api/v1/members/:id/gardens`, `/members/:id/photos`) appear to return the resource whose id equals the parent's id, not the parent's children. Importing from `/members/15/gardens` gave a different member's garden.
- `GET /api/v1/photos/:id?include=plantings,gardens` returned a 500.
- Relationship linkage (`data`) is missing from responses even for `always_include_linkage_data` relationships.
- `GardenResource` only exposes `name`.
- The JSON:API needs `Accept: application/vnd.api+json`; `application/json` gets a 406.

## Test plan
- [x] `bundle exec rspec spec/services/remote_api_client_spec.rb spec/services/crop_import_service_spec.rb spec/services/member_import_service_spec.rb` (41 examples; HTTP is stubbed, nothing touches production)
- [x] `rubocop` clean on the changed files
- [x] Ran `members:pull_from_production` against production for a real account: gardens and plantings import correctly; a run stopped cleanly on production 503s as designed


CloCkWeRX:
Ah, through heroku we can just get a full DB dump pretty easily - https://devcenter.heroku.com/articles/heroku-postgres-import-export might be a better approach

Br3nda:
I was thinking this is useful for people who don't' have heroku access.

CloCkWeRX (approved):
Fair enough!

PR #4835, opened 2026-09-22
Add a garden layout page for mapping where plants sit in a bed

### LINKED ISSUE #464: Garden map/plan
Let people create (or upload?) a map/plan of their garden, and lay out plants in it.

Br3nda:
Closes #464.

Adds a garden layout page where a member lays out plants in a bed on a drawn map. The home page, the garden cards and the garden page all point to it.

image

## What it does

The page is at `/members/:member_slug/gardens/:slug/layout`. It shows the bed as soil in a timber frame, with a twine grid sized by the owner. To its left is a sidebar of the garden's plantings as crop chips, listed by crop name, which stays in view as a tall bed is scrolled. Above the bed is a toolbar of garden features.

- **Placing:** dragging a chip onto the bed places one plant of that crop. The chip stays, so it can be dragged again. Plants can go anywhere on the bed, not just in a grid cell, and are drawn as circles with the crop's icon.
- **Moving and lifting:** plants on the bed can be dragged around. Dragging one back to the sidebar takes it off the bed.
- **Composting:** dropping a plant, or a chip, in the compost bin deletes one plant, so the planting has one fewer.
- **Resizing:** each plant has a handle for making it bigger or smaller, since a tomato takes more room than basil. The next plant taken from the same planting comes out at the last size used. Plants always stay wholly on the bed.
- **Telling plants apart:** a plant big enough on screen has its crop's name under it. Pointing at a plant, or at its chip, picks out all of that planting's plants and fades the rest. Plantings drawn with the same icon get a coloured ring, with a matching dot on their chip. There are five colours, the ones that stay distinguishable from each other against the soil, colour blind or not.
- **Bed size:** the owner can drag the bed's right edge, bottom edge or corner, with an outline showing the new size, or type the columns and rows above it. Shrinking it past anything on it is refused.
- **Garden features:** besides plants, the toolbar has stepping stones, sprinklers, taps and stakes (points), rows, paths, drip lines, fences and trellises (lines, with draggable ends), labels, and netting (an area over the plants, with draggable corners). They're dragged onto the bed, moved, named by double-clicking, and removed in the compost bin.
- **Add a planting:** opens the same dialog as the garden cards. **Clear bed** takes every plant off the bed, and leaves the garden features where they are.

Anyone who can see the garden can see its layout. The owner and collaborators can arrange plants. Only the owner can resize the bed or plant into it, because both edit the garden itself.

**Pointers to it:**

- **Home page:** a card at the top. A signed-in member goes to the layout of their active garden with the most growing in it. A member with no active garden is asked to add one, and anyone signed out to sign up.
- **Garden page:** a "Try out the new garden layout tool" card, not shown on inactive gardens.
- **Gardens list:** a "Layout" button on each card.

## Data

- **`plants`** (new): one row per individual plant in a planting, with its position (`bed_x`, `bed_y`, in grid cells, as floats) and an optional `diameter`. A planting of ten tomatoes is ten plants. This also gives per-plant state, such as one plant failing, somewhere to live later.
- **Plants are made on first use.** Plants are only created when someone who can arrange a garden opens its layout, for the plantings growing there that don't have plants yet (`Garden#prepare_layout`, in one insert). Most plantings never go on a layout, so existing and new plantings get no rows until then, and there is no backfill. Someone only looking at a layout creates nothing.
- **`plantings.quantity`** follows the number of plants once a planting is on a layout. Dragging one out adds a plant and composting one removes it. Changing the quantity elsewhere adds or removes unplaced plants to match, but only for plantings already on a layout. A blank quantity counts as one plant; zero means none, so a planting composted to nothing isn't given a new plant. Each planting is capped at 100 plants (`Planting::MAX_PLANTS`).
- **`gardens.grid_columns` and `grid_rows`**: the bed's size, 10 × 10 by default, at most 50 each way.
- **`gardens.layout_features`** (JSON): the garden features on the bed, each with a kind and its position in grid cells, and for a line or netting its second point. They're saved with the plants and checked for being on the bed, but written without a full garden save, which would geocode the garden on every drag.
- **`crops.default_diameter`**: how big a crop's plants are drawn until one is resized. If a crop has none it uses its parent crop's, and failing that, one cell. It's empty for every crop for now, and there's no form to set it yet.

The layout saves the whole arrangement in one request, inside a transaction, so a bad placement leaves the bed as it was.

## Not done yet

- **Setting crop default sizes.** There's no form for `crops.default_diameter` yet.
- **Keyboard use.** Placing, moving and resizing are mouse or touch only.
- **Empty plantings.** Composting a planting's last plant leaves it at zero plants. It isn't marked finished.

## Testing

tested by hand in the browser throughout.


CloCkWeRX:
```
rspec ./spec/features/gardens/garden_layout_spec.rb:41 # The garden layout puts a plant on the bed where a chip is dropped
rspec ./spec/features/gardens/garden_layout_spec.rb:125 # The garden layout the size of the bed refuses to shrink past a plant, and says why
rspec ./spec/features/gardens/inactive_gardens_spec.rb:42 # A member's inactive gardens when the owner is signed in lets them reactivate a garden
```

Br3nda:
there's a flaky spec in `dev` branch. I'll fix in it's own pr

CloCkWeRX:
What's the mobile UX like?
(Will try on own setup when I am free from work)

Br3nda:
> What's the mobile UX like?
> (Will try on own setup when I am free from work)

It collapses so it's harder to drag around if you have a large list of planting in your bed. But it's works nicely for most.

PR #4837, opened 2026-09-22
Give crops icons from Fluent Emoji, chosen by crop wranglers

Br3nda:
Gives crops icons again, from Microsoft's Fluent Emoji, and lets crop wranglers choose them.

Stacked on #4836, so its CI can install gems. Once that's merged, this retargets to `dev`.

## Why

Crop icons came from OpenFarm, which is gone. Its fetching code was removed as defunct, so the icons production has are a frozen snapshot: crops added since have none, and there's no way to add one. Any crop without an icon is drawn as the same generic sprout, which makes plantings hard to tell apart on the garden cards and the new garden layout.

## What changes

- **Icons:** 56 fruit, vegetable, herb, flower, tree and grain icons, in the flat style of [Fluent Emoji](https://github.com/microsoft/fluentui-emoji), are added to `app/assets/images/crops/`. They're MIT licensed, which allows them to be redistributed from this public, AGPL-licensed repository. The licence and the source commit are recorded alongside them.
- **`crops.icon`:** names one of those icons. Crop wranglers choose it on the crop form, next to "Parent crop".
- **The order icons are chosen in:** a crop is drawn with its own icon, then the icon OpenFarm had for it, then its parent crop's. Production keeps its OpenFarm icons wherever no new icon has been chosen, and choosing one for "tomato" covers every tomato variety. Everything that shows crop icons goes through `Crop#svg_icon`, so crop chips, garden cards, the crop pages and `/crops/:slug.svg` all pick them up with no other changes.
- **`rake crops:assign_icons`** gives existing crops icons by name. It knows other names crops go by (aubergine, capsicum) and sends herbs to the herb icon. A crop with no parent is also matched on the end of its name, so "cherry tomato" gets the tomato. Named chillies are listed, so "habanero pepper" isn't drawn as a bell pepper. It only fills in crops without an icon, never replacing one a wrangler has chosen, and `DRY_RUN=true` lists what it would do.

## Why not Icons8

The app's existing interface icons, in `app/assets/images/icons/`, are from Icons8. They were added in 2019 with Icons8's permission (`fcf4e0f15`, "Icons from icon8, used with permission"), and they stay as they are.

For crop icons we'd need a much larger set, a hundred or more, to live in this public repository. Rather than stretch that original permission to cover it, this uses an openly licensed set, which settles the question outright. Icons8 icons could still fill gaps later, with permission that covers them, through the same `crops.icon` column.

## Deploying

After the migration, run it once, dry run first:

```
rake crops:assign_icons DRY_RUN=true
rake crops:assign_icons
```

On a development copy of 68 crops it tagged 52. The rest are crops with no matching emoji (fig, ackee, achiote, beet, ginger), which wranglers can set by hand.

## Checked

The crop model, matcher, crop icon request and garden card serializer specs pass: 122 examples. They cover the choice order, validation, name matching and dry run, the SVG endpoint, and a wrangler setting an icon on the form.


CloCkWeRX:
Could you chuck some notes into https://github.com/Growstuff/growstuff/pull/4825 so neither of us forget to run the rake/db migraiton bits on deploy?

